PR #40, opened 2019-06-26
Master docker compose

sukaant:
Create a docker compose for local environment.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,6 +22,61 @@ services:
       interval: 5s
       timeout: 20s
       retries: 3
+  zookeeper:
+    image: wurstmeister/zookeeper
+    ports:
+    - "2181:2181"
+  kafka:
+    image: wurstmeister/kafka
+    depends_on:
+    - zookeeper
+    ports:
+    - "9094:9094"
+    environment:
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,248 @@
...
+    - minio
+    image: smartcitiesdata/presto-testo:development
+    ports:
+    - "8081:8080"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -s http://localhost:8080/v1/info | grep -q '\"starting\":false'"]
+      interval: 10s
+      timeout: 30s
+      retries: 10
+  data-generator:
```
ManApart: Way to be resourceful on this! 
I'm going to request changes that we don't merge this in until we've played the open source card so this points at an actually built image.

jakeprem:
https://hub.docker.com/r/smartcitiesdata/data-generator

@ManApart 
@ManApart 
@ManApart

docker-compose.yaml
jakeprem: https://hub.docker.com/r/smartcitiesdata/data-generator
https://hub.docker.com/r/smartcitiesdata/data-generator
https://hub.docker.com/r/smartcitiesdata/data-generator

PR #139, opened 2019-09-13
Move andi app into this umbrella project

jdenen:
This PR turns smartcitiesdata into an umbrella project and pulls `smartcitiesdata/andi` in as its first app. There are a bunch of commits and files, but most of them are pulled in via `andi` history. 

Check out the last two commits for things I changed.

As of the opening of this PR (Sept 13, 3:30 PM), this is up-to-date with `andi`'s `master` branch.


apps/tasks/lib/mix/tasks/app/version.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Mix.Tasks.App.Version do
+  @moduledoc false
+  def run(_) do
+    Mix.Project.config()
+    |> Keyword.get(:version)
+    |> IO.puts()
```
jeffgrunewald: Is this worth making a Logger call?

apps/tasks/config/config.exs
```diff
@@ -0,0 +1,30 @@
+# This file is responsible for configuring your application
```
jeffgrunewald: Lotta boiler plate in here.

apps/tasks/README.md
```diff
@@ -0,0 +1,21 @@
+# Tasks
+
+**TODO: Add description**
```
jeffgrunewald: What do I do or what am I intended to do?

apps/tasks/.formatter.exs
```diff
@@ -0,0 +1,4 @@
+# Used by "mix format"
```
jeffgrunewald: Is this another instance of a file that could inherit from the top-level formatter or do we want to have one per-app?

scripts/deploy.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -e
+
+apps=(andi)
```
LtChae: This releases all the apps every time we tag the mono-repo. Is that intended to be a strategic stop-gap maneuver™ or is this the new state of things going forward?

apps/tasks/lib/mix/tasks/app/version.ex
jdenen: No, the point of the task is to write the app version to your console. If it put the log level in there too, we'd have to parse that piece.

apps/tasks/.gitignore
```diff
@@ -0,0 +1,24 @@
+# The directory Mix will write compiled artifacts to.
```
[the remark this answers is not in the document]
jdenen: No we don't. I'll fix this.

mix.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Smartcitiesdata.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      apps_path: "apps",
```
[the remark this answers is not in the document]
jdenen: Not worth it, IMO.

apps/tasks/config/config.exs
jdenen: 🤷‍♂

apps/tasks/README.md
jdenen: Yeah, yeah. Documentation is important.

scripts/deploy.sh
jdenen: This is the way I intend it to work, but I need to use the strategic stop-gap maneuver™ of per-app versioning in the meantime. So I need to fix that for now.

apps/tasks/.formatter.exs
jdenen: Top level should be fine. Nice catch.

apps/andi/config/config.exs
```diff
@@ -0,0 +1,32 @@
...
+# This configuration file is loaded before any dependency and
+# is restricted to this project.
+
+# General application configuration
+use Mix.Config
+
+# Configures the endpoint
+config :andi, AndiWeb.Endpoint,
+  url: [host: "localhost"],
+  secret_key_base: "z7Iv1RcFiPow+/j3QKYyezhVCleXMuNBmrDO130ddUzysadB1stTt+q0JfIrm/q7",
```
jeffgrunewald: Is this a good time to remove this secret key from the source control before we migrate it into another (more permanent) repo?

jdenen: I think it's best (for local development if nothing else) to leave this here. I put a comment in saying that it should be overwritten.

jdenen:
I completely forgot about how we set that up in `discovery_api`, so that's a good call. Looks like everyone has given the all clear, so I'm going to merge.

PR #151, opened 2019-09-19
Add Forklift to the umbrella

jdenen:
- Merges forklift into the umbrella
- Configures forklift for being a part of the umbrella
- Update andi dependencies as needed
- Adds `mix test` alias (temporary pending `brook` updates)


jdenen:
I made changes in the last 4 commits, so you'll probably want to focus on those. Feel free to re-review the entire history of Forklift if you'd like, though.

jdenen:
Yes. That change was made before this PR, but there is some stuff there that needs to be tweaked. I'll probably do that as part of my next PR (EOF messages) or a third for that specifically.

apps/forklift/rel/plugins/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.*
```
jeffgrunewald: this still needed?

apps/forklift/rel/config.exs
```diff
@@ -0,0 +1,34 @@
+~w(rel plugins *.exs)
```
jeffgrunewald: this needed?

apps/forklift/rel/prod.vm.args
```diff
@@ -0,0 +1,7 @@
+-name <%= release_name %>@${ERLANG_NAME}
```
jeffgrunewald: this needed?

apps/forklift/rel/vm.args
```diff
@@ -0,0 +1,7 @@
+-name <%= release_name %>@127.0.0.1
```
jeffgrunewald: this needed?

PR #169, opened 2019-09-23
Adding endpoint to disable a dataset

JarredOlson:


apps/andi/lib/andi/services/dataset_disable_service.ex
```diff
@@ -0,0 +1,23 @@
...
+  """
+  import SmartCity.Event, only: [dataset_disable: 0]
+
+  @doc """
+  Disable a dataset
+  """
+  @spec disable(term()) :: {:ok, SmartCity.Dataset.t()} | {:error, any()} | {:not_found, any()}
+  def disable(dataset_id) do
+    with {:ok, dataset} when not is_nil(dataset) <- Brook.get(:dataset, dataset_id),
+         :ok <- Brook.Event.send(dataset_disable(), :andi, dataset) do
```
jeffgrunewald: we pretty sure we want to send the whole dataset?

apps/andi/lib/andi/services/dataset_disable_service.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Andi.Services.DatasetDisableService do
```
jeffgrunewald: if we envision this being a recurring pattern for grouping code, would it make sense to just name this module `Andi.Services.DatasetDisable`? or `Andi.Services.DisableDataset` if you prefer.

apps/andi/lib/andi/services/dataset_disable_service.ex
JarredOlson: yes

PR #190, opened 2019-10-14
Mod date validate

kennyatpillar:


apps/andi/lib/andi/validation.ex
```diff
@@ -0,0 +1,27 @@
+defmodule DatasetValidator do
```
jdenen: This doesn't line up with the filename/directory convention. I'd expect it to be `Andi.Validation` right now.

apps/andi/lib/andi/validation.ex
```diff
@@ -0,0 +1,27 @@
...
+  @moduledoc false
```
jdenen: Can you document this module? Since it's going to potentially blow up dataset definitions, and we expect to extend it, I think we'll want it documented.

apps/andi/lib/andi_web/validators/dataset_validator.ex
```diff
@@ -32,4 +37,21 @@ defmodule AndiWeb.DatasetValidator do
 
   defp same_system_name(a, b), do: a.technical.systemName == b.technical.systemName
   defp different_ids(a, b), do: a.id != b.id
+
+  defp description_required do
```
jdenen: Why do we have this function twice?

kennyatpillar: That was us trying to merge in our change with @jakeprem's We have a bit of cleanup to do.

apps/andi/lib/andi/validation.ex
rajones-hntb: We deleted this file and moved it to the Validator/Dataset_Validation.ex

apps/andi/lib/andi/validation.ex
rajones-hntb: We deleted this file and moved the code to the dataset_validation.ex file

PR #191, opened 2019-10-14
Create tables for ingestible datasets only

jdenen:
We're currently creating tables for ALL datasets, even remote and hosted.

### DIFF
--- a/apps/forklift/test/unit/integration/event_handling_test.exs
+++ b/apps/forklift/test/unit/integration/event_handling_test.exs
@@ -12,7 +12,7 @@ defmodule Forklift.Integration.EventHandlingTest do
   setup :verify_on_exit!
 
   describe "on dataset:update event" do
-    test "ensures table exists for a dataset" do
+    test "ensures table exists for ingestible dataset" do
       test = self()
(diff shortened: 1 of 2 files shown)


apps/forklift/test/unit/integration/event_handling_test.exs
```diff
@@ -23,6 +23,12 @@ defmodule Forklift.Integration.EventHandlingTest do
       Brook.Event.send(instance_name(), dataset_update(), :author, dataset)
       assert_receive table: ^table_name, schema: ^schema
     end
+
+    test "does not create table for non-ingestible dataset" do
+      expect(MockTable, :init, 0, fn _ -> :ok end)
+      dataset = TDG.create_dataset(%{technical: %{sourceType: "remote"}})
```
bbrewer-papa-x: Is `ingestible` something we should define somewhere central? Or is it different for each app? LGTM and I don't think we should address that here. However, it has come up before.

jdenen: It's the same across the platform, so you're probably right.

bbrewer-papa-x: Good.

PR #203, opened 2019-10-16
Use @ to delimit app and version

jdenen:


jdenen:
It won't be used in the Docker tag. It's just part of the Git tag to delimit. I was planning to use `:` but that's not a valid Git tag character.

`forklift@1.0.0` -> `docker build -t smartcitiesdata/forklift:1.0.0 .`

PR #208, opened 2019-10-22
Update brook to 0.4.5

bbrewer-papa-x:
smartcitiesdata/andi#???


bbrewer-papa-x:
Confirmed this works as desired in live environment

PR #209, opened 2019-10-22
Update andi to the newest version of smart_city

jmorris-pillar:
smartcitiesdata/smartcitiesdata/smartcitiesdata#149

co-authored-by: Richard Jones <rajones@hntb.com>


apps/andi/mix.exs
```diff
@@ -51,7 +51,7 @@ defmodule Andi.MixProject do
       {:placebo, "~> 1.2", only: [:dev, :test, :integration]},
       {:plug_cowboy, "~> 2.1"},
       {:simply_validate, ">= 0.2.0"},
-      {:smart_city, "~> 3.0.0", override: true},
+      {:smart_city, "~> 3.0", override: true},
```
sukaant: The current version of smart_city is 3.2, shall we consider updating it here?

jmorris-pillar: It will use 3.2.1

PR #229, opened 2019-10-25
Fix integration test flakiness

jdenen:
I can't reproduce these locally, so it seems to be a problem when running on Travis VMs and whatever resources they're given.


LtChae:
Can we get this merged? It's blocking the builds of the other two PRs

Issue #207, opened 2019-10-22
Andi Creates Update User-Organization Association Events


NathanielScottStevens:
Andi will create `user:organization:associate` events so that the rest of the system can be notified.

AC:
* Provide an endpoint `POST`:`/api/v1/organization/{id}/users/add` that takes an array of user ids and an organization id
* Validate the organization
* Send `user:organization:associate` events for each user in the array

PR #232, opened 2019-10-25
Adds User Organization Association endpoint

NathanielScottStevens:
#207


apps/andi/lib/andi/event_handler.ex
```diff
@@ -12,4 +13,13 @@ defmodule Andi.EventHandler do
   def handle_event(%Brook.Event{type: organization_update(), data: %Organization{} = data}) do
     {:merge, :org, data.id, data}
   end
+
+  def handle_event(%Brook.Event{type: user_organization_associate(), data: %UserOrganizationAssociate{} = data}) do
```
ManApart:
```suggestion
  def handle_event(%Brook.Event{type: user_organization_associate(), data: %UserOrganizationAssociate{user_id: user_id, org_id: org_id} = data}) do
```

apps/andi/lib/andi/event_handler.ex
```diff
@@ -12,4 +13,13 @@ defmodule Andi.EventHandler do
...
+    %UserOrganizationAssociate{user_id: user_id, org_id: org_id} = data
```
ManApart:
```suggestion

```

Issue #218, opened 2019-10-24
Andi: Status page displays list of datasets


carriebrinkley:
#### Acceptance Criteria: 

- [ ] There is a list of all datasets on the page including dataset name and date modified
- [ ] @jakeclark to provide comps/input on column headers, etc.
- [ ] Handle no data, error, and loading 

#### Amigos: 

@jakeclark, @NathanielScottStevens, @jakeprem 

#### Tech Notes: 

- This will be done as an effort to validate that Phoenix will work going forward
- Datasets as links is out of scope
- Make sure the browser pipeline is correct

jakeprem:
##### Questions for entry:

  * What is the date modified we want to be using here? As far as I can see, we only have modifiedDate currently, which should be the last date the dataset was modified (by the source?) not the last time it was modified in our system.

PR #236, opened 2019-10-28
Add support for server rendered pages to Andi

jakeprem:
co-authored-by: Sukaant Chaudhary<sukaant.chaudhary@gmail.com>

Todo:
- [ ] Movie API Controllers to API namespace? (`AndiWeb.API.DatasetController`)
- [ ] Should `dev.exs` import `integration.exs` and layer live reloading code over top of it
- [x] Fix CSS
- [x] Live reload controllers too?

#218


apps/andi/priv/static/css/app.css
```diff
@@ -0,0 +1,137 @@
+/* Includes some default style for the starter application.
```
NathanielScottStevens: This css file appears to be the same as phoenix.css. You're probably already thinking about this but, I think we'll want to wipe out this file with the exception perhaps of Milligram (Though it's odd to me that Milligram was just copied into here instead of being imported as a dependency)

NathanielScottStevens:
The priv directory got committed by accident

apps/andi/mix.exs
```diff
@@ -46,7 +46,7 @@ defmodule Andi.MixProject do
       {:sobelow, "~> 0.8", only: :dev},
       {:phoenix, "~> 1.4"},
       {:phoenix_html, "~> 2.13"},
-      {:phoenix_live_reload, "~> 1.2", only: :dev},
+      {:phoenix_live_reload, "~> 1.2", only: [:dev, :integration]},
```
mcsearchin: Didn't we remove this requirement from this first card?  Is there something in this view that requires live view?

apps/andi/priv/static/css/app.css
```diff
@@ -0,0 +1,137 @@
...
...om:inherit;padding:0 1.0rem}}a{color:#0069d9;text-decoration:none}a:focus,a:hover{color:#606c76}dl,ol,ul{list-style:none;margin-top:0;padding-left:0}dl dl,dl ol,dl ul,ol dl,ol ol,ol ul,ul dl,ul ol,ul ul{font-size:90%;margin:1.5rem 0 1.5rem 3.0rem}ol{list-style:decimal inside}ul{list-style:circle inside}.button,button,dd,dt,li{margin-bottom:1.0rem}fieldset,input,select,textarea{margin-bottom:1.5rem}blockquote,dl,figure,form,ol,p,pre,table,ul{margin-bottom:2.5rem}table{border-spacing:0;width:100%}td,th{border-bottom:0.1rem solid #e1e1e1;padding:1.2rem 1.5rem;text-align:left}td:first-child,th:first-child{padding-left:0}td:last-child,th:last-child{padding-right:0}b,strong{font-weight:bold}p{margin-top:0}h1,h2,h3,h4,h5,h6{font-weight:300;letter-spacing:-.1rem;margin-bottom:2.0rem;margin-top:0}h1{font-size:4.6rem;line-height:1.2}h2{font-size:3.6rem;line-height:1.25}h3{font-size:2.8rem;line-height:1.3}h4{font-size:2.2rem;letter-spacing:-.08rem;line-height:1.35}h5{font-size:1.8rem;letter-spacing:-.05rem;line-height:1.5}h6{font-size:1.6rem;letter-spacing:0;line-height:1.4}img{max-width:100%}.clearfix:after{clear:both;content:' ';display:table}.float-left{float:left}.float-right{float:right}
```
JarredOlson: What is this line? Can it be standard CSS?

apps/andi/priv/static/js/app.js
```diff
@@ -0,0 +1,120 @@
+/******/ (function(modules) { // webpackBootstrap
```
JarredOlson: This appears to be a generated file, but it made me think about have we given thought to how we would TDD JavaScript inside of a phoenix app?

apps/andi/assets/webpack.config.js
```diff
@@ -0,0 +1,41 @@
+const path = require('path');
```
JarredOlson: Have we thought about how we would cache bust css and JavaScript?

apps/andi/priv/static/css/app.css
bbrewer-papa-x: Please do find a way to not check in `priv/static`. Thank you

apps/andi/assets/css/app.scss
```diff
@@ -0,0 +1,49 @@
+/* This file is for your main application css. */
```
ManApart:
```suggestion

```

apps/andi/assets/js/app.js
```diff
@@ -0,0 +1,18 @@
...
+// its own CSS file.
+import normalize_css from 'normalize.css'
+import scss from "../css/app.scss"
+
+// webpack automatically bundles all modules in your
+// entry points. Those entry points can be configured
+// in "webpack.config.js".
+//
+// Import dependencies
+//
```
ManApart: Feels like we can delete most of these comments.

apps/andi/assets/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
ManApart: Can we do this at the top level .gitignore? If not, what is the reason we need to be under `andi/assets`?

apps/andi/mix.exs
ManApart: I'd like to know the answer to this as well; I would appreciate a documented response in case others are also curious.

jakeprem: Live reload is a dev tool, LiveView is something different.

https://github.com/phoenixframework/phoenix_live_reload

apps/andi/lib/andi_web/controllers/dataset_page_controller.ex
```diff
@@ -0,0 +1,15 @@
+defmodule AndiWeb.DatasetPageController do
+  use AndiWeb, :controller
+
+  import Andi, only: [instance_name: 0]
```
LtChae: Do we still need this?

apps/andi/assets/css/app.scss
ManApart: Resolved without change or comment??

apps/andi/assets/.gitignore
ManApart: If we can't do this I'd at least appreciate an explanation.

apps/andi/assets/js/app.js
ManApart: Again, I'm disappointed to see all of these conversations resolved without any response or explanation.

apps/andi/assets/css/app.scss
jakeprem: These comments are generated by Phoenix. We will be leaving them to help with onboarding the team to our new use of Phoenix server generated HTML.

apps/andi/lib/andi_web/plugs/redirect.ex
```diff
@@ -0,0 +1,15 @@
+defmodule AndiWeb.Redirect do
+  @moduledoc """
+  A Plug to allow for easily doing redirects within a Plug or Phoenix router.
+
+  Based on code found at:
+    https://www.viget.com/articles/how-to-redirect-from-the-phoenix-router/
+  """
+  def init([to: _] = opts), do: opts
+  def init(_default), do: raise("Missing required option ':to' in redirect")
```
jdenen:
```elixir
  def init(opts), do: Keyword.fetch!(opts, :to)
```
While this won't give you the custom error message, this is the more typical way to ensure a keyword argument is passed.

apps/andi/lib/andi/services/dataset_retrieval_service.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Andi.Services.DatasetRetrieval do
+  @moduledoc """
+  Interface for retrieving datasets.
+  """
+  import Andi, only: [instance_name: 0]
+
+  def get_all(instance \\ instance_name()) do
```
jdenen:
```elixir
def get_all do
  Brook.get_all_values(instance_name(), :dataset)
end
```

Why are we making `instance_name` a macro to be included in function heads as a default that callers always use anyway? It seems unnecessary.

apps/andi/lib/andi_web/plugs/redirect.ex
mcsearchin: Okay - we'll break this into a function that checks for the existence of the key.
```
if Keyword.has_key?(opts) do
  opts
else
 raise "error"
end
```

apps/andi/lib/andi/services/dataset_retrieval_service.ex
mcsearchin: This made it easier to isolate the brook instance we're using for testing.  Using the Andi instance_name in all of our tests modified view state across tests.  By allowing to optionally pass in the instance, we keep that state from being littered across tests that care about this view state.

LtChae: I generally agree that a macro is more optimal from a runtime standpoint. I would honestly love it if Brook gave us the instance name for free as a fallback default via `:application.get_application(__MODULE__)`

apps/andi/lib/andi_web/controllers/dataset_page_controller.ex
```diff
@@ -0,0 +1,11 @@
+defmodule AndiWeb.DatasetPageController do
```
NathanielScottStevens: Not sure if I agree with calling this page `Dataset`. I was anticipating it being called `Status` or something like that. I mostly just want to avoid the proliferation of files with dataset in the name that we have in React. Were you imagining this controller would be responsible for other templates such as the dataset entry form as well?

PR #260, opened 2019-11-06
Fix System.get_env call to be  compatible with elixir 1.8

bbalser:


jakeprem:
Did Elixir 1.8 remove the ability to have a default there?

Issue #219, opened 2019-10-24
Search bar filters dataset names


carriebrinkley:
#### Acceptance Criteria: 

- [x] Search bar exists 
- [x] Datasets are filtered by name containing exact match of search term 
- [x] The search criteria is maintained in the url

#### Amigos: 

@jakeclark , @NathanielScottStevens 

#### Tech Notes: 

- Enter will trigger the search
- Consider adding Live View

PR #261, opened 2019-11-06
An initial draft of implementing live_view

jakeprem:
Here's an initial take on adding live view. 

I wrote some pretty unfriendly code here, especially in the templates that I expect to seriously refactor before merging this. I'm still trying to find the best way to break up the templates into more manageable chunks.

#219


apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -0,0 +1,46 @@
+defmodule AndiWeb.DatasetLiveView do
+  use Phoenix.LiveView
+
+  def render(assigns) do
```
jakeprem: This doesn't have to go here. 

```elixir
def render(assigns) do
  AndiWeb.DatasetView.render("index.html")
end
```
That code should render it from a separate template file.

jakeprem:
Currently having an issue with the host when running in dev.

Here's the error that Phoenix is logging in dev:
```
Origin of the request: https://andi.dev.internal.smartcolumbusos.com

This happens when you are attempting a socket connection to
a different host than the one configured in your config/
files. For example, in development the host is configured
to "localhost" but you may be trying to access it from
"127.0.0.1". To fix this issue, you may either:

  1. update [url: [host: ...]] to your actual host in the
     config file for your current environment (recommended)

  2. pass the :check_origin option when configuring your
     endpoint or when configuring the transport in your
     UserSocket module, explicitly outlining which origins
     are allowed:

        check_origin: ["https://example.com",
                       "//another.com:888", "//other.com"]
```

apps/andi/lib/andi_web/live/dataset_live_view.ex
NathanielScottStevens: You should be able to just put it in a .leex file and have the render function be generated, right?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -0,0 +1,92 @@
...
+    {:ok, assign(socket, datasets: nil, search_text: "")}
+  end
+
+  def handle_params(params, _uri, socket) do
+    search_text = Map.get(params, "search-value", "")
+    socket_pid = self()
+
+    Task.start(fn ->
+      # Process.sleep(1000)
+      datasets = Andi.Services.DatasetRetrieval.get_all!()
```
NathanielScottStevens: Ultimately, I would rather we keep a list of all datasets in the state and update it through a PubSub. But I don't know that we need to go that far for this card. Honestly, I don't even know that I'd worry about blocking the render atm.

apps/andi/lib/andi_web/live/dataset_live_view.ex
jakeprem: We're going to use LiveComponents to break this up, but still keep the template in the files.

apps/andi/assets/css/datasets.scss
```diff
@@ -6,6 +6,10 @@
   font-size: 1.5em;
 }
 
+.datasets-search {
+  margin: 0.5em;
```
mcsearchin: We're going to make these all rems, but Brewer will clean up any of the old ones on his card.

apps/andi/config/dev.exs
```diff
@@ -19,6 +19,9 @@ config :andi, AndiWeb.Endpoint,
       "--watch-stdin",
       cd: Path.expand("../assets", __DIR__)
     ]
+  ],
+  live_view: [
+    signing_salt: "TOP SECRET!!!"
```
mcsearchin: Should we just put this value in `config.exs` and remove it from these environment-specific files? It looks like the actual value isn't really important to us yet.

jakeprem: I don't think so because it needs to be set in each environment, but only in Prod does it need to be a secret

apps/andi/assets/css/datasets.scss
jakeprem: Is REM going to be properly responsive to different screen sizes/resolutions?

apps/andi/lib/andi/dataset_cache.ex
```diff
@@ -0,0 +1,40 @@
...
+    # Warning: Be extremely careful using :public for ETS tables. This can lead to race conditions and all kinds of bad things.
+    # In this case Brook is already single threaded so it should be ok.
+    pid = :ets.new(__MODULE__, [:set, :public, :named_table])
+
+    DatasetRetrieval.get_all!() |> put_datasets()
+
+    {:ok, pid}
+  end
+
+  def handle_call(:reset, _from, _state) do
```
mcsearchin: I would expect that since we're doing the client callback pattern here that there would be a client function for this. Though I don't think that's important at the moment.

jakeprem: That's the usual pattern, but in this case we only intend for this functionality to be triggered by our tests, so we didn't implement a client function to help discourage resetting outside the tests.

apps/andi/assets/css/datasets.scss
jakeprem: Advice from https://zellwk.com/blog/rem-vs-em/:

1. Size in em if the property scales according to its font-size
2. Size everything else in rem.

So it seems like rems are a good choice here.

apps/andi/mix.exs
```diff
@@ -39,12 +39,14 @@ defmodule Andi.MixProject do
       {:divo, "~> 1.1", only: [:dev, :integration]},
       {:divo_kafka, "~> 0.1.5", only: [:dev, :integration]},
       {:divo_redis, "~> 0.1.4", only: [:dev, :integration]},
+      {:floki, "~> 0.23"},
```
mcsearchin: Should this be scoped to `[:test, :integration]` ?

apps/andi/test/unit/andi_web/live/dataset_live_view_test.exs
```diff
@@ -0,0 +1,176 @@
...
+
+  describe "Basic live page load" do
+    test "loads all datasets", %{conn: conn} do
+      datasets = Enum.map(1..3, fn _x -> TDG.create_dataset([]) end)
+
+      DatasetCache.put_datasets(datasets)
+
+      assert {:ok, _view, html} = live(conn, @url_path)
+
+      table_text = floki_get_text(html, ".datasets-index__table")
```
mcsearchin: 💯

apps/andi/test/unit/andi_web/live/dataset_live_view_test.exs
```diff
@@ -0,0 +1,176 @@
...
+
+      Enum.each(datasets, fn dataset ->
+        assert table_text =~ dataset.business.dataTitle
+      end)
+    end
+
+    test "shows No Datasets when there are no rows to show", %{conn: conn} do
+      assert {:ok, view, html} = live(conn, @url_path)
+
+      assert html =~ "All Datasets"
```
mcsearchin: *This assertion* in *this test* confuses me.

jakeprem: Me too.

We keep that "All Datasets" header up above the table at all times.

This test was initially just verifying that the live view loads and we didn't take it out.

apps/andi/mix.exs
jakeprem: Floki is pretty small and doesn't have any running components, so I think it should be fine to leave here. Do we have much of a pattern either way?

apps/andi/test/unit/andi_web/live/dataset_live_view_test.exs
jakeprem: I updated the assertions to be more specific.

apps/andi/mix.exs
bbrewer-papa-x: As far as I know we've been pretty diligent about making sure things like `placebo`, `checkov`, `smartcities_test`, etc. are test and integration.

jakeprem: I scoped it to dev, testing, and integration

apps/andi/config/integration.exs
```diff
@@ -71,7 +71,11 @@ config :andi, AndiWeb.Endpoint,
       ~r{priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$},
       ~r{priv/gettext/.*(po)$},
       ~r{lib/andi_web/controllers/.*(ex)$},
+      ~r{lib/andi_web/live/.*(ex)$},
       ~r{lib/andi_web/views/.*(ex)$},
       ~r{lib/andi_web/templates/.*(eex)$}
     ]
+  ],
+  live_view: [
+    signing_salt: "SUPER VERY TOP SECRET!!!"
```
jmorris-pillar: As part of this card can you override this in runtime.exs and make the chart changes, even if not in this PR?

jmorris-pillar: Sorry i meant to leave this on prod.exs

jakeprem: We're writing a bug card for that

apps/andi/lib/andi_web.ex
```diff
@@ -37,6 +39,8 @@ defmodule AndiWeb do
       # Import convenience functions from controllers
       import Phoenix.Controller, only: [get_flash: 1, get_flash: 2, view_module: 1]
 
+      import Phoenix.LiveView, only: [live_render: 2, live_render: 3]
+
```
jmorris-pillar: are these different than the controller methods?

PR #263, opened 2019-11-07
Validate modified dates

ManApart:


apps/andi/test/unit/andi_web/validators/dataset_validator_test.exs
```diff
@@ -15,6 +15,8 @@ defmodule AndiWeb.DatasetValidatorTest do
           technical: %{orgName: "some-cool-data"},
           business: %{description: "something", modifiedDate: "2019-10-14T17:30:16Z"}
         )
+        |> Jason.encode!()
```
kennyatpillar: Wrap this with a private function

PR #265, opened 2019-11-08
Reaper

bbalser:


apps/reaper/mix.lock
```diff
@@ -0,0 +1,96 @@
+%{
```
ManApart: I believe you can drop this file

ManApart: This is jessie btw lol

mcsearchin:
It looks like this is just moving the files over with no other changes sans configuration. Is that correct?

jdenen:
That's what I see @mcsearchin 

That's what he did for `valkyrie` too.

ManApart:
Even when I don't review a PR, I review that PR. lol

PR #269, opened 2019-11-12
Modified Date Migration

ManApart:


apps/andi/lib/andi/migration/date_coercer.ex
```diff
@@ -0,0 +1,38 @@
...
+    formatted_dates =
+      Enum.map(format_strings, fn format ->
+        Timex.parse(date, format, :strftime)
+      end)
+
+    ok_dates = for {:ok, date} <- formatted_dates, do: date
+
+    if length(ok_dates) > 0 do
+      ok_dates
+      |> List.first()
```
jdenen: You could replace this with `Enum.reduce_while` and halt on the first successfully converted format.

ManApart: Good call. We've adjusted accordingly.

apps/andi/lib/andi/migration/migrations.ex
```diff
@@ -0,0 +1,30 @@
...
+
+    {:ok, :ok, {:continue, :stop}}
+  end
+
+  def handle_continue(:stop, state) do
+    {:stop, :normal, state}
+  end
+
+  def migrate_modified_dates do
+    if is_nil(Brook.get!(@instance, :migration, "modified_date_migration_completed")) do
```
jdenen: `if` block seems like a smell to me. Could you move the `Brook.get!` call to `init` and pipe it to `migrate_modified_dates` to guard on `nil`?

```elixir
def inti(_) do
  Brook.get!(@instance, :migration, "modified_date_migration_completed")
  |> migrate_modified_dates()

  {:ok, :ok, {:continue, :stop}}
end

defp migrate_modified_dates(dates) when is_nil(dates) do
  Brook.Event.send(@instance, "migration:modified_dates", :andi, %{})
end

defp migrate_modified_dates(_), do: nil
```

apps/andi/lib/andi/migration/modified_date_migration.ex
```diff
@@ -0,0 +1,50 @@
...
+      Brook.ViewState.merge(:dataset, updated_dataset.id, updated_dataset)
+    end
+  end
+
+  defp migrate_dataset(invalid_dataset) do
+    Logger.warn("Could not migrate invalid dataset #{inspect(invalid_dataset)}")
+  end
+
+  defp log_bad_date(id, old_date, new_date) do
+    if new_date == "" && old_date != "" do
```
jdenen: Pattern match out this `if` condition.

apps/andi/lib/andi/migration/modified_date_migration.ex
```diff
@@ -0,0 +1,50 @@
...
+        dataset.business
+        |> Map.from_struct()
+        |> Map.put(:modifiedDate, corrected_date)
+
+      {:ok, updated_dataset} =
+        dataset
+        |> Map.from_struct()
+        |> Map.put(:technical, Map.from_struct(dataset.technical))
+        |> Map.put(:business, updated_business)
+        |> SmartCity.Dataset.new()
```
jdenen: This function is rather large, and it seems like this code could be pulled out into its own function or two.

apps/andi/lib/andi/migration/migrations.ex
ManApart: I'd like init to be just a list of migration function calls so you can scan the init and know what all migrations are being run without having to understand details.
```
 def init(_args) do
    migrate_modified_dates()
    migrate_skyrim_mods()
    migrate_friends()

    {:ok, :ok, {:continue, :stop}}
  end
```
The counter proposal I'd make is below, though I think I'd just prefer the if at that point.
```
def migrate_modified_dates do
    Brook.get!(@instance, :migration, "modified_date_migration_completed")
    |> migrate_modified_dates()
  end

  defp migrate_modified_dates(nil), do: Brook.Event.send(@instance, "migration:modified_dates", :andi, %{})

  defp migrate_modified_dates(_), do: nil
```

apps/andi/lib/andi/migration/modified_date_migration.ex
ManApart: Good call

apps/andi/lib/andi/migration/modified_date_migration.ex
ManApart: We've pulled it out. We'd like to perhaps create some helpers in smart_city so that we can more easily update a field on those structs without having to go through this (or implement the behaviors).

apps/andi/lib/andi/migration/migrations.ex
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def migrate_modified_dates do
+    has_migration_run?("modified_date_migration_completed")
+    |> migrate_modified_dates()
+  end
+
+  defp has_migration_run?(migration_name) do
+    Brook.get!(@instance, :migration, migration_name)
+  end
```
jdenen: I would expect this function to return `true|false` because it ends with `?`.

jdenen: How about `get_migration`? If it returns `nil`, then it hasn't run a migration and thus you'll need to run it.

apps/andi/lib/andi/migration/migrations.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  defp migrate_modified_dates(nil), do: Brook.Event.send(@instance, "migration:modified_dates", :andi, %{})
```
jdenen: Pattern matching on `nil` is much better than my proposed guard against `nil`. 👍

apps/andi/lib/andi/migration/date_coercer.ex
```diff
@@ -0,0 +1,40 @@
...
+    case DateTime.from_iso8601(date) do
+      {:ok, _parsed_date, _offset} -> date
+      _ -> fix_date(date)
+    end
+  end
+
+  def fix_date(""), do: ""
+
+  def fix_date(date) do
+    format_strings = [
```
cameronmarsh: Did we decide to not check for d/m/y dates?

apps/andi/lib/andi/migration/modified_date_migration.ex
```diff
@@ -0,0 +1,56 @@
...
+      Brook.Event.send(@instance, dataset_update(), :andi, updated_dataset)
+      Brook.ViewState.merge(:dataset, updated_dataset.id, updated_dataset)
+    end
+  end
+
+  defp migrate_dataset(invalid_dataset) do
+    Logger.warn("Could not migrate invalid dataset #{inspect(invalid_dataset)}")
+  end
+
+  defp log_bad_date(id, old_date, "") when old_date != "" do
```
jdenen: 👍

apps/andi/lib/andi/application.ex
```diff
@@ -7,7 +7,8 @@ defmodule Andi.Application do
     children = [
       AndiWeb.Endpoint,
       {Brook, Application.get_env(:andi, :brook)},
-      Andi.DatasetCache
+      Andi.DatasetCache,
+      Andi.Migration.Migrations
```
jmorris-pillar: Do you want this to happen AFTER the cache is established?  Probably doesnt matter but just asking the question

ManApart: We want the app to start up first and then run all migrations in its own time.

apps/andi/lib/andi/migration/date_coercer.ex
ManApart: Added.

apps/andi/lib/andi/migration/date_coercer.ex
```diff
@@ -0,0 +1,44 @@
...
+  Either parse to valid iso8601 or convert to empty string.
+  """
+  def coerce_date(date) do
+    case DateTime.from_iso8601(date) do
+      {:ok, _parsed_date, _offset} -> date
+      _ -> fix_date(date)
+    end
+  end
+
+  def fix_date(""), do: ""
```
NathanielScottStevens: Privatize

PR #271, opened 2019-11-12
Order datasets 

cameronmarsh:
smartcolumbusos/monitoring#85


apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -4,9 +4,11 @@ defmodule AndiWeb.DatasetLiveView do
   alias AndiWeb.DatasetLiveView.Table
 
   def render(assigns) do
+    IO.inspect(assigns.search_text, label: "in the render")
+
     ~L"""
     <div class="datasets-index">
-      <h1 class="datasets-index__title">All Datasets</h1>
+      <h1 id="stuff" class="datasets-index__title">All Datasets</h1>
```
ManApart: lol

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -22,24 +24,48 @@ defmodule AndiWeb.DatasetLiveView do
...
 
   def mount(_session, socket) do
-    {:ok, assign(socket, datasets: nil, search_text: "")}
+    {:ok, assign(socket, datasets: nil, search_text: nil, order: {"data-title", "asc"}, params: %{})}
+  end
+
+  def handle_info({:order, field}, socket) do
+    order_dir =
+      case socket.assigns.order do
+        {^field, "asc"} -> "desc"
```
ManApart: Why do we need to pin the field?

apps/andi/assets/css/app.scss
```diff
@@ -12,7 +12,7 @@ body {
   flex-direction: column;
 
   font-family: Arial, Helvetica, sans-serif;
-  font-size: 0.75em;
+  font-size: 1em;
```
jmorris-pillar: Rems vs ems?

apps/andi/assets/css/datasets.scss
```diff
@@ -62,3 +61,13 @@
...
 }
+
+.datasets-table__th--asc{
+    -webkit-box-shadow: inset 0 0.2em 0 0 $fadedblack;
+    box-shadow: inset 0 0.2em 0 0 $fadedblack;
+}
+
+.datasets-table__th--desc {
+    -webkit-box-shadow: inset 0 -3px 0 0 $fadedblack;
+    box-shadow: inset 0 -3px 0 0 $fadedblack;
```
jmorris-pillar: this one is px but the one above is em?  Should they both be rems or ems?

apps/andi/lib/andi_web/live/dataset_live_view.ex
jeffgrunewald: this is to allow it to be an explicit pattern match against whatever comes out of the `case` expression. Without the pin, the `field` variable is rebound to whatever comes out of that expression and that first branch will always evaluate to true and execute (the `"desc"` branch).

apps/andi/assets/css/variables.scss
```diff
@@ -1,5 +1,6 @@
 $lightgray: lightgray;
 $darkgray: darkgray;
+$fadedblack: rgba(0, 0, 0, 0.6);
```
ManApart: Chic

apps/andi/assets/css/app.scss
mcsearchin: I think we normed on moving to rems over ems, but nobody actually made the change yet.  Do y'all mind managing that find/replace?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -25,27 +25,53 @@ defmodule AndiWeb.DatasetLiveView do
...
 
   def mount(_session, socket) do
-    {:ok, assign(socket, datasets: nil, search_text: "")}
+    {:ok, assign(socket, datasets: nil, search_text: nil, order: {"data-title", "asc"}, params: %{})}
+  end
+
+  def handle_info({:order, field}, socket) do
+    order_dir =
+      case socket.assigns.order do
+        %{^field => "asc"} -> "desc"
```
ManApart: Still curious about this guy.

apps/andi/assets/css/variables.scss
mcsearchin: Is there a non-color name for this color?  Something like `header-shadow`?  I'd like to do the same with the grays/greys, but I know you didn't add those.

apps/andi/lib/andi_web/live/dataset_live_view.ex
ManApart: Ah, I was thinking you would just pin if you didn't want field to be changed. I didn't think about how it'd make a more forgiving pattern match for the case statement. Thanks

apps/andi/lib/andi_web/live/dataset_live_view.ex
ManApart: Nevermind. I didn't refresh and see Jeff's explination.

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -57,4 +83,19 @@ defmodule AndiWeb.DatasetLiveView do
...
+    case order_dir do
+      "asc" -> Enum.sort_by(datasets, fn dataset -> Map.get(dataset, order_by) end)
+      "desc" -> Enum.sort_by(datasets, fn dataset -> Map.get(dataset, order_by) end, &>=/2)
+      _ -> datasets
+    end
+  end
+
+  defp to_view_model(dataset) do
+    %{
+      "org-title" => dataset.business.orgTitle,
```
jmorris-pillar: is kabob case the norm here?  Normally we expect elixir vars to be snake case?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -25,27 +25,53 @@ defmodule AndiWeb.DatasetLiveView do
...
     search_text = Map.get(params, "search-value", "")
-    datasets = Andi.DatasetCache.get_datasets()
 
-    {:noreply, assign(socket, search_text: search_text, datasets: filter_datasets(datasets, search_text))}
+    datasets =
+      filter_on_search_change(search_text, socket)
+      |> sort_by_dir(order_by, order_dir)
+
+    {:noreply,
+     assign(socket, search_text: search_text, datasets: datasets, order: %{order_by => order_dir}, params: params)}
```
bbrewer-papa-x: Did we confirm that the order part is not additive? That is, if we change the column that the other one doesn't hang around in the map? I'm 90% sure it kills other entries (like it should), but just double checking

apps/andi/lib/andi_web/live/dataset_live_view.ex
bbrewer-papa-x: These are the user facing ones. I'm open to changing it, or maybe mixing a little of both in?

apps/andi/lib/andi_web/live/dataset_live_view.ex
cameronmarsh: Yep, otherwise the css indicator for sorting would show on both columns when you switch between them

apps/andi/assets/css/variables.scss
cameronmarsh: We can fix that

apps/andi/assets/css/datasets.scss
cameronmarsh: reeeeeeeems

apps/andi/assets/css/variables.scss
cameronmarsh: Merci beaucoup!

apps/andi/assets/css/app.scss
cameronmarsh: no problemo

apps/andi/lib/andi_web/live/dataset_live_view.ex
bbrewer-papa-x: I can't see Jeff's explanation

PR #318, opened 2019-11-14
Change cursor on sortable col and update search query param

cameronmarsh:
smartcitiesdata/smartcitiesdata#220

co-authored-by: Kenny Drobnack <kdrobnack@pillartechnology.com>


jakeprem:
:fire:

PR #323, opened 2019-11-15
Xml schema validation

LtChae:


apps/andi/lib/andi_web/validators/dataset_schema_validator.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  def validate_schema(_), do: []
+
+  defp validate_schema_has_selectors(schema) do
+    results =
+      schema
+      |> Enum.map(&build_field_validator/1)
+      |> List.flatten()
+
+    results
```
sukaant: Is "results" required here?

LtChae: Good call.  We can just return the result of the functions.

apps/andi/lib/andi_web/validators/dataset_schema_validator.ex
```diff
@@ -0,0 +1,53 @@
+defmodule AndiWeb.DatasetSchemaValidator do
+  @moduledoc """
+  Used to validate dataset schemas
+  """
+
+  def validate(%{"technical" => %{"sourceType" => source_type}} = dataset) when source_type in ["ingest", "stream"] do
+    SimplyValidate.validate(dataset, [validate_schema_required()]) ++ validate_schema(dataset)
+  end
+
+  def validate(_), do: []
```
ManApart: So if a definition isn't ingest or stream, or has no source type or technical, we just mark it valid? I'm guessing because xml is the first time we actually validate a schema?

apps/andi/lib/andi_web/validators/dataset_validator.ex
```diff
@@ -67,4 +86,25 @@ defmodule AndiWeb.DatasetValidator do
         false
     end
   end
+
+  # Handles preconverting datasets from structs to maps for comparison purposes
```
ManApart: I was going to ask why we're doing this, but this comment meant I didn't need to. Thanks!

apps/andi/lib/andi_web/validators/dataset_schema_validator.ex
LtChae: Correct. It's likely that we'll need to hang more schema validation off of this in the future but for now this is it.

PR #330, opened 2019-11-18
Status indicator

ManApart:


bbrewer-papa-x:
Please fix your commit history.!

apps/andi/lib/andi/dataset_cache.ex
```diff
@@ -4,26 +4,46 @@ defmodule Andi.DatasetCache do
...
 
-  def put_dataset(invalid_dataset) do
+  def put(invalid_dataset) do
     Logger.warn("Not caching dataset because it is invalid: #{inspect(invalid_dataset)}")
   end
 
-  def get_datasets do
+  defp get(id) do
+    case :ets.match(__MODULE__, {id, :"$1"}) do
+      [[h | _inner_t] | _t] -> h
```
jakeprem: What is this line of code doing?

apps/andi/lib/andi/dataset_cache.ex
```diff
@@ -4,26 +4,46 @@ defmodule Andi.DatasetCache do
...
+  def put(%SmartCity.Dataset{} = dataset) do
+    updated =
+      dataset.id
+      |> get()
+      |> Map.merge(%{"id" => dataset.id, "dataset" => dataset})
+
+    :ets.insert(__MODULE__, {dataset.id, updated})
+  end
+
+  def put(%{"id" => id, "ingested_time" => time_stamp}) do
```
jakeprem: This feels like it should be a separate function rather than just a function head on put.

apps/andi/lib/andi/dataset_cache.ex
```diff
@@ -33,7 +53,8 @@ defmodule Andi.DatasetCache do
     # In this case Brook is already single threaded so it should be ok.
     pid = :ets.new(__MODULE__, [:set, :public, :named_table])
 
-    DatasetRetrieval.get_all!() |> put_datasets()
+    Brook.get_all_values!(instance_name(), :dataset) |> put()
```
jakeprem: Why did you go away from using a service here?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -3,6 +3,8 @@ defmodule AndiWeb.DatasetLiveView do
   alias AndiWeb.Router.Helpers, as: Routes
   alias AndiWeb.DatasetLiveView.Table
 
+  @ingested_time_topic "ingested_time_topic"
```
jakeprem: This is hardcoded in two places. Should it be set in configuration or maybe as a function on andi.ex?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -46,56 +49,92 @@ defmodule AndiWeb.DatasetLiveView do
...
+          |> Enum.at(exisiting_index)
+          |> Map.put("ingested_time", ingested_time)
+
+        List.replace_at(datasets, exisiting_index, updated_dataset)
     end
   end
 
-  defp to_view_model(dataset) do
+  defp to_view_model(model) do
     %{
```
jakeprem: I think the View Model should be a struct

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -46,56 +49,92 @@ defmodule AndiWeb.DatasetLiveView do
...
-      "desc" -> Enum.sort_by(datasets, fn dataset -> Map.get(dataset, order_by) end, &>=/2)
-      _ -> datasets
+      "asc" -> Enum.sort_by(models, fn model -> Map.get(model, order_by) end)
+      "desc" -> Enum.sort_by(models, fn model -> Map.get(model, order_by) end, &>=/2)
+      _ -> models
+    end
+  end
+
+  defp update_ingest_time(id, ingested_time, datasets) do
+    exisiting_index = Enum.find_index(datasets, fn dataset -> id == dataset["id"] end)
```
jakeprem: Doing live updates of the dataset is a good reason to starting storing the datasets as a map with `%{id: dataset}`. Then you can do updates just by doing a `Map.update`

apps/andi/lib/andi/dataset_cache.ex
jakeprem: Also if you're trying to get just one thing, you might look at `:ets.match_object`. It should return a list of the tuples you put into the table.

apps/andi/lib/andi_web/live/dataset_live_view.ex
rajones-hntb: We considered that but wanted to retain the order of the list.

apps/andi/lib/andi_web/live/dataset_live_view.ex
rajones-hntb: At this point I'm happy with keeping it a map.

apps/andi/lib/andi_web/live/dataset_live_view.ex
rajones-hntb: If we wanted it to be configurable then I'd agree, but at this point I don't see the value/jsr.

apps/andi/lib/andi/dataset_cache.ex
rajones-hntb: Keep the same level of abstraction between the two lines. Doesn't make sense to directly use brook in one line and then use a wrapper service in the next line.

apps/andi/lib/andi/dataset_cache.ex
rajones-hntb: These are all putting to the cache with different structures (but the same result) and so different function heads with pattern matching makes the most sense to me.

apps/andi/lib/andi/dataset_cache.ex
rajones-hntb: Good catch; we've updated it.

jdenen:
Squash out those wip commits

ManApart:
@jdenen You better believe it!

apps/andi/lib/andi/dataset_cache.ex
jakeprem: I think it would be better to have an `update_ingested_time` function.

You're also not "putting", you're fetching and existing one and modifying a field on it.

apps/andi/lib/andi/dataset_cache.ex
jakeprem: :+1: I feel like that suggests both calls to Brook should be in the service layer

apps/andi/lib/andi_web/live/dataset_live_view/table.ex
```diff
@@ -32,4 +36,11 @@ defmodule AndiWeb.DatasetLiveView.Table do
     send(self(), {:order, field})
     {:noreply, socket}
   end
+
+  defp ingest_status(dataset) do
+    case dataset["ingested_time"] do
```
jakeprem: This should be pulled out to a function codifies this logic as meaning a dataset is ingested rather than being implicit within the template.

bbrewer-papa-x:
> Squash out those wip commits

You're off the team Johnson

PR #331, opened 2019-11-19
Flair uses pipeline lib

jdenen:
## Pipeline.Reader.TopicReader
This is a generic Kafka topic reader. It's not dependent on dataset information. It ensures a topic exists and sets up `Elsa.Supervisor` with the correct group consumer configuration.

So it does all the things a `Pipeline.Reader.DatasetTopicReader` does, but takes a more generic set of arguments. It does the last bit (`Elsa.Supervisor` setup) in a new and better way. It relies on Elsa to register its supervisor using a via tuple. This is how `DatasetTopicReader` should have worked all along.

(I can say these things. I wrote `DatasetTopicReader`.)

For now, Flair is the only app that uses `TopicReader`.

A soon-to-follow PR will change `DatasetTopicReader` to use `TopicReader` and remove `Pipeline.Reader.DatasetTopicReader.InitTask`. 

**I could make the change in this PR**, but I'm guessing y'all would rather keep this small and Flair-focused.

## Flair
I made a few changes in Flair. The duration calculation logic has been namespaced in a more idiomatic way.

Flair uses `Pipeline.Reader.TopicReader` to setup Elsa and pass messages to `Flair.MessageHandler`. And Flair uses `Pipeline.Writer.TableWriter` to write to the stats table.

## e2e
The e2e tests now check that Flair has profiled ingested data messages. I removed Flair's integration test, because the e2e tests now cover it.


apps/pipeline/lib/pipeline/reader/topic_reader.ex
```diff
@@ -0,0 +1,109 @@
...
+          topic: String.t(),
+          handler: module()
+        ]
+
+  @impl Pipeline.Reader
+  @spec init(init_args()) :: :ok | {:error, term()}
+  @doc """
+  Sets up infrastructure to consume messages off a Kafka topic. Ensures the topic is created.
+
+  Optional arguments:
```
ManApart: I appreciate this documentation.

apps/flair/config/integration.exs
```diff
@@ -18,7 +17,7 @@ config :flair,
   divo_wait: [dwell: 1000, max_tries: 60],
   elsa_brokers: endpoint,
   message_processing_cadence: 5_000,
-  data_topic: "persisted",
+  data_topic: "streaming-persisted",
```
jeffgrunewald: is this a meaningful distinction to draw?

apps/flair/test/integration/test_helper.exs
```diff
@@ -1,23 +1 @@
 ExUnit.start()
```
jeffgrunewald: can we just get rid of the whole integration subdirectory?

apps/pipeline/lib/pipeline/reader/topic_reader.ex
```diff
@@ -0,0 +1,109 @@
...
+    end
+  end
+
+  @impl Pipeline.Reader
+  @spec terminate(instance: atom(), topic: String.t()) :: :ok | {:error, term()}
+  @doc """
+  Destroys topic consumer infrastructure.
+  """
+  def terminate(args) do
+    with name <- :"#{args[:instance]}-#{args[:topic]}-pipeline-supervisor",
```
jeffgrunewald: you've got a couple of different methods for grabbing elements from input args/opts going on throughout. in general you seem to be leaning on inlined options (dot accessing from a map, bracket accessing here, `Keyword.fetch` below) that will blow up if the value is missing but in this case you've got `nil` being returned.
any way we can true up the methods / expected return behavior from how these are all working?

apps/flair/config/integration.exs
ManApart: :fire:

jdenen: Forklift specifies `streaming-persisted` in its `config/integration.exs`. I could have changed that value to `persisted`, but I was try to touch as few apps as possible in this PR.

If everyone would prefer the integration/e2e tests use `persisted` or some other value, we can make that change. It doesn't matter IMO.

apps/flair/test/integration/test_helper.exs
jdenen: Not yet. `scripts/travis-test.sh` is going to go to each app and run `mix test.integration`. That'll blow up if the integration subdirectory doesn't exist.

I didn't want to touch CI as part of this PR, so I'm just leaving the subdirectory.

apps/pipeline/lib/pipeline/reader/topic_reader.ex
jdenen: Good points.

jdenen:
@jeffgrunewald 

Take another look. The keyword access comment led me to a slightly larger refactor. I like it a lot better now. Let me know what you think.

I replied to some comments too.

apps/flair/config/integration.exs
jeffgrunewald: i don't really care either beyond not wanting to propagate the perception from outsiders that it's a meaningful name.

apps/flair/test/integration/test_helper.exs
jeffgrunewald: fair point; as long as it's just skips over it when it doesn't find any files it doesn't bug me.

apps/flair/lib/flair/durations/init.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  def start_link(_) do
+    Task.start_link(__MODULE__, :run, [])
+  end
+
+  def run do
+    reader = bootstrap_reader()
+    writer = Flair.Durations.Consumer.bootstrap()
+
+    ^reader = writer
```
jeffgrunewald: these are both `:ok` or exception? interesting...

apps/flair/lib/flair/durations/consumer.ex
```diff
@@ -0,0 +1,73 @@
...
+  def bootstrap() do
+    case Application.get_env(:flair, :table_name_timing) do
+      nil -> :ok
+      _ -> @table_writer.init(table: @table_name, schema: @table_schema)
+    end
+  end
+
+  defp convert_events(events) do
+    events
+    |> Enum.map(fn {dataset_id, stats_map} ->
```
jeffgrunewald: is there any way to use a reduce on this to avoid the nested maps? is there any chance this operation will need to be faster than a map over a map operation can keep up with?

jdenen: Probably. I renamed this module and moved the file, but I didn't write this code. I didn't want to touch it as part of this PR.

apps/flair/config/integration.exs
jdenen: I get that. Maybe we change both values to `test-profiling` or something in the future?

apps/flair/lib/flair/durations/init.ex
jdenen: Yeah, and it returns `:ok` if both are `:ok`. In an intermediary commit, I had a complicated `with` statement that I really didn't like. This ended up being the simplest thing that I could think of.

ManApart: I think an explanatory comment would be nice unless you're planning on immediately refactoring.

PR #339, opened 2019-11-22
Reaper responds to file:ingest:end from odo and processes geojson ingest

rajones-hntb:


apps/reaper/lib/reaper/init.ex
```diff
@@ -17,6 +17,7 @@ defmodule Reaper.Init do
     |> Enum.each(fn dataset ->
       Logger.debug("Migrating extraction dataset #{dataset.id}")
       Reaper.Horde.Supervisor.start_data_extract(dataset)
+      Process.sleep(250)
```
ManApart: @JarredOlson Is this sleep still needed?

apps/reaper/lib/reaper/decoder/geojson.ex
```diff
@@ -28,7 +28,8 @@ defmodule Reaper.Decoder.GeoJson do
 
   @impl Reaper.Decoder
   def handle?(source_format) when is_binary(source_format) do
-    String.downcase(source_format) == "application/geo+json"
+    String.downcase(source_format) == "application/geo+json" ||
+      String.downcase(source_format) == "application/zip"
```
JarredOlson: Do we need this here? Odo should be emitting a hosted file with the source format of `application/geo+json`  if it isn't I think we should change Odo and not do this here.

apps/reaper/lib/reaper/init.ex
ManApart: Gonna keep the sleep.

apps/reaper/lib/reaper/decoder/geojson.ex
rajones-hntb: Removed this code

PR #349, opened 2019-11-25
CI for estuary

dunson062786:
…ockerfile to Estuary project.


apps/estuary/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM smartcitiesdata:build as builder
+RUN MIX_ENV=prod mix distillery.release --name estuary
```
jmorris-pillar: looks like distillery is not a depdenency in this app yet.

jmorris-pillar: sorry to give you peacemeal changes.

.travis.yml
```diff
@@ -48,4 +48,5 @@ jobs:
 branches:
   only:
   - master
+  - estuary_umbrella
```
ManApart: This doesn't seem consistent with how we've done the other apps. Are we using this just temporarily and then planning on removing him once we're no longer branching from a branch?

apps/estuary/lib/estuary/helper/kafka_helper.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Estuary.KafkaHelper do
+  def list_exisiting_topics do
+    Elsa.list(localhost: 9092)
```
ManApart: May be worth doing an IO.inspect on this guy to see the output when you call this (I'm assuming little test) function.

dunson062786: yes

kcarra:
@jmorris-pillar @ManApart Since we've decided to go with 1.9 should we be looking to replace Distillery with `mix release`?

apps/estuary/mix.lock
```diff
@@ -1,12 +0,0 @@
-%{
```
ManApart: Why was this guy deleted?

dunson062786: This is not required for each app. 
-Sukaant

ManApart: Ok, I missed that. Thanks.

.travis.yml
dunson062786: yes

PR #355, opened 2019-11-26
Forklift.Init monitors Pipeline.DynamicSupervisor

jdenen:
If `Pipeline.DynamicSupervisor` were to crash, Forklift would never re-init its dataset topic readers. This PR refactors `Forklift.Init` into a GenServer and monitors the DynamicSupervisor. On crash, `Forklift.Init` will re-init those readers.

### DIFF
--- a/apps/forklift/lib/forklift/init.ex
+++ b/apps/forklift/lib/forklift/init.ex
@@ -16,9 +16,16 @@ defmodule Forklift.Init do
 
   def init(_) do
     with :ok <- Forklift.DataWriter.bootstrap(),
-         :ok <- init_readers() do
-      {:ok, []}
+         :ok <- init_readers(),
+         pid <- Process.whereis(Pipeline.DynamicSupervisor) do
+      Process.monitor(pid)
+      {:ok, %{pipeline: pid}}
     end
+  end
+
+  def handle_info({:DOWN, _, _, down, _}, %{pipeline: pid}) when pid == down do
+    :ok = init_readers()
+    {:noreply, %{pipeline: Process.whereis(Pipeline.DynamicSupervisor)}}
   end
 
   defp init_readers do
@@ -51,5 +58,4 @@ defmodule Forklift.Init do
       [error | _] -> {:error, error}
     end
   end
-
 end
--- a/apps/forklift/lib/forklift/init.ex
+++ b/apps/forklift/lib/forklift/init.ex
@@ -2,31 +2,35 @@ defmodule Forklift.Init do
   @moduledoc """
   Task to initialize forklift and start ingesting each previously recorded dataset
   """
-  use Task, restart: :transient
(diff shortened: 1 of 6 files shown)


apps/forklift/lib/forklift/init.ex
```diff
@@ -2,31 +2,42 @@ defmodule Forklift.Init do
   @moduledoc """
   Task to initialize forklift and start ingesting each previously recorded dataset
   """
-  use Task, restart: :transient
+  use GenServer
 
   alias Forklift.MessageHandler
-  import Forklift
 
+  @name :forklift_init_server
```
jeffgrunewald: should this be a function call that extends the application name set by the user (if they set one explicitly)?

apps/forklift/lib/forklift/init.ex
```diff
@@ -35,4 +46,16 @@ defmodule Forklift.Init do
       topic_subscriber_config: Application.get_env(:forklift, :topic_subscriber_config)
     ]
   end
+
+  defp initialize(args) do
+    Process.sleep(250)
```
jeffgrunewald: What’s this sleep get us if we still get a whole bunch at once? Can we throw some jitter in here?

jdenen: Don't know. Y'all put that sleep in here awhile back. I just moved it.

apps/forklift/lib/forklift/init.ex
jdenen: Maybe? The ability to name it as a user is a testing thing. It should be `:forklift_init_server` for all production releases IMO, which is why it's not a configurable thing via `runtime.exs`.

jdenen: Maybe? The ability to name it as a user is a testing thing. It should be `:forklift_init_server` for all production releases IMO, which is why it's not a configurable thing via `runtime.exs`.

jeffgrunewald: Fair enough.

apps/forklift/lib/forklift/init.ex
jeffgrunewald: :slow_clap:

jeffgrunewald:
I’ll resist the urge to _seriously_ suggest `Forklift.Reanimator` and lean in the direction of `Forklift.FeedMonitor` or similar.

apps/forklift/lib/forklift/init.ex
jdenen: Thank you

jdenen:
It doesn't monitor feeds. It monitors a single dynamic supervisor.

jdenen:
I'm going with `Forklift.InitServer` since it's a process that sits and waits to initialize dataset readers now. 

🤷‍♂

PR #357, opened 2019-11-26
Andi -- View Dataset Details

NathanielScottStevens:
smartcitiesdata/issues/322


apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -0,0 +1,90 @@
...
+          <%= Form.label(f, :language, "Language", class: "label") %>
+          <%= Form.text_input(f, :language, value: @dataset.business.language, class: "input") %>
+        </div>
+        <div class="metadata-form__homepage">
+          <%= Form.label(f, :homepage, "Data Homepage URL", class: "label") %>
+          <%= Form.text_input(f, :homepage, value: @dataset.business.homepage, class: "input") %>
+        </div>
+      <% end %>
+      <%= Link.link("Cancel", to: "/", class: "btn btn--cancel metadata-form__cancel-btn") %>
+    </div
```
jdenen:
```suggestion
    </div>
```
I'm guessing you want that `>`?

apps/andi/assets/css/app.scss
```diff
@@ -23,4 +24,65 @@ body {
   margin: 1em;
 }
 
+@mixin area($block, $element) {
+  .#{$block}__#{$element} {
+    grid-area: $element
```
LtChae: What does this do? I'm not familiar with this scss syntax.

apps/andi/assets/css/app.scss
```diff
@@ -23,4 +24,65 @@ body {
...
+  border-radius: 2px; 
+}
+
+.textarea {
+  width: 100%;
+  max-width: 35rem;
+  height: 65%;
+}
+
+
```
mcsearchin: I'm wondering if all of this CSS should go into another file.  Maybe it's premature, but it already seems like `app.scss` is getting kinda big.

apps/andi/assets/css/variables.scss
```diff
@@ -18,3 +18,6 @@ $table-padding: .5em 1.5em;
 $table-height: 2em;
 
 $box-shadow: 0 0 6px rgba(20, 20, 20, 0.2);
+
+$label-color: #7F7676;
+$border-color: #E2E5EA;
```
mcsearchin: What is the intended difference between `border-color` and `color-border`?  😁

apps/andi/README.md
```diff
@@ -22,7 +22,7 @@ datasets = Enum.map(1..3, fn _ -> SmartCity.TestDataGenerator.create_dataset([])
 ```
 
 ```
-Brook.Test.with_event(:andi, fn -> Enum.each(datasets, fn dataset ->  Brook.ViewState.merge(:dataset, dataset.id, dataset)  end) end)
+Enum.each(datasets, &(Brook.Event.send(:andi, "dataset:update", :andi, &1)))
```
ManApart: As the author of the line you deleted, I appreciate this refactor.

apps/andi/lib/andi_web/controllers/edit_controller.ex
```diff
@@ -0,0 +1,17 @@
+defmodule AndiWeb.EditController do
+  use AndiWeb, :controller
+  alias Andi.DatasetCache
+
+  def show(conn, %{"id" => id}) do
+    case DatasetCache.get(id) do
+      %{"dataset" => dataset} ->
+        live_render(conn, AndiWeb.EditLiveView, session: %{dataset: dataset})
+
+      _ ->
```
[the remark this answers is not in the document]
NathanielScottStevens: 404s are logged by phoenix by default. It feels to me like we shouldn't be adding additional logging when users try to hit a non existing route. Though I can see the point about this occuring because of an error with our event handling or caching being worth logging. However, I wouldn't be able to differentiate between the two causes at this point in the code.

TLDR: I wouldn't log this but I could be convinced otherwise.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -0,0 +1,90 @@
+defmodule AndiWeb.EditLiveView do
+  use Phoenix.LiveView
+  alias Phoenix.HTML.Form
+  alias Phoenix.HTML.Link
+
+  def render(assigns) do
+    ~L"""
+    <div class="edit-page">
+      <%= Form.form_for :metadata, "#", [class: "metadata-form"], fn f -> %>
+        <div class="metadata-form__title">
```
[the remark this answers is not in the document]
NathanielScottStevens: The css is in edit.scss. It is used to set up the grid.

apps/andi/lib/andi_web/router.ex
```diff
@@ -22,8 +26,8 @@ defmodule AndiWeb.Router do
     pipe_through :browser
 
     get "/", Redirect, to: "/datasets"
-    live "/datasets/live", DatasetLiveView, session: [:path_params]
-    resources "/datasets", DatasetController, only: [:index]
+    live "/datasets", DatasetLiveView, session: [:path_params]
```
ManApart: Thank you for this.

apps/andi/assets/css/variables.scss
ManApart: If you don't understand this intuitively, McSearchin, I'm not sure we can explain it to you.

(Seriously though good catch. I'm also curious).

apps/andi/test/unit/andi_web/live/dataset_live_view/table_test.exs
```diff
@@ -154,7 +154,7 @@ defmodule AndiWeb.DatasetLiveViewTest.TableTest do
 
     {:ok, _view, html} = live(conn, @url_path)
 
-    assert get_rendered_table_cells(html) == [["", dataset.business.dataTitle, dataset.business.orgTitle]]
+    assert get_rendered_table_cells(html) == [["", dataset.business.dataTitle, dataset.business.orgTitle, "Edit"]]
```
mcsearchin: Maybe I'm missing it, but is there a test verifying the href of the "Edit" link?

apps/andi/assets/css/app.scss
NathanielScottStevens: `#{$var}` = string interpolation (I assume that's what's confusing about it)
It will generate this:
```
.my-block__my-element {
  grid-area: my-element;
}
```
If you look in edit.scss you can see where I called it a ton and make that value judgement about whether the complexity is worth the space saved (I'm expecting us to use css grid a lot in this app so it should have utility outside of this set of changes).

apps/andi/assets/css/app.scss
NathanielScottStevens: Yeah, I was stressing about this ealier lol. I came to the point where I couldn't decide whether to move it out into one "generic" file or a file for each "component" and then decided that it was premature. You could easily convince me otherwise though.

apps/andi/test/unit/andi_web/live/dataset_live_view/table_test.exs
NathanielScottStevens: No, it seemed like that fell under that category of things not worth testing in the UI. We added it because the tests failed otherwise. I think that's a smell and thought about addressing it in a better way but we decided the JSR wasn't there. Do you feel differently?

apps/andi/assets/css/variables.scss
NathanielScottStevens: I'm just going to rip these variables out

apps/andi/assets/css/app.scss
mcsearchin: No need to stress, man.  I'm good with leaving it in here for now as all of this styling is for generic components.  Just wondered if you'd thought about it.

apps/andi/test/unit/andi_web/live/dataset_live_view/table_test.exs
mcsearchin: This feels like something that I *think* would be easy to test and documents exactly what the expected path should be.  Seems like that's enough juice for very not much squeeze.  But that's just this juicer's opinion.

apps/andi/test/unit/andi_web/live/edit_live_view_test.exs
```diff
@@ -0,0 +1,68 @@
...
+      dataset = TDG.create_dataset(%{business: %{keywords: nil}})
+      DatasetCache.put(dataset)
+
+      assert {:ok, _view, html} = live(conn, @url_path <> dataset.id)
+      [subject] = Floki.find(html, "#metadata_keywords") |> Floki.attribute("value")
+
+      assert subject == ""
+    end
+  end
+end
```
mcsearchin: I thought I added this comment earlier, but looks like it got lost:

Would it be worth adding a Checkov-style test just to verify that all of the other 'pass-through' metadata fields are showing up on the form?  Again - doesn't seem like that much squeeze for a decent amount of juice.

apps/andi/assets/css/app.scss
ManApart: Man, how do I follow such a kind sincere comment like that. It does my heart good to see.

apps/andi/test/unit/andi_web/live/dataset_live_view/table_test.exs
ManApart: If I were working the card I would have agreed with Nathaniel... but now that Tony's called it out I can see it making sense... 🤷‍♂

NathanielScottStevens: Cool, we'll do that.

apps/andi/test/unit/andi_web/live/edit_live_view_test.exs
NathanielScottStevens: We'll do this as well

PR #361, opened 2019-11-26
Estuary create kafka topic

kcarra:


apps/estuary/config/test.exs
```diff
@@ -1 +1,5 @@
 use Mix.Config
+
+config :estuary,
+  elsa_endpoint: [localhost: 9092],
+  event_stream_topic: "event-stream"
```
jmorris-pillar: I dont think we need this anymore

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Estuary.StartTest do
+  use ExUnit.Case
+  use Divo
+
+  @elsa_endpoint Application.get_env(:estuary, :elsa_endpoint)
+  @event_stream_topic Application.get_env(:estuary, :event_stream_topic)
```
jmorris-pillar: are these two lines necessary?

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -0,0 +1,19 @@
...
+
+  setup do
+    case Application.ensure_all_started(:estuary) do
+      {:ok, _apps} -> :ok
```
jmorris-pillar: do we need this block anymore now that we're using divo?

apps/estuary/config/integration.exs
```diff
@@ -2,6 +2,8 @@ use Mix.Config
 
 config :estuary,
   divo: [
-    {DivoKafka, [create_topics: "topic1:1:1,topic2:1:1:compact", outside_host: "localhost"]}
+    {DivoKafka, [create_topics: "event-stream-integration:1:1", outside_host: "localhost"]}
```
jmorris-pillar: do we need to create a topic here anymore since our app does it?

apps/estuary/config/dev.exs
```diff
@@ -0,0 +1,5 @@
+use Mix.Config
+
+config :estuary,
+  elsa_endpoint: [localhost: 9092],
+  event_stream_topic: "event-stream"
```
jmorris-pillar: this probably belongs in either config.exs or integration.exs

apps/estuary/lib/estuary.ex
```diff
@@ -1,18 +1,2 @@
 defmodule Estuary do
```
dunson062786: Do we need this file?

apps/estuary/test/integration/application_start_test.exs
dunson062786: I think they are helpful for readability. We will probably use them in the future when we write test for consuming events off the event stream.

jmorris-pillar: I dont think we will since these should not need to be used in the test.  I say YAGNI unless the tests fail without these lines of code.

apps/estuary/lib/estuary.ex
jmorris-pillar: I do not believe so.

apps/estuary/test/integration/application_start_test.exs
kcarra: They are used in the first integration test to connect to Elsa

jmorris-pillar: D'oh.  I didnt notice that.  Disregard.

apps/estuary/config/integration.exs
jmorris-pillar: We will need this if we move the dev.exs to integration.exs, but not if we move it to config.exs

apps/estuary/config/test.exs
jmorris-pillar: We will need this if we move the dev.exs to integration.exs, but not if we move it to config.exs

apps/estuary/test/integration/application_start_test.exs
kcarra: I would say this is a nice to have for now if we don't want to run our integration tests using divo.

jmorris-pillar: My fear is it getting copied and pasted in other places.  I wont strong arm you into deleting it, but if it ends up in more than 1 file, I will probably put in a PR to delete it.

apps/estuary/config/test.exs
kcarra: Will the dev.exs config only get used if we set `MIX_ENV=dev`? Do we currently do that for any environments? I've only seen the `MIX_ENV` set to integration

jmorris-pillar: dev.exs is the default environment for iex -S mix.  If these properties are in config, they're used in all environments.  If they're in integration, they're only used in integration.  Typically we only put these in integration and actually i mis spoke you dont have unit tests that need these so they dont need to be in test.exs either way.

apps/estuary/config/dev.exs
jeffgrunewald: Keywords can be tricky since they get merged when you layer them from different config files (config.exs, prod.exs, runtime.exs for instance). It may be safer to keep these values out of the root config file unless you know they're not likely to change.

apps/estuary/config/test.exs
jeffgrunewald: Per my comment above, you probably want to keep these values out of the root config file, but that being said, if you focus your tests accordingly (and possibly write conditionals to _not_ start certain components like elsa outside of integration or prod) you can just omit these values from the config file for that environment.

apps/estuary/mix.exs
```diff
@@ -38,7 +38,10 @@ defmodule Estuary.MixProject do
   end
 
   defp aliases do
-    [verify: ["format --check-formatted", "credo"]]
+    [
+      verify: ["format --check-formatted", "credo"],
```
jeffgrunewald: this command is going to fail if you don't include credo in your deps.

apps/estuary/test/integration/application_start_test.exs
jeffgrunewald: not sure i understand the comment; you're going to run your tests with Divo by virtue of including `use Divo` in the header of the file.
unless you explicitly say `--no-start` when you run the mix task to prevent the VM from starting your app, they will always be started. this setup is not buying you anything in the integration environment.

apps/estuary/test/integration/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+version: '2'
```
jeffgrunewald: not sure what this is getting you.
you've already go `divo_kafka` in your dependencies so you can define the service in that single child spec-style line. it also doesn't look like you're actually using this file anywhere.
if you do have a reason to keep the compose file, you're going to want to add a healthcheck to the file to ensure that kafka is fully started and "clustered" appropriately or you'll get unstable behavior interacting with it in your tests.

apps/estuary/test/integration/application_start_test.exs
kcarra: It seems like by including `use divo` it starts `docker-compose` and does the health check. This was adding extra time when running the tests locally if kafka doesn't start up right away which seems to be very sporadic. What I was doing locally was commenting out the `use divo` manually starting the containers and just letting them run so I could run my integration tests multiple times without having to start and stop the containers and do the health check. To do that though you have to make sure the full supervisor tree starts which is what this is doing. I believe divo is doing something similar as well.

PR #362, opened 2019-11-27
Tag version and Mix version must match

jdenen:
Adds the `app.version` mix task to reaper and valkyrie. Forces mix version and tag version to match before a release's Docker image can be built.


scripts/travis-publish.sh
```diff
@@ -5,6 +5,12 @@ set -e
 if [ ! -z "$TRAVIS_TAG" ]; then
     app=$(echo "$TRAVIS_TAG" | cut -d@ -f1)
     vsn=$(echo "$TRAVIS_TAG" | cut -d@ -f2)
+    mix_vsn=$(mix cmd --app $app mix app.version | tail -1)
```
jmorris-pillar: what monster couldn't be bothered to just spell out version

PR #363, opened 2019-11-27
Patching the high-vulnerability node module used in andi

jeffgrunewald:
This should resolve the issue in the node `js-yaml` dependency in use by Andi's phoenix static assets.


jakeprem:
This should fix #363

PR #364, opened 2019-11-27
Updating architecture diagram

jeffgrunewald:


jeffgrunewald:
i lost track of where the new diagram came from once we handed it over to Big Jake to make prettier than any of us could do. he has it in an editable format somewhere if no one else does.

PR #367, opened 2019-12-02
Create table

sukaant:


apps/estuary/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    - metastore
+    - minio
+    image: smartcitiesdata/presto-testo:development
+    ports:
+    - "8080:8080"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -s http://localhost:8080/v1/info | grep -q '\"starting\":false'"]
+      interval: 10s
+      timeout: 30s
+      retries: 10
```
mcsearchin: Can this configuration be done with the canned libraries that @jeffgrunewald built for Divo?  I know there's one for kafka, not sure about presto.

jeffgrunewald: Presto is a tricky one b/c of the configuration that’s required and the fact that it more or less requires config files on the pods’ file system. We’ve abstracted that out with the “testo” images by baking them specifically for our own test environment requirements. 

In order to feel comfortable publishing these to Hex as a publicly available “pre-baked” package we’d need to create the image with a templating engine built in for writing out those files at runtime based on envars or requiring a config value pointing to a file in the project directory for mounting in.

ManApart: Can we not use the presto-testo image here?

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -4,8 +4,34 @@ defmodule Estuary.StartTest do
...
 
   test "Topic is created when Estuary starts" do
     assert Elsa.Topic.exists?(@elsa_endpoint, @event_stream_topic)
   end
+
+  test "should create event_stream table when estuary starts" do
+    expected_table_value = [[@event_stream_table_name]]
+
+    actual_table_value =
+      Prestige.execute("SHOW TABLES LIKE '#{@event_stream_table_name}'")
```
ManApart: That's a cool operation. I've not used `show tables` with a `like` before.

apps/estuary/test/integration/event_table_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Estuary.EventTableTest do
```
ManApart: Personally, I'd move your other create table test into this file. You could also put all 3 tests in the same file, which would speed up the test run time (though I like that you broke them out). I think I'd also prefer that you do the same  show tables check as it'd be easy to just return `[[true]]` from `EventTable.create_table()` to make the test pass. Doing the show tables really proves to me that the table is (still) there.

apps/estuary/lib/estuary/event_table.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Estuary.EventTable do
+  @moduledoc """
+  This modules does the CRUD operation for event_stream in the presto database
```
ManApart: This module currently does NOT do CRUD operation, and it may never do all of CRUD. I'd either do a module doc false or explain the current behavior of the module (just creates the table), otherwise this comment is misleading.

apps/estuary/docker-compose.yml
jeffgrunewald: That’s what they’re using but you still need to setup the port mapping, envs it does accept and start/stop it. Presto-testo just encapsulates the config files.

apps/estuary/test/integration/event_table_test.exs
sukaant: I agree with you to put all the create table test to EventTableTest file but the reason I've left 2 tests in the ApplicationStartTest file is that it will show those test belongs to startup (in future it may be easy to understand) and the 1 test in EventTableTest is to make sure if someone has broke the create table function which belongs to EventTable, then it will be useful in that scenario. Let me know your thoughts.

apps/estuary/lib/estuary/event_table.ex
sukaant: That make sense to me, I'll update it to show the actual behavior.

apps/estuary/test/integration/event_table_test.exs
sukaant: Regarding `Show Table`, as per my understanding it is already tested in the `ApplicationStartTest` and it may be duplicated and for `EventTable.create_table()` it should always return true because we are using `IF NOT EXISTS`, we may avoid the other call to the database by firing the other `SHOW TABLE` query

apps/estuary/docker-compose.yml
sukaant: Thanks @jeffgrunewald for clarification!

apps/estuary/test/integration/event_table_test.exs
ManApart: Generally I think more along the lines of 'what are we doing' than 'when are we doing things' in an app. If I care about the table creator I'd go to the table creator test as opposed to the 'startup test', so it makes more sense to me to have all tests regarding table creation in the table creator test as opposed to a startup test. Just my 2 cents.

EventTable.create_table() is returning the result of a prestige call. That call may change its return value which would break this test (even though it may still be doing its job correctly). Also, I don't care _what_ prestige returns, I care that the table (still) _exists_. So I think checking the outcome is more robust / explicit than checking the return value.

apps/estuary/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    depends_on:
+    - zookeeper
+    ports:
+    - "9092:9092"
+    environment:
+    - "KAFKA_ADVERTISED_LISTENERS=INSIDE://:9094,OUTSIDE://${HOST_IP:-localhost}:9092"
+    - "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT"
+    - "KAFKA_LISTENERS=INSIDE://:9094,OUTSIDE://:9092"
+    - "KAFKA_INTER_BROKER_LISTENER_NAME=INSIDE"
+    - "KAFKA_CREATE_TOPICS=test-topic:1:1"
```
kcarra: I don't think it's necessary to create this topic when kafka starts. We only care about the `event-stream` topic at this point

sukaant: Unfortunately it seems that the healthcheck for kafka does not succeed unless we create a topic here.  We might be able to get around this but I'm unsure if the JSR is there at this moment

apps/estuary/docker-compose.yml
sukaant: No, it is not possible.

apps/estuary/test/integration/event_table_test.exs
sukaant: Added one more assert for `show table`

PR #368, opened 2019-12-03
Pull Request for reading events from the event-stream

dunson062786:
[Pull Request for reading events from the event-stream](https://github.com/smartcitiesdata/smartcitiesdata/issues/342)


apps/estuary/config/prod.exs
```diff
@@ -0,0 +1,4 @@
+use Mix.Config
+
+# Do not print debug messages in production
```
ManApart: I appreciate the comments here, but they're probably unnecessary.

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -1,11 +1,68 @@
...
+    end
+
+    test "Topic is created when Estuary starts" do
+      assert Elsa.Topic.exists?(@elsa_endpoint, @event_stream_topic)
+    end
+
+    test "Estuary.Message reads message from eventstream when event stream receives an event" do
+      Elsa.produce(@elsa_endpoint, @event_stream_topic, {"key", "value1"})
+
+      Process.sleep(2000)
```
ManApart: Do we for sure need this sleep?

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -1,11 +1,68 @@
...
+  require Logger
 
   @elsa_endpoint Application.get_env(:estuary, :elsa_endpoint)
   @event_stream_topic Application.get_env(:estuary, :event_stream_topic)
 
-  test "Topic is created when Estuary starts" do
-    assert Elsa.Topic.exists?(@elsa_endpoint, @event_stream_topic)
+  describe "Estuary.MessageHandler receives messages sent to the event stream" do
+    setup do
+      allow(Estuary.MessageHandler.handle_messages(any()),
```
ManApart: With this being mocked (in an integration test), it feels a lot like we're just testing the Elsa does what it says it does. Maybe we could get rid of this mocking and the test on line 25, and keep the create topic test and the debug test?

dunson062786: I agree that it looks like we are testing that elsa does what it says what it does. I just don't know of a better way to test that we read the message without testing what we do with the message. I broke them up into two steps, because I think "reading the message" and "logging the message" are two different tests.

apps/estuary/mix.exs
```diff
@@ -34,6 +34,7 @@ defmodule Estuary.MixProject do
       {:divo, "~> 1.1", only: [:dev, :integration]},
       {:divo_kafka, "~> 0.1.5", only: [:dev, :integration]},
       {:distillery, "~> 2.1"},
+      {:placebo, "~> 1.2", only: [:dev, :test, :integration]},
```
sukaant: Shall we put all the dependencies in alphabetical order, it will be easy to find out in future

apps/estuary/config/prod.exs
dunson062786: Agreed

apps/estuary/test/integration/application_start_test.exs
dunson062786: Yeah I caught that this morning I will make this an eventually. Thanks!

apps/estuary/test/integration/application_start_test.exs
```diff
@@ -1,11 +1,68 @@
...
   use Divo
+  import ExUnit.CaptureLog
+  require Logger
 
   @elsa_endpoint Application.get_env(:estuary, :elsa_endpoint)
   @event_stream_topic Application.get_env(:estuary, :event_stream_topic)
 
-  test "Topic is created when Estuary starts" do
-    assert Elsa.Topic.exists?(@elsa_endpoint, @event_stream_topic)
+  describe "Estuary.MessageHandler receives messages sent to the event stream" do
```
sukaant: Shall we consider moving the tests to separate file if those are not related to startup of the application

apps/estuary/lib/estuary/application.ex
```diff
@@ -12,12 +12,31 @@ defmodule Estuary.Application do
...
+      {Elsa.Supervisor, elsa_options()}
+    ]
 
     opts = [strategy: :one_for_one, name: Estuary.Supervisor]
     Supervisor.start_link(children, opts)
   end
 
+  defp elsa_options do
+    [
+      endpoints: @elsa_endpoint,
```
jakeprem: This is going to give you problems when you go to deploy.

Module attributes (`@thing`) are evaluated at compile time, so this is going to hardcode your Elsa endpoint rather than getting it from `runtime.exs`

apps/estuary/mix.exs
dunson062786: sure

apps/estuary/lib/estuary/application.ex
dunson062786: Agred

apps/estuary/test/integration/application_start_test.exs
ManApart: For me the proof that you read the message (for now) is that you were able to log it.

apps/estuary/test/integration/application_start_test.exs
ManApart: Yeah, it'd be good to either have all the tests in the same file with a general name (for testing speed) or to have them broken out by feature/ area under test. (Slower but clearer).

apps/estuary/test/integration/application_test.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Estuary.ApplicationTest do
```
ManApart: I like this name a lot better.

apps/estuary/test/integration/message_handler_test.exs
```diff
@@ -1,9 +1,10 @@
-defmodule Estuary.StartTest do
+defmodule Estuary.MessageHandlerTest do
```
ManApart: Nice rename.

apps/estuary/test/integration/application_start_test.exs
dunson062786: @jmorris-pillar what do you think?

apps/estuary/test/integration/application_start_test.exs
dunson062786: Done

apps/estuary/test/integration/application_test.exs
dunson062786: thanks!

apps/estuary/test/integration/message_handler_test.exs
dunson062786: Thanks!

apps/estuary/test/integration/application_start_test.exs
dunson062786: done

PR #371, opened 2019-12-03
Adding Estuary to umbrella

sukaant:


apps/estuary/config/prod.exs
```diff
@@ -0,0 +1,4 @@
+use Mix.Config
+
+# Do not print debug messages in production
```
ManApart: Are these comments helpful? (this and the one in the test config)

apps/estuary/test/unit/estuary_test.exs
```diff
@@ -0,0 +1,9 @@
+defmodule EstuaryTest do
```
ManApart: I think we can safely delete this module/file at this point.

dunson062786: fixed

apps/estuary/config/prod.exs
dunson062786: Sorry, I thought I had removed this earlier. It's gone now.

PR #372, opened 2019-12-04
Create user-story.md

ManApart:
Create a template that fits the team norm.

### DIFF
--- a/.github/ISSUE_TEMPLATE/user-story.md
+++ b/.github/ISSUE_TEMPLATE/user-story.md
@@ -23,4 +23,4 @@ So that my life is better
 - Not necessarily an acceptance criteria, but a hint for the people that play the card
 
 #### Amigos: 
-A list of tagged users that should amigo this card
+A list of tagged users that should be involved in reviewing (or exiting) this work
--- a/.github/ISSUE_TEMPLATE/user-story.md
+++ b/.github/ISSUE_TEMPLATE/user-story.md
@@ -8,7 +8,7 @@ assignees: ''
 ---
 
 #### User Story
-As a issue creature
+As a issue creator
 I want to be able to create issues
 So that I can see my issue resolved
 
--- a/.github/ISSUE_TEMPLATE/user-story.md
+++ b/.github/ISSUE_TEMPLATE/user-story.md
@@ -8,15 +8,15 @@ assignees: ''
 ---
 
 #### User Story
-As a issue creator
+As a user
 
-I want to be able to create issues
+I want to see something changed
 
-So that I can see my issue resolved
+So that my life is better
 
 #### Acceptance Criteria: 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -0,0 +1,26 @@
...
+
+#### User Story
+As a issue creator
+
+I want to be able to create issues
+
+So that I can see my issue resolved
+
+#### Acceptance Criteria: 
+- [ ] The issue exists in the list of issues
```
jakeprem: I think the "meta" examples here and on the User story are pretty confusing

.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] Issue has all fields properly filled out
+
+#### Tech Notes: 
+- Notes that are technical in nature
```
jakeprem: I think this section and the Amigos section are much clearer and more helpful

.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -0,0 +1,26 @@
...
+- Not necessarily an acceptance criteria, but a hint for the people that play the card
+
+#### Amigos:
```
jakeprem: Some explanation of what an Amigo is might be helpful as part of this

ManApart: This is for our team to use, and I think we understand each of these sections. What would you propose?

.github/ISSUE_TEMPLATE/user-story.md
ManApart: These are just here to give us a formatting jump start and have been approved by Carrie. What would you propose as an alternative?

jakeprem: - [ ] This condition that must be met for the card to be complete
- [ ] E.g. Reaper can do a thing

Exact wording can be whatever but I think having explanations or examples makes it clearer what part of the template is part of the format and what part should be replaced with details specific to the card.

.github/ISSUE_TEMPLATE/user-story.md
jakeprem: + A list of tagged users that should be involved in reviewing (or exiting) this work

Our team will be creating most of the issues on this repo (for now), however this template will be an option presented on our main open source repo. Anyone adding an issue here will see this as an option so we shouldn't assume that they have an in depth knowledge of terms our team uses.

Additionally, making this clear will help provide documentation for new contributors and developers joining the team.

ManApart: I've changed amigos in the not to be 'users that should be involved in reviewing (or exiting) this work'

Issue #276, opened 2019-11-05
Reaper ingest XML data sources


JarredOlson:
As a data curator, I want to load XML data sources

Business Value
We have multiple systems providing data in XML format. When these are loaded into the system, they should be able to be searched, viewed and downloaded in a way similar to JSON or CSV data sources.

Acceptance Criteria
- [ ] When reaper encounters a dataset with a `sourceType` of XML it successfully pushes data into the data pipeline.

Technical Notes
This work will largely be done in a XML decoder module in reaper and encompass translating a schema definition with xpath elements into something that is useable by sweet XML.
[Spike for Reference](https://github.com/SmartColumbusOS/migration/issues/231)
Blocked By: https://github.com/smartcitiesdata/smartcitiesdata/issues/256

- test against one or many of the prioritized XML datasets

Amigos:
@JarredOlson

PR #376, opened 2019-12-04
Add ability to ingest XML to Reaper

jakeprem:
Still doing some refactoring work but the code should be largely stable now.

#276


apps/reaper/test/support/test_data/ChargePoint.xml
```diff
@@ -0,0 +1,126 @@
+<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/">
```
LtChae: Should we commit something this Columbus specific to the general platform repo?

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+  ###############
+  def handle_event(:start_document, _, state) do
+    ok(state)
+  end
+
+  def handle_event(:end_document, _, state) do
+    ok(state)
+  end
+
+  def handle_event(:start_element, {tag_name, attributes}, %State{} = state) do
```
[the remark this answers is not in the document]
jakeprem: Oops I missed that. They should all be `%State{}`. Mostly just for clarity and because it should make the handler fail fast in development if we mess up a return value.

apps/reaper/lib/reaper/xml/xml_stream.ex
```diff
@@ -0,0 +1,83 @@
...
+  @moduledoc """
+  Extracts a stream of records from an XML document with minimal memory overhead
+  """
+  use GenStage
+
+  alias XMLStream.SaxHandler
+
+  defmodule State do
+    @moduledoc false
+    @enforce_keys []
```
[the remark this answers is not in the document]
jakeprem: I forgot to go back and add the keys. I think we should be enforcing any keys that must be provided when creating a new struct

apps/reaper/test/support/test_data/ChargePoint.xml
jakeprem: I think as long as we're OK to publish this XML then this provides a valuable test case.
  1. It has more complex tags than most of our other tests
  2. It has a block(s) of XML before the block containing the data, so it helps to verify that we can descend and ascend to find the data.

apps/reaper/lib/reaper/xml/xml_stream.ex
```diff
@@ -0,0 +1,83 @@
...
+    def update(%__MODULE__{} = state, fields) do
+      struct!(state, fields)
+    end
+  end
+
+  ######################
+  ## Client Functions ##
+  ######################
+  @doc """
+  Creates a stream of records found at the tls as XML strings that can then be parsed using xpaths.
```
ManApart: tls?

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+      struct!(__MODULE__, fields)
+    end
+
+    def update(%__MODULE__{}=state, fields) do
+      struct!(state, fields)
+    end
+  end
+
+  @behaviour Saxy.Handler
+  @file_chunk_size 40_000
```
ManApart: Should this be configurable?

jakeprem: Not anymore than it is as a module attribute. If we ever discover a scenario where it should be we can make it configurable then.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+          |> ok()
+      end
+    else
+      ok(state)
+    end
+  end
+
+  #######################
+  ## Private Functions ##
+  #######################
```
jdenen: 🚫

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+      state
+      |> State.update(stack: [tag | state.stack], accumulate: true)
+      |> ok()
+    else
+      ok(state)
+    end
+  end
+
+  def handle_event(:characters, chars, %State{stack: stack, accumulate: accumulate} = state) do
+    if accumulate do
```
jdenen: Can you pattern match the falsey case out in a function head and get rid of this `if/else`?

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  def handle_event(
+        :end_element,
+        tag_name,
+        %State{stack: stack, accumulate: accumulate, tag_stack: [_stack_tag_name | tag_stack]} = state
+      ) do
+    state = %{state | tag_stack: tag_stack}
+
+    if accumulate do
```
jdenen: Again, I think you can pattern out this `if/else`.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,133 @@
...
+      struct!(state, fields)
+    end
+  end
+
+  @behaviour Saxy.Handler
+  @file_chunk_size 40_000
+
+  ######################
+  ## Client Functions ##
+  ######################
```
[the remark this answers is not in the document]
jakeprem: It shows a clear separation between arbitrary functions that have I've added and things that are a standard part of the behavior.

I think they're helpful and make it easier to quickly parse and organize the file.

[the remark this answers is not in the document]
jdenen: So I get wanting to denote behaviour requirements from custom functions, but that's why you tag behaviour implementations with `@impl Saxy.Handler`.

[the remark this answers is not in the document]
jdenen: So I get wanting to denote behaviour requirements from custom functions, but that's why you tag behaviour implementations with `@impl Saxy.Handler`.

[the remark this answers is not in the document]
jdenen: So I get wanting to distinguish behaviour implementations from custom functions, but that's why you tag each behaviour implementation with `@impl Saxy.Handler`.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jakeprem: Good catch. That makes sense for this one.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jakeprem: This one I'm not sure about. In this one the function head is popping the `_stack_tag_name` off the list and then updating the state before the `if` block.

We could pattern match the accumulate but then we'd have to repeat that logic in two functions/heads, or create a function to share that functionality. Both options would work but they seem more indirect/complicated than just using an `if` in this case.

What are your thoughts?

apps/reaper/lib/reaper/decoder/xml.ex
```diff
@@ -0,0 +1,36 @@
...
+  defmodule XmlError do
+    defexception [:message]
+  end
+
+  @behaviour Reaper.Decoder
+
+  @impl Reaper.Decoder
+  def decode(
+        {:file, filename},
+        %SmartCity.Dataset{technical: %{schema: schema, topLevelSelector: top_level_selector}} = dataset
```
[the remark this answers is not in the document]
jakeprem: Agreed, I just didn't come up with a better name at the time. I switched all uses of `top_level_selector` to `selector` since I think only the dataset really needs to care that it's the "topLevelSelector"

apps/reaper/lib/reaper/xml/xml_stream.ex
```diff
@@ -0,0 +1,104 @@
...
+  def handle_call({:emit, record}, _from, %State{demand: demand} = state) when demand > 1 do
+    {:reply, :ok, [record], Map.update!(state, :demand, &(&1 - 1))}
+  end
+
+  def handle_call({:emit, record}, from, state) do
+    {:noreply, [record], State.update(state, demand: 0, blocked: from)}
+  end
+
+  def handle_demand(demand, state) do
+    if state.blocked != nil do
```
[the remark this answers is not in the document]
jakeprem: Yeah I think that's fine.

This also made me realize that I wasn't resetting the `blocked` field after unblocking it.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jdenen: I see what you're saying. This function is really big though, so it should probably be broken down anyway. How about something like:

```elixir
def handle_event(:end_element, tag_name, %State{tag_stack: [_ | tag_stack]} = state), do: private_function(tag_name, %{state | tag_stack: tag_stack})

defp private_function(tag, %State{accumulate: true} = state) do
  # if accumulate true logic
end

defp private_function(_tag, state), do: ok(state)
```

Obviously come up with a better private function name 😛

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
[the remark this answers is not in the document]
jakeprem: I tagged everything with `@impl true`. I think it does the same thing as including the behaviour module as well.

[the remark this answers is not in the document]
ManApart: I agree with Johnson. I really dislike seeing these comment blocks and it's not our team norm to add them.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jakeprem: I did a big refactor of the handler that should cover this

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,159 @@
...
+      [tag_name | tag_stack] == tag_path
+    end
+
+    def ascend(%__MODULE__{}=state) do
+      State.pop(state, :tag_stack)
+    end
+
+    def descend(%__MODULE__{tag_stack: tag_stack}=state, tag_name) do
+      State.update(state, tag_stack: [tag_name | tag_stack])
+    end
```
jdenen: Am I crazy or should these names be rearranged? I would consider ascending a stack to be putting thing onto it and descending to be popping them off.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
```diff
@@ -0,0 +1,159 @@
...
+
+  @impl true
+  def handle_event(:characters, chars, %State{stack: stack, accumulate: true} = state) do
+      [{tag_name, attributes, content} | stack] = stack
+
+      current = {tag_name, attributes, [chars | content]}
+
+      state
+      |> State.update(stack: [current | stack])
+      |> ok()
```
jdenen: Didn't look to see if the formatter complained, but the indentation is off here.

jakeprem: I didn't run the formatter yet, to more clearly show code changes vs formatting changes

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jakeprem: I was naming them like a depth-first search. So as you enter tags in the document, you're descending into the tree, then ascending again as you encounter the end tags.

apps/reaper/lib/reaper/xml/xml_stream/sax_handler.ex
jakeprem: When I ran it the formatter didn't do anything. Do you prefer the follow up lines to be indented here?

jdenen: I'm just referring to the entire function block being indented 4 spaces instead of the usual 2.

jdenen: I would have expected the formatter to catch that.

PR #378, opened 2019-12-05
Release config for CD

sukaant:


apps/estuary/config/integration.exs
```diff
@@ -9,6 +9,7 @@ config :prestige,
   ]
 
 config :estuary,
+  elsa_endpoint: [localhost: 9092],
   divo: "docker-compose.yml",
```
dunson062786: Is this still correct? Is this what it is on aws?

sukaant: This is the integration config

apps/estuary/lib/estuary/application.ex
```diff
@@ -41,9 +37,18 @@ defmodule Estuary.Application do
   end
 
   defp validate_topic_exists do
-    case Elsa.Topic.exists?(@elsa_endpoint, @event_stream_topic) do
-      true -> :ok
-      false -> Elsa.Topic.create(@elsa_endpoint, @event_stream_topic)
+    case Elsa.Topic.exists?(
```
ManApart: Personally I think it'd be cleaner to extract these guys into local variables (just for readability).

NathanielScottStevens: I think the way we've solved that before is by throwing them in functions.

dunson062786:
Is there anything else that this pull request does than move elsa endpoint from the main config to the integration config.

apps/estuary/lib/estuary/application.ex
```diff
@@ -2,14 +2,10 @@ defmodule Estuary.Application do
   # See https://hexdocs.pm/elixir/Application.html
   # for more information on OTP Applications
   @moduledoc false
-
   use Application
 
   alias Estuary.EventTable
 
-  @elsa_endpoint Application.get_env(:estuary, :elsa_endpoint)
```
jmorris-pillar: you found the footgun :-)

apps/estuary/lib/estuary/application.ex
jmorris-pillar: Agree^ . This is being done in more than 1 place so they should be in functions

apps/estuary/lib/estuary/event_table.ex
```diff
@@ -3,11 +3,9 @@ defmodule Estuary.EventTable do
   This module will create event_stream table and insert data in the table
   """
 
-  @event_stream_table_name Application.get_env(:estuary, :event_stream_table_name)
-
   def create_table do
     Prestige.execute(
-      "CREATE TABLE IF NOT EXISTS #{@event_stream_table_name} (author varchar, create_ts bigint, data varchar, type varchar)"
+      "CREATE TABLE IF NOT EXISTS #{Application.get_env(:estuary, :event_stream_table_name)} (author varchar, create_ts bigint, data varchar, type varchar)"
```
jmorris-pillar: toss this in a local function imo

PR #381, opened 2019-12-06
Events inserted into events table

dunson062786:
[Events can be inserted into events table](https://github.com/smartcitiesdata/smartcitiesdata/issues/344)

co-authored-by: Jessie Morris <jmorris2@pillartechnology.com>


apps/estuary/lib/estuary/application.ex
```diff
@@ -1,6 +1,4 @@
 defmodule Estuary.Application do
-  # See https://hexdocs.pm/elixir/Application.html
-  # for more information on OTP Applications
```
ManApart: Glad to see these comments go.

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -1,8 +1,33 @@
 defmodule Estuary.MessageHandler do
+  @moduledoc """
+  This module will read an event from the event stream and persist it.
```
ManApart:
```suggestion
  MessageHandler reads an event from the event stream and persists it.
```

ManApart: We should document what something does today as opposed to what it may/will do in the future.

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -1,8 +1,33 @@
...
+        )
+      else
+        {:error, %Jason.DecodeError{}} ->
+          Yeet.process_dead_letter("", message, "estuary",
+            reason: "event's JSON could not be decoded"
+          )
+
+        bad_keys ->
+          Yeet.process_dead_letter("", message, "estuary",
+            reason: "event #{inspect(bad_keys)} was decoded but did not have the right keys."
```
ManApart: Is it possible we're yeeting this message because we couldn't decode the json?

apps/estuary/config/integration.exs
```diff
@@ -1,5 +1,13 @@
 use Mix.Config
 
+host =
```
jmorris-pillar: do we need to break out this host here?  What is the usecase for taking the host from the environment in the integration config?

apps/estuary/lib/estuary/event_table.ex
```diff
@@ -9,6 +9,13 @@ defmodule Estuary.EventTable do
     |> Prestige.prefetch()
   end
 
+  def insert_event(author, create_ts, data, type) do
+    Prestige.execute("INSERT INTO #{table_name()}
+      (author, create_ts, data, type)
+      values ('#{author}', #{create_ts}, '#{data}', '#{type}')")
+    |> Prestige.prefetch()
```
jmorris-pillar: Do we want to handle failures to write to presto, such as presto down by dlq'ing messages that invoke presto errors?  I feel like other apps dont do this?  Wondering what the rest of the team thinks.

apps/estuary/runtime.exs
```diff
@@ -40,3 +40,7 @@ config :estuary,
 
 config :logger,
   level: :warn
+
+config :yeet,
+  endpoint: endpoints,
+  topic: System.get_env("DLQ_TOPIC")
```
jmorris-pillar: We will need to add this to the chart and deploy repo as a new environment variable that gets passed in.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+      assert events == [[1]]
+    end)
+
+    Prestige.execute("DELETE from #{@event_stream_table_name}") |> Prestige.prefetch()
+  end
+
+  test "estuary persists batch of events to the event stream" do
+    produce_event(@event_stream_topic, [
+      ~s({"__brook_struct__":"Elixir.Brook.Event","__struct__":"Elixir.SmartCity.Dataset","author":"forklift","create_ts":1,"data":"some data1","forwarded":false,"type":"some type1"}),
+      ~s({"__brook_struct__":"Elixir.Brook.Event","__struct__":"Elixir.SmartCity.Dataset","author":"valkyrie","create_ts":2,"data":"some data2","forwarded":false,"type":"some type2"})
```
jmorris-pillar: I wonder if this sigil should be in a parameterized function so the test is more readable.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+      events =
+        Prestige.execute(
+          "SELECT COUNT(*) from #{@event_stream_table_name} WHERE author = 'valkyrie' and create_ts = 2 and data = 'some data2' and type = 'some type2'"
+        )
+        |> Prestige.prefetch()
+
+      assert events == [[1]]
+    end)
+
+    Prestige.execute("DELETE from #{@event_stream_table_name}") |> Prestige.prefetch()
```
jmorris-pillar: This is in multiple tests, should it be in a setup block?

apps/estuary/lib/estuary/event_table.ex
sukaant: Yes, that will help in logging the insert failure early.

apps/estuary/docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
     - "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT"
     - "KAFKA_LISTENERS=INSIDE://:9094,OUTSIDE://:9092"
     - "KAFKA_INTER_BROKER_LISTENER_NAME=INSIDE"
-    - "KAFKA_CREATE_TOPICS=clusterready:1:1"
+    - "KAFKA_CREATE_TOPICS=clusterready:1:1,dead-letters:1:1"
```
kcarra: Now that we create the `dead-letters` topic on start of kafka can we remove `clusterready` topic since it's not used anywhere

apps/estuary/lib/estuary/event_table.ex
sukaant: Is there any mandatory field we care about?

apps/estuary/test/integration/event_table_test.exs
```diff
@@ -20,4 +20,24 @@ defmodule Estuary.EventTableTest do
     assert expected_value == actual_value
     assert expected_table_value == actual_table_value
   end
+
+  test "insert_event inserts an event into the #{@event_stream_table_name} table" do
+    aloha_events =
```
jmorris-pillar: lets remove aloha here and anywhere else

apps/estuary/lib/estuary/event_table.ex
```diff
@@ -9,6 +9,13 @@ defmodule Estuary.EventTable do
     |> Prestige.prefetch()
   end
 
+  def insert_event(author, create_ts, data, type) do
```
sukaant: Is it possible to minimize the no of paramaters, in the function?

apps/estuary/config/integration.exs
dunson062786: You're right we probably don't need this. I was copy/pasting.

apps/estuary/lib/estuary/event_table.ex
kcarra: Agree, might be worth patterning matching out an `:error` case

apps/estuary/lib/estuary/message_handler.ex
dunson062786: Agreed good catch

apps/estuary/lib/estuary/message_handler.ex
dunson062786: No it would match the above condition if that was the case.

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -1,8 +1,33 @@
 defmodule Estuary.MessageHandler do
+  @moduledoc """
+  This module will read an event from the event stream and persist it.
+  """
   use Elsa.Consumer.MessageHandler
   require Logger
 
   def handle_messages(messages) do
+    Enum.each(messages, fn message ->
```
kcarra: Can we possibly split some of this logic out into smaller functions to make it more readable? Maybe pipe the result of `Jason.decode` into a function that handles some of the edge/degenerate cases instead of having to do `else`

apps/estuary/test/integration/estuary_test.exs
kcarra: @jmorris-pillar i believe there's a `on_exit` callback that could accomplish this teardown

apps/estuary/test/integration/event_table_test.exs
```diff
@@ -20,4 +20,24 @@ defmodule Estuary.EventTableTest do
...
+    Estuary.EventTable.insert_event("Steve Aloha", 5, "some data", "some type")
+
+    aloha_events =
+      Prestige.execute(
+        "SELECT COUNT(*) from #{@event_stream_table_name} WHERE author = 'Steve Aloha' and create_ts = 5 and data = 'some data' and type = 'some type'"
+      )
+      |> Prestige.prefetch()
+
+    assert aloha_events == [[1]]
+    Prestige.execute("DELETE from #{@event_stream_table_name}") |> Prestige.prefetch()
```
sukaant: Do we need prefetch here?

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  test "estuary sends event to the dlq if it is properly formatted, but doesn't have the right keys" do
+    produce_event(@event_stream_topic, ~s({"foo": "bar"}))
+
+    eventually(fn ->
+      {:ok, _, events} = Elsa.fetch(@elsa_endpoint, "dead-letters")
+
+      if length(events) > 0 do
+        Enum.any?(events, fn event -> event.value == "{\"foo\": \"bar\"}" end)
```
kcarra: What is the truthy assertion for this test? Wouldn't you want to do something like `assert Enum.any?(events, fn event -> event.value == "{\"foo\": \"bar\"}" end)` it seems like it just asserts false if the length of the events is 0

apps/estuary/test/integration/event_table_test.exs
```diff
@@ -20,4 +20,24 @@ defmodule Estuary.EventTableTest do
...
+    aloha_events =
+      Prestige.execute(
+        "SELECT COUNT(*) from #{@event_stream_table_name} WHERE author = 'Steve Aloha' and create_ts = 5 and data = 'some data' and type = 'some type'"
+      )
+      |> Prestige.prefetch()
+
+    assert aloha_events == [[0]]
+    Estuary.EventTable.insert_event("Steve Aloha", 5, "some data", "some type")
+
+    aloha_events =
```
sukaant: Can we avoid reassigning to the same variable in the function, we may put some name which serves the purpose

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+      else
+        assert false
+      end
+    end)
+  end
+
+  defp produce_event(topic, payload) do
```
kcarra: Is this functions only purpose to produce Elsa events? If so can we rename it to `produce_elsa_event` if the purpose of this function is to be generic so we can eventually add other types of events then could we do something like 
``` 
defp produce_event(:elsa, topic, payload) do
defp produce_event(:brook, topic, payload) do
etc
```

apps/estuary/lib/estuary/event_table.ex
dunson062786: agreed. I'll pass it an event.

apps/estuary/test/integration/event_table_test.exs
kcarra: I know this was a separate card but might be worth moving this to `on_exit` call back since the first test in this file should have done the same thing. Looks like it was just missed

apps/estuary/runtime.exs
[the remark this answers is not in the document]
dunson062786: done

dunson062786: agreed

apps/estuary/lib/estuary/event_table.ex
ManApart: IF presto is down we don't want to DLQ. We want the app to crash / retry / reveal that it is missing the infrastructure it needs. If a message is bad/ we have an error with a single message we want to dlq that message.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+
+    actual_table_value =
+      "SHOW TABLES LIKE '#{@event_stream_table_name}'"
+      |> Prestige.execute()
+      |> Prestige.prefetch()
+
+    assert expected_table_value == actual_table_value
+  end
+
+  test "should check all the columns exists in the table when estuary starts" do
```
ManApart: I'd combine this with  the test above. The test above is checking that the table exists and may as well also check the table has the proper columns.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+      |> Prestige.execute()
+      |> Prestige.prefetch()
+
+    assert expected_column_value == actual_column_value
+  end
+
+  test "estuary persists event to the event_stream table" do
+    produce_event(
+      @event_stream_topic,
+      ~s({"__brook_struct__":"Elixir.Brook.Event","__struct__":"Elixir.SmartCity.Dataset","author":"reaper","create_ts":5,"data":"some data","forwarded":false,"type":"some type"})
```
ManApart: Nice.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,119 @@
...
+    ])
+
+    eventually(fn ->
+      events =
+        Prestige.execute(
+          "SELECT COUNT(*) from #{@event_stream_table_name} WHERE author = 'forklift' and create_ts = 1 and data = 'some data1' and type = 'some type1'"
+        )
+        |> Prestige.prefetch()
+
+      assert events == [[1]]
```
ManApart: I think `assert events == [[1]]` works well, but I'd like it even more if we did a select and saw we were getting an actual record back. That would prove that we don't always get [[1]] back for any query, it be different between tests, and it would explain what we expect to get back from prestige.

apps/estuary/test/integration/estuary_test.exs
ManApart: Yeah, do you even need the length check, or can  you just assert on the any?

apps/estuary/test/integration/estuary_test.exs
ManApart: How helpful is it to have this function if it's just converting a 2 param function into a 3 param function? I'd argue it's not worth having.

apps/estuary/test/integration/event_table_test.exs
ManApart: Good call Sukaant only a _monster_ would name shadow their variables.



@jmorris-pillar @jmorris-pillar @jmorris-pillar @jmorris-pillar @jmorris-pillar @jmorris-pillar

jdenen:
Why aren't you using `Pipeline.Writer.TableWriter` to interface with the Presto table? Then you could use `mox` style integration testing too.

apps/estuary/test/integration/estuary_test.exs
jmorris-pillar: Yea id yagni that.  I only pushed for this refactor because i wanted to semantically say "hey produce an event" rather than making readers of the test have to understand that elsa'ing to a specific place produced events

apps/estuary/config/integration.exs
```diff
@@ -1,5 +1,7 @@
 use Mix.Config
 
+endpoints = [localhost: 9092]
+
 config :prestige,
   base_url: "http://127.0.0.1:8080",
   headers: [
```
jmorris-pillar: Can we also change this schema to not be hive.default

apps/estuary/config/integration.exs
```diff
@@ -1,5 +1,13 @@
 use Mix.Config
 
+host =
+  case System.get_env("HOST_IP") do
+    nil -> "127.0.0.1"
```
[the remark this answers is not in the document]
dunson062786: seems like Jessie thinks we aren't going to need to break up the host here.

apps/estuary/config/integration.exs
dunson062786: What would you like me to do instead?

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -52,10 +50,8 @@ defmodule Estuary.EstuaryTest do
 
   test "should persist batch of events to the event stream" do
     produce_event(@event_stream_topic, [
-      ~s({"__brook_struct__":"Elixir.Brook.Event",
-      "__struct__":"Elixir.SmartCity.Dataset","author":"forklift","create_ts":1,"data":"some data for forklift","forwarded":false,"type":"some type for forklift"}),
-      ~s({"__brook_struct__":"Elixir.Brook.Event",
-      "__struct__":"Elixir.SmartCity.Dataset","author":"valkyrie","create_ts":2,"data":"some data for valkyrie","forwarded":false,"type":"some type for valkyrie"})
+      event_struct("forklift", 1, "some data for forklift", "some type for forklift"),
+      event_struct("valkyrie", 2, "some data for valkyrie", "some type for valkyrie")
```
ManApart: Much cleaner. :)

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  test "should send event to the dlq if it is not a properly formatted event" do
+    produce_event(@event_stream_topic, {"key", "value"})
+
+    eventually(fn ->
+      {:ok, _, events} = Elsa.fetch(@elsa_endpoint, "dead-letters")
+
+      if length(events) > 0 do
+        Enum.any?(events, fn event -> event.value == "{\"key\": \"value\"}" end)
```
kcarra: Same comment for the next test as well. I don't see any truth assertion for this test.

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,134 @@
...
+      "__struct__":"Elixir.SmartCity.Dataset",
+      "author":"#{event_data["author"]}",
+      "create_ts":"#{event_data["create_ts"]}",
+      "data":"#{event_data["data"]}",
+      "forwarded":false,
+      "type":"#{event_data["type"]}"
+      })
+  end
+
+  defp event_value(expected_value) do
```
JarredOlson: I had 0 idea what this function was doing based on its name.  Something along the lines of `event_exists_in_dlq?` might be clearer

apps/estuary/test/integration/estuary_test.exs
```diff
@@ -0,0 +1,134 @@
...
+
+    eventually(fn ->
+      actual_value =
+        "'forklift', 'valkyrie'"
+        |> EventTableHelper.select_table_data()
+
+      assert expected_value == actual_value
+    end)
+
+    EventTableHelper.delete_table_data()
```
JarredOlson: This probably belongs in an [ExUnit `on_exit` Callback](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html) since this won't run if this test fails which could cause invalid cascading failures.

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -1,9 +1,33 @@
...
+      {:ok, %{"author" => _, "create_ts" => _, "data" => _, "type" => _} = event} ->
+        EventTable.insert_event_to_table(event)
+
+      {:ok, term_without_keys} ->
+        Yeet.process_dead_letter("", message, "estuary",
+          reason:
+            "event #{inspect(term_without_keys)} was decoded but did not have the right keys"
+        )
+
+      {:error, %Jason.DecodeError{}} ->
```
JarredOlson: I'm not sure that we really need to differentiate between the above error and this one.

apps/estuary/test/integration/estuary_test.exs
dunson062786: agreed. I should probably of put wip on this pull request. We started refactoring and that was the quick name we made at first. It is now say is_in_dlq.

apps/estuary/test/integration/estuary_test.exs
dunson062786: Agreed

apps/estuary/lib/estuary/message_handler.ex
dunson062786: The option above it was an ok tuple and it was handling the case where json can be decoded, but it didn't have the correct keys so it couldn't be inserted into the DB.

dunson062786: It is still in the code, but I've broken it up into separate functions with different function signatures.

apps/estuary/test/integration/event_table_test.exs
```diff
@@ -20,4 +21,44 @@ defmodule Estuary.EventTableTest do
...
+      "data" => "some data",
+      "type" => "some type"
+    }
+    |> EventTable.insert_event_to_table()
+
+    actual_value_after_insert =
+      "'Steve'"
+      |> EventTableHelper.select_table_data()
+
+    assert expected_value_before_insert == actual_value_before_insert
```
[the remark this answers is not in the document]
dunson062786: Agreed

apps/estuary/lib/estuary/message_handler.ex
ManApart: My guess is that Jarred understands _how_ this code works and is stating he's not sure it's beneficial to handle those error cases individually instead of doing the same thing in either case. (You lose the granular error message, but how much does it buy us?)

```
case Jason.decode(message.value) do
      {:ok, %{"author" => _, "create_ts" => _, "data" => _, "type" => _} = event} ->
        EventTable.insert_event_to_table(event)

      _anything_else -> #yeet record
```

PR #395, opened 2019-12-13
update discovery streams to have app.version task and make publish sc…

bbrewer-papa-x:
…ript fail correctly when that does not come in right

co-authored-by: Kenny Drobnack <kdrobnack@pillartechnology.com>


bbrewer-papa-x:
Apparently, there was a wiki for this

PR #398, opened 2019-12-16
Edit metadata

ManApart:


apps/andi/assets/css/edit.scss
```diff
@@ -20,27 +20,31 @@
...
-@include area(metadata-form, keywords)
-
-@include area(metadata-form, last-updated)
-@include area(metadata-form, spatial)
-@include area(metadata-form, temporal)
-@include area(metadata-form, organization)
-@include area(metadata-form, level-of-access)
-@include area(metadata-form, language)
-@include area(metadata-form, homepage)
+.div__success-message {
```
NathanielScottStevens: I wouldn't call this block `div`

apps/andi/assets/css/edit.scss
```diff
@@ -20,27 +20,31 @@
...
+  text-align: right;
+  padding-right: 2em;
+  color: green;
```
NathanielScottStevens: Would it make sense to create a color variable and tie this to the checkmark color?

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,184 @@ defmodule AndiWeb.EditLiveView do
...
+              <%= error_tag(fp, :temporal) %>
+            </div>
+            <div class="metadata-form__organization">
+              <%= Form.label(fp, :orgTitle, "Organization", class: "label label--required") %>
+              <%= Form.text_input(fp, :orgTitle, [class: "input input--text", disabled: true]) %>
+              <%= error_tag(fp, :orgTitle) %>
+            </div>
+            <div class="metadata-form__language">
+              <%= Form.label(fp, :language, "Language", class: "label") %>
+              <%= Form.select(fp, :language, [[key: "English", value: "english"], [key: "Spanish", value: "spanish"]], value: get_language(Form.input_value(fp, :language)), class: "select") %>
```
NathanielScottStevens: I would pull the select options out into a module attribute

apps/andi/assets/css/edit.scss
ManApart: Lol, I thought YOU made that change. We'll call it something else.

apps/andi/assets/css/edit.scss
ManApart: I was going to wait until it was used in more than one place, but we can do that.

apps/andi/lib/andi_web/live/edit_live_view.ex
ManApart: JMS

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,184 @@ defmodule AndiWeb.EditLiveView do
...
+
     """
   end
 
   def mount(%{dataset: dataset}, socket) do
-    {:ok, assign(socket, dataset: dataset)}
+    new_business = dataset.business |> Map.from_struct()
+    new_technical = dataset.technical |> Map.from_struct()
+
+    change =
```
NathanielScottStevens: I would put this pipe into `Andi.DatasetSchema`. Maybe into the changeset function or perhaps a different changeset function (having multiple special changeset functions seems to be idiomatic)

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,184 @@ defmodule AndiWeb.EditLiveView do
...
+      |> Map.put(:business, new_business)
+      |> Map.put(:technical, new_technical)
+      |> Andi.DatasetSchema.changeset()
+
+    {:ok, assign(socket, changeset: change, is_saved: false)}
   end
 
-  defp get_private(%{technical: %{private: true}}), do: "Private"
-  defp get_private(_), do: "Public"
+  def handle_event(
```
NathanielScottStevens: I think we've talked about this some before, but reading this function it _seems_ like you should get everything that happens here for free. At the very least I would think you could do a `Changeset.apply` and avoid the `put_in` pipe.

apps/andi/assets/css/edit.scss
ManApart: It was also used as a checkmark, as you had guessed, so we're updating this as well.

apps/andi/lib/andi_web/live/edit_live_view.ex
ManApart: The `put_in` lines are parsing keywords and preventing changes to 'read-only' fields. I'll add a comment.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,171 @@ defmodule AndiWeb.EditLiveView do
...
+        %{"dataset_schema" => dataset_schema} = event,
+        socket
+      ) do
+    {:noreply, assign(socket, changeset: apply_changes(dataset_schema), is_saved: false)}
+  end
+
+  def handle_event("save", %{"dataset_schema" => dataset_schema} = event, socket) do
+    change = apply_changes(dataset_schema)
+
+    case change.valid? do
```
ManApart: This certainly reads better.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,171 @@ defmodule AndiWeb.EditLiveView do
...
+
+        %{
+          original_dataset
+          | business: Map.merge(Map.from_struct(original_dataset.business), Map.from_struct(schema.business)),
+            technical: Map.merge(Map.from_struct(original_dataset.technical), Map.from_struct(schema.technical))
+        }
+        |> SmartCity.Dataset.new()
+        |> send_dataset_update
+
+        {:noreply, assign(socket, changeset: change, is_saved: true, update_stepper_state: "meta-data-save")}
```
ManApart: What's the update stepper state?

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,171 @@ defmodule AndiWeb.EditLiveView do
...
+
+      false ->
+        {:noreply, assign(socket, changeset: change, is_saved: false, display_errors: true)}
```
ManApart: Do we still need the `display_errors` workaround?

ManApart:
It's looking a lot cleaner with the latest changes...

apps/andi/lib/andi_web/live/edit_live_view.ex
kcarra: The arbitrary property that we have to add to get live view to re-render the view if there is no diff... At some point, I'm assuming we well perform some sort of UI action on save

apps/andi/lib/andi_web/live/edit_live_view.ex
kcarra: I spent a decent bit of time looking into this and wasn't able to find the proper way to get the view to re-render. It's almost like live view is expecting some redirect to happen when a form is submitted and don't re render the view

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -153,7 +154,9 @@ defmodule AndiWeb.EditLiveView do
     Brook.Event.send(instance_name(), dataset_update(), :andi, dataset)
   end
 
-  defp send_dataset_update({:error, e}), do: {:error, e}
+  defp send_dataset_update({:error, e}) do
+    Logger.warn("Unable to update dataset: #{IO.inspect({:error, e})}")
```
ManApart: IO.inspect?

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -2,89 +2,178 @@ defmodule AndiWeb.EditLiveView do
...
+  defp apply_changes(data) do
+    data
+    |> put_in(["business", "keywords"], get_keywords_as_list(data["business"]["keywords"]))
+    |> Andi.DatasetSchema.changeset()
+  end
+
+  defp send_dataset_update({:ok, dataset}) do
+    case DatasetValidator.validate(dataset) do
+      :valid -> Brook.Event.send(instance_name(), dataset_update(), :andi, dataset)
+      {:invalid, errors} -> Logger.warn("Invalid dataset: #{IO.inspect({:invalid, errors})}")
```
ManApart: Do we need an IO.inspect or can we just do inspect()
```suggestion
      {:invalid, errors} -> Logger.warn("Invalid dataset: #{inspect({:invalid, errors})}")
```

Same question for a few lines down.

PR #402, opened 2019-12-17
Move odo to the umbrella

LtChae:
This move has negated the need for the odo integration tests. They have either been incorporated into the e2e test or downgraded to unit tests.


.gitignore
```diff
@@ -25,3 +25,5 @@ erl_crash.dump
 
 node_modules/
 .DS_Store
+test/support/minio_data/.minio.sys/*
```
bbrewer-papa-x: i think you can safely delete these lines now that we've changed the way that we pre-load data into the minio container image during testing.

apps/odo/.formatter.exs
```diff
@@ -0,0 +1,5 @@
+# Used by "mix format"
```
bbrewer-papa-x: do we have a formatter for each app in the apps directory or can we get away with a single top-level one?

apps/odo/.gitignore
```diff
@@ -0,0 +1,30 @@
+# The directory Mix will write compiled artifacts to.
```
bbrewer-papa-x: i think we can safely remove this sub-gitignore

apps/odo/config/config.exs
```diff
@@ -0,0 +1,11 @@
+use Mix.Config
+
+config :odo,
+  collector: StreamingMetrics.PrometheusMetricCollector
+
+config :mime, :types, %{
```
bbrewer-papa-x: this `mime` block should be removed now as it will inherit from the top-level config file in the umbrella.

apps/odo/mix.exs
```diff
@@ -0,0 +1,62 @@
+defmodule Odo.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :odo,
+      version: "0.2.0",
```
bbrewer-papa-x: might want to check and make sure we don't already have a release tagged with `0.2.0` and if we do, rev this slightly.

apps/odo/test/unit/odo/angry_case_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  import SmartCity.Event, only: [file_ingest_end: 0, error_file_ingest: 0]
+  alias ExAws.S3
+  alias SmartCity.HostedFile
+
+  test "Individual task failures do not stop others" do
+    start_supervised!(Odo.TestEventHandler)
+
+    # Application.put_env(:odo, :working_dir, "test/support")
+
+    File.cp!("test/support/my-data.shapefile", "/tmp/123.shapefile")
```
bbrewer-papa-x: do this tempdir work without using the `Temp` module to create a platform-independent tempdir that the app has permissions to write to? we had issues with trying to write to `/tmp` when we were moving the integration test to the e2e integration.

apps/odo/test/unit/odo/angry_case_test.exs
```diff
@@ -0,0 +1,74 @@
...
+      type: "error:file:ingest"
+    }
+
+    allow(ExAws.request(any()), return: {:ok, :there})
+
+    Brook.Event.send(Odo.event_stream_instance(), file_ingest_end(), :odo, bad_event)
+    Brook.Event.send(Odo.event_stream_instance(), file_ingest_end(), :odo, good_event)
+
+    SmartCity.TestHelper.eventually(fn ->
+      events = Odo.TestEventHandler.get_events()
```
bbrewer-papa-x: Is there a reason we don't just use `assert_receive` here like several other tests in the umbrella already do?

bbrewer-papa-x: Ex. in `reaper`'s handler_test it does a

```elixir
      assert_receive {:brook_event, %Brook.Event{type: "data:ingest:start", data: dataset}}
```

This seems slightly more straightforward to me at least.

bbrewer-papa-x: Balser/a person needs to update the Brook docs :/

The Test driver will send those `:brook_event` messages. I had to search the repo to confirm this https://github.com/bbalser/brook/search?q=%3Abrook_event&unscoped_q=%3Abrook_event

apps/odo/.formatter.exs
LtChae: We have one per app. Not sure if we can do a top level one or not.

apps/odo/test/unit/odo/angry_case_test.exs
LtChae: We decided to stick with the current implementation. We were not able to use assert_received because the test driver does not propagate events. It requires the test to use `Brook.Test.send`, but the event handler creates new events using `Brook.Event.send`, which is not picked up by the test driver.

apps/odo/test/unit/odo/angry_case_test.exs
LtChae: Moved to use the Temp library

apps/odo/config/config.exs
LtChae: Removed

apps/odo/mix.exs
LtChae: Bumped to 0.3.0

apps/odo/.gitignore
LtChae: Removed

apps/estuary/mix.exs
```diff
@@ -19,15 +19,13 @@ defmodule Estuary.MixProject do
     ]
   end
 
-  # Run "mix help compile.app" to learn about applications.
```
ManApart: Thanks for getting rid of these boiler plate comments.

apps/odo/config/integration.exs
```diff
@@ -0,0 +1,64 @@
+use Mix.Config
+# We should turn this into a repo to share with Reaper
```
ManApart: ^ Do we get this for free now that they're all umbrella?

apps/odo/lib/odo.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Odo do
+  @moduledoc """
+  Convenience functions for the Odo application.
+  """
+
+  def event_stream_instance() do
```
ManApart: It'd be nice if this was closer to our convention in other apps (just `instance()`) but  I understand if that feels out of scope for this card.

jeffgrunewald: this was more or less a backwards compatibility issue. We created reaper's `instance()` function (or equivalent) to return the atom `:reaper_brook` but now we have other non-Brook use cases for an `instance`-type function and we didn't want to break the existing view state by renaming that instance return value when we have a live system using it.

apps/odo/config/integration.exs
jeffgrunewald: yes but that block should theoretically be able to be removed now that we don't have any integration tests for Odo (they should all be covered by unit tests or e2e).

apps/odo/.formatter.exs
jeffgrunewald: we should experiment if those different files will conflict with each other. If they have different values (like line length) and you run the mix command at one level, i suspect you could be undoing the work the formatter did when you ran it at the top level and vice versa.

PR #403, opened 2019-12-18
Updating Forklift to send data:write:complete to event stream after writing data

kennyatpillar:


apps/forklift/lib/forklift/data_writer.ex
```diff
@@ -113,15 +114,16 @@ defmodule Forklift.DataWriter do
   defp do_write(data, dataset) do
     started_data = Enum.map(data, &add_start_time/1)
 
-    retry with: exponential_backoff(100) |> cap(@max_wait_time) do
+    retry with: exponential_backoff(@retry_count) |> expiry(@max_wait) do
```
bbrewer-papa-x: I think we misunderstood this when we edited it. We ultimately want to either set a max number of retries or a max total time so our tests can actually fail. @jdenen - the `cap` usage in this case, was to keep the exponential backoff from growing forever, right?

apps/andi/mix.exs
```diff
@@ -54,7 +54,7 @@ defmodule Andi.MixProject do
       {:placebo, "~> 1.2", only: [:dev, :test, :integration]},
       {:plug_cowboy, "~> 2.1"},
       {:simply_validate, ">= 0.2.0"},
-      {:smart_city, "~> 3.0", override: true},
+      {:smart_city, "~> 3.11"},
```
jmorris-pillar: does this need to be specifically 3.11?  pinning to 3.0 should suffice here.

apps/discovery_streams/mix.exs
```diff
@@ -58,7 +58,7 @@ defmodule DiscoveryStreams.Mixfile do
       {:placebo, "~> 1.2", only: [:dev, :test, :integration]},
       {:redix, "~> 0.10.2"},
       {:sweet_xml, "~> 0.6"},
-      {:smart_city, "~> 3.0"},
+      {:smart_city, "~> 3.11"},
```
jmorris-pillar: same as above

apps/andi/mix.exs
kennyatpillar: We were removing the override true in places it didn't need to be and just put the current version. We'll switch this to 3.0.

apps/forklift/lib/forklift/data_writer.ex
bbrewer-papa-x: This is fixed.

bbrewer-papa-x:
```suggestion
    retry with: exponential_backoff(100) |> cap(@max_wait) |> Stream.take(@retry_count) do
```

apps/forklift/lib/forklift/migrations.ex
```diff
@@ -1,92 +1,46 @@
...
-            acc ++
-              [
-                ["RENAME", key, "old:" <> key],
-                ["EXPIRE", "old:" <> key, "600000"]
-              ]
-          end)
-
-        Redix.pipeline!(client, commands)
+  def migrate_once(completed_flag_name, event_name) do
+    Patiently.wait_for(
```
bbrewer-papa-x: We need to put in a PR on Elsa and/or Brook to get this fixed ASAP

kennyatpillar: Actual issue was having config: where it should have been consumer_config in the integration test. Fixed.

apps/forklift/lib/forklift/data_writer.ex
kennyatpillar: I merged in this suggestion change. Unit tests, integration tests, and e2e all pass.

PR #404, opened 2019-12-18
Fix andi build (maybe)

LtChae:
smartcitiesdata/smartcitiesdata#396

co-authored-by: Jake Prem <jakeprem@gmail.com>


bbrewer-papa-x:
Closing as we have a solution that actually works.

PR #422, opened 2020-01-03
Added support for jpath selectors for json dataset ingests

LtChae:
smartcitiesdata/smartcitiesdata#415

co-authored-by: Jessie Morris <jmorris2@pillartechnology.com>


apps/reaper/test/integration/reaper/reaper_full_test.exs
```diff
@@ -201,6 +203,30 @@ defmodule Reaper.FullTest do
...
+
+      Brook.Event.send(@instance, dataset_update(), :reaper, json_dataset)
+
+      eventually(fn ->
+        results = TestUtils.get_data_messages_from_kafka(topic, @endpoints)
+
+        assert [%{payload: %{"name" => "Fred"}} | [%{payload: %{"name" => "Bob"}} | _]] = results
+      end)
+    end
+
```
mcsearchin: Should we add any tests to document some of the non-happy paths?  Failure to parse or failure to find a node that matches the top-level selector?

apps/reaper/lib/reaper/decoder/json.ex
```diff
@@ -5,6 +5,18 @@ defmodule Reaper.Decoder.Json do
...
   @impl Reaper.Decoder
+
+  def decode({:file, filename}, %{technical: %{topLevelSelector: top_level_selector}} = _dataset)
+      when not is_nil(top_level_selector) do
+    with {:ok, query} <- Jaxon.Path.parse(top_level_selector),
+         {:ok, data} <- json_file_query(filename, query) do
+      {:ok, data}
+    else
+      {:error, error} ->
+        {:error, File.read!(filename), error}
```
NathanielScottStevens: Do we really want to log out the whole file?

LtChae: This maintains the existing functionality. If we don't want to do this that's another discussion I feel.

Issue #325, opened 2019-11-15
Make calendar fields editable 


carriebrinkley:
As a data curator
I want to be able to edit dates that have been selected
To ensure accuracy on datasets

#### Acceptance Criteria: 
- [ ] Calendar fields can be edited 
- [ ] Date format is validated
- [ ] Use built in browser date picker
- [ ] Return an error message if the dataset is not valid

#### Tech Notes: 
- Implementation will be decided upon entry 
- Only includes date (not time)

#### Amigos: 
@jakeclark , @NathanielScottStevens , @carriebrinkley

Issue #328, opened 2019-11-18
Add validation for metadata fields 


carriebrinkley:
As a data curator
I want for information that I submit to be validated
So that I can ensure technical accuracy

#### Acceptance Criteria:
- [ ] Validation is in place

#### Tech Notes:
- Validation that exists should still run in the backend
- Mockup: https://pillartechnology.invisionapp.com/share/UFUH0HI7N84#/screens/393712560

#### Amigos:
@jakeclark , @NathanielScottStevens , @carriebrinkley

NathanielScottStevens:
TODOs
* Get list of all expected validations
* Figure out why empty fields aren't getting persisted

PR #425, opened 2020-01-03
Update Andi Edit dataset

kcarra:
#325 

* Change date fields to use a date input. Browsers will then render a datepicker.
* Date format is validated
* Shows errors from DatasetValidator
* Refactor and fix bugs


apps/andi/lib/andi_web.ex
```diff
@@ -66,6 +66,16 @@ defmodule AndiWeb do
     end
   end
 
+  def live_view do
+    quote do
+      use Phoenix.LiveView
+      import AndiWeb.ErrorHelpers
+      import Phoenix.HTML.Form
```
NathanielScottStevens: I'm not sure that I'd import 'Form' and 'Link' since they won't be used in all live_view components

apps/andi/lib/andi_web/input_schemas/metadata.ex
```diff
@@ -0,0 +1,102 @@
...
+    |> validate_format(:contactEmail, @email_regex, message: "Email is invalid.")
+    |> validate_required([:issuedDate], message: "Release Date is required.")
+    |> validate_required([:license], message: "License is required.")
+    |> validate_required([:publishFrequency], message: "Publish Frequency is required.")
+    |> validate_required([:orgTitle], message: "Organization is required.")
+    |> validate_required([:sourceFormat], message: "Format is required.")
+  end
+
+  defp get_business(map) when is_map(map) do
+    Map.take(map, Map.keys(@business_fields))
```
NathanielScottStevens: Nice! Didn't know about `Map.take`

apps/andi/lib/andi_web/input_schemas/metadata.ex
```diff
@@ -0,0 +1,102 @@
...
+    dataset
+    |> Map.put(:business, business)
+    |> Map.put(:technical, technical)
+  end
+
+  def new_changeset(params \\ %{}), do: changeset({%{}, @types}, params)
+
+  def changeset(changeset, params \\ %{}) do
+    params =
+      case params do
```
NathanielScottStevens: I would stick this case statement in `update_keywords_params` to keep this function at a higher level of abstraction.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -1,178 +1,179 @@
 defmodule AndiWeb.EditLiveView do
-  use Phoenix.LiveView
-  alias Phoenix.HTML.Form
+  use AndiWeb, :live_view
+
   alias Phoenix.HTML.Link
```
NathanielScottStevens: If you do keep `Phoenix.HTML.Link` in the live_view macro you won't need this here.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -1,178 +1,179 @@
...
+          <%= error_tag(f, :private) %>
+        </div>
         <div class="metadata-form__cancel-btn">
           <%= Link.button("Cancel", to: "/", method: "get", class: "btn btn--cancel") %>
         </div>
         <div class="metadata-form__save-btn">
-          <%= Link.button("Next", to: "/", method: "get", id: "next-button", class: "btn btn--next") %>
-          <%= Form.submit("Save", id: "save-button", class: "btn btn--save") %>
+          <%= unless is_nil(@validation_errors) do %>
+            <div class="metadata__error-message">
```
NathanielScottStevens: Duplicating the error messages at the bottom of the page feels wonky to me. I'd take it out unless Jake asked for it.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -1,178 +1,179 @@
...
-  def handle_event("save", %{"dataset_schema" => dataset_schema}, socket) do
-    change = apply_changes(dataset_schema)
+  def handle_event("validate", %{"metadata" => form_data}, socket) do
+    socket = reset_save_success(socket)
 
-    case change.valid? do
-      true ->
-        schema = Ecto.Changeset.apply_changes(change)
-        original_dataset = socket.assigns.dataset
+    new_changeset =
```
NathanielScottStevens: Nice simplification here

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -1,178 +1,179 @@
...
+      form_data
+      |> Metadata.new_changeset()
+      |> Map.put(:action, :update)
```
NathanielScottStevens: What does changing the action here do?

apps/reaper/1234
```diff
@@ -0,0 +1,2 @@
+one,two,three
```
NathanielScottStevens: !!!???

NathanielScottStevens: The same goes for the rest of these reaper files

apps/reaper/cameras.xml
```diff
@@ -0,0 +1 @@
...th St]]></Location><Latitude>40.801972</Latitude><Longitude>-81.398287</Longitude><CameraFeeds><CameraFeed><Direction>PTZ</Direction><Type>Pan/Tilt/Zoom</Type><Description><![CDATA[I-77 at 4th St.  This is a Pan/Tilt/Zoom camera.  The camera view direction is variable and may not be relative to its current position. If this camera image is not currently updating or displaying properly, ODOT personnel are aware of the situation and are working to address the issue.]]></Description><SmallImage><![CDATA[http://cmhimg01.dot.state.oh.us/images/CLE/CLE120a.jpg]]></SmallImage><LargeImage><![CDATA[http://cmhimg01.dot.state.oh.us/images/CLE/CLE120a.jpg]]></LargeImage><ImageUpdateInterval>5</ImageUpdateInterval></CameraFeed></CameraFeeds></CameraSite><CameraSite><Location><![CDATA[I-75 at Stanley Ave]]></Location><Latitude>39.788843</Latitude><Longitude>-84.183595</Longitude><CameraFeeds><CameraFeed><Direction>PTZ</Direction><Type>Pan/Tilt/Zoom</Type><Description><![CDATA[I-75 at Stanley Ave.  This is a Pan/Tilt/Zoom camera.  The camera view direction is variable and is indicated in the top left corner of this image (N, S, E, W, NE, NW, SE, SW).   If this camera image is not currently updati
```
ManApart: Should these reaper files be part of this andi card?

apps/andi/lib/andi_web/live/edit_live_view.ex
kcarra: These are the error messages that come from the DatasetValidator, not the form errors. We'll double check the UX with Jake in the exit, but expect that more work will be done on this with the validation card (#328)

apps/andi/lib/andi_web/input_schemas/metadata.ex
kcarra: We refactored this. We kept that in the changeset because the concern felt like it fell under the changesets responsibility to validate and transform raw input data.

But we split it to be more clearly separate the logic for handling a list of keywords from the struct from the logic to handle the raw string from the form.

apps/andi/lib/andi_web/live/edit_live_view.ex
kcarra: Putting an action onto the changeset will cause the `Phoenix.HTML` code to show errors on the form. This keeps a whole page of errors from showing up before the user has even edited a field. Not having an action on the changeset is the reason the code had to access `form.source.errors` on the `ErrorHelper` module.

apps/reaper/1234
kcarra: Got them in the rebase

apps/reaper/cameras.xml
kcarra: Got them in the rebase

apps/andi/lib/andi_web/input_schemas/metadata.ex
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @technical_fields %{
+    private: :boolean,
+    sourceFormat: :string
+  }
+
+  @types Map.merge(@business_fields, @technical_fields)
+
+  @email_regex ~r/^[A-Za-z0-9._%+-+']+@[A-Za-z0-9.-]+\.[A-Za-z]+$/
```
NathanielScottStevens: This looks like it's outside the scope of your card. Also, I don't see it being used anywhere.

apps/andi/lib/andi_web.ex
jakeprem: These will generally be used in all components.

This is basically in line with what you would import in a non-live Phoenix View

apps/andi/lib/andi_web/input_schemas/metadata.ex
jakeprem: We just moved it out from the place

PR #431, opened 2020-01-08
update forklift to use single utility for starting and stopping data readers

bbrewer-papa-x:
smartcitiesdata/smartcitiesdata#430

co-authored-by: Timothy Regan <tregan@pillartechnology.com>

### DIFF
--- a/apps/forklift/lib/forklift/data_reader_helper.ex
+++ b/apps/forklift/lib/forklift/data_reader_helper.ex
@@ -1,0 +1,32 @@
+defmodule Forklift.DataReaderHelper do
+  @moduledoc """
+  Simple wrapper around the reader init and terminate behaviors
+  """
+  @reader Application.get_env(:forklift, :data_reader)
+
+  def init(dataset) do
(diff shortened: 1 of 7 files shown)


apps/forklift/lib/forklift/data_reader_helper.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Forklift.DataReaderHelper do
```
jeffgrunewald: Does this need to be "DataReaderHelper" or can it just be "DataReader" = "the forklift implementation of the pipeline data reader"?

bbrewer-papa-x: We thought about that, but it's technically not an implementation of the behavior so we wanted to make that clear(er)

jeffgrunewald: That’s fair. I just don’t like the “Helper” part. For some reason it feels like a bolt-on

PR #433, opened 2020-01-08
Add pipeline

sukaant:


apps/e2e/test/e2e_test.exs
```diff
@@ -215,6 +215,16 @@ defmodule E2ETest do
...
       end)
     end
+
+    test "events have been stored in estuary" do
+      table = Application.get_env(:estuary, :table_name)
+
+      eventually(fn ->
+        [[actual]] = query("SELECT count(1) FROM #{table}")
+
+        assert actual > 0
```
jmorris-pillar: Anyway to have this be a specific number instead of > 0?

ManApart: We don't know how many events the system will create (and this will probably increase over time). In my opinion all we care about is at least one event has been written to the table.
I'm open to alternatives.

apps/estuary/config/test.exs
```diff
@@ -1,6 +1,17 @@
 use Mix.Config
 
+config :logger, level: :warn
```
jmorris-pillar: This will likely get overridden btw.  Do not depend on this being set this way, per the arts and crafts thing i brought up.

apps/estuary/config/test.exs
```diff
@@ -1,6 +1,17 @@
...
+
 config :estuary,
-  elsa_endpoint: nil
+  topic_reader: MockReader,
+  table_writer: MockTable,
+  endpoints: [localhost: 9092],
```
jmorris-pillar: is this still needed for the mocks?  Just curious.

apps/estuary/lib/estuary/application.ex
```diff
@@ -2,70 +2,12 @@ defmodule Estuary.Application do
...
-
-  defp get_children do
-    if elsa_endpoint() != nil do
-      [
-        {Elsa.Supervisor, elsa_options()}
-      ]
-    else
-      []
-    end
+    |> List.flatten()
```
jmorris-pillar: Do we need to flatten here?

apps/e2e/test/e2e_test.exs
NathanielScottStevens: Do we have any evidence that the count wasn't greater than 0 before this test?

apps/estuary/config/test.exs
ManApart: JMS. I think we're just doing this 1) to be consistent and 2) for warn when running just our tests, but I don't think we're relying on it.

apps/estuary/lib/estuary/application.ex
```diff
@@ -2,70 +2,12 @@ defmodule Estuary.Application do
...
-      group_consumer: [
-        group: "estuary-consumer-group",
-        topics: [event_stream_topic()],
-        handler: Estuary.MessageHandler,
-        config: [
-          begin_offset: :earliest,
-          offset_reset_policy: :reset_to_earliest
-        ]
-      ]
+      {DynamicSupervisor, strategy: :one_for_one, name: Estuary.Dynamic.Supervisor},
```
jmorris-pillar: Do we use this?  May remove this comment as i scroll down.

apps/estuary/config/test.exs
ManApart: Dunno. We'll check and remove it if our tests still pass.

apps/e2e/test/e2e_test.exs
ManApart: Yup. I'll come over.

apps/estuary/config/config.exs
```diff
@@ -1,8 +1,13 @@
 use Mix.Config
 
+connection = :estuary_elsa
```
NathanielScottStevens: Why is this a variable?

apps/e2e/test/e2e_test.exs
bbrewer-papa-x: Yeah, we're got foot gun blasted by a specific count check in e2e just today, so I'd advise we be a little foot gun shy about that

apps/estuary/test/unit/data_reader_test.exs
```diff
@@ -0,0 +1,19 @@
...
+  use ExUnit.Case
+
+  import Mox
+
+  alias Estuary.DataReader
+
+  setup :set_mox_global
+  setup :verify_on_exit!
+
+  test "should read the event" do
```
jmorris-pillar: Is this name correct?

apps/estuary/lib/estuary/data_writer.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  def init(args) do
+    :ok =
+      args
+      |> @table_writer.init()
+  rescue
+    e -> {:error, e, "Presto Error"}
+  end
+
+  @impl Pipeline.Writer
```
NathanielScottStevens: Duplicated from above (nathaniel)

apps/estuary/lib/estuary/data_writer.ex
```diff
@@ -0,0 +1,51 @@
...
+  alias Estuary.Datasets.DatasetSchema
+
+  @behaviour Pipeline.Writer
+
+  @table_writer Application.get_env(:estuary, :table_writer)
+
+  alias Estuary.Datasets.DatasetSchema
+
+  @impl Pipeline.Writer
+  @doc """
```
NathanielScottStevens: I don't think we need docs for the methods in the module

apps/estuary/test/unit/data_reader_test.exs
ManApart: ^ We're going to go ahead and delete this guy.

apps/estuary/lib/estuary/data_writer.ex
```diff
@@ -0,0 +1,51 @@
...
+  Ensures a table exists using `:table_writer` from
+  Estuary's application environment.
+  """
+
```
NathanielScottStevens: I don't think there should be a newline between the @impl and docs and the method they're annotating

apps/estuary/lib/estuary/data_writer.ex
```diff
@@ -0,0 +1,51 @@
...
+  rescue
+    e -> {:error, e, "Presto Error"}
+  end
+
+  @impl Pipeline.Writer
+  @doc """
+  Writes data to PrestoDB and Kafka using `:table_writer` from
+  Estuary's application environment.
+  """
+
```
NathanielScottStevens: Same thing here

apps/e2e/test/e2e_test.exs
ManApart: We re-ran the test without calling data writer and our test timed out/failed because it found 0 events inserted.

jmorris-pillar: I figured as much.  Fine with me for now

apps/estuary/lib/estuary/datasets/dataset_schema.ex
```diff
@@ -0,0 +1,54 @@
...
+      },
+      %{
+        name: "type",
+        type: "string",
+        description: "N/A"
+      }
+    ]
+  end
+
+  def make_datawriter_payload(event) do
```
NathanielScottStevens: I would move this to `DataWriter` since it's only used there

apps/estuary/lib/estuary/init_server.ex
```diff
@@ -0,0 +1,35 @@
...
+  def init(_) do
+    with :ok <- create_table(),
+         :ok <- DataReader.init(),
+         pid <- Process.whereis(Pipeline.DynamicSupervisor) do
+      Process.monitor(pid)
+      {:ok, %{pipeline: pid}}
+    end
+  end
+
+  def handle_info({:DOWN, _, _, pid, _}, %{pipeline: pid}) do
```
NathanielScottStevens: I think I'm missing context on why it's necessary to babysit Pipeline in this way and go through the init steps again when it goes down.

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -2,40 +2,25 @@ defmodule Estuary.MessageHandler do
...
-  defp process_message(message) do
-    case Jason.decode(message.value) do
-      {:ok, %{"author" => _, "create_ts" => _, "data" => _, "type" => _} = event} ->
-        do_insert(message, event)
-
-      {_, term} ->
-        process_error(message, term)
-    end
+  defp error_dead_letter({:error, event, reason}) do
+    DeadLetter.process("Unknown", event, "estuary", reason: reason)
```
NathanielScottStevens: Are we DLQing these event messages?

apps/estuary/test/unit/data_reader_test.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Estuary.DataReaderTest do
```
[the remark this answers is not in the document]
jeffgrunewald: i believe the forklift reader init test is at least asserting that the function was called with valid args and then returning them to the test pid for assertion.

apps/estuary/test/unit/test_helper.exs
```diff
@@ -1 +1,4 @@
-ExUnit.start()
+Application.ensure_all_started(:mox)
+Application.ensure_all_started(:dead_letter)
```
jeffgrunewald: are these necessary?

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -2,40 +2,25 @@ defmodule Estuary.MessageHandler do
...
-  require Logger
+  alias Estuary.DataWriter
 
   def handle_messages(messages) do
-    Enum.each(messages, fn message -> process_message(message) end)
+    messages
+    |> Enum.each(fn message ->
+      message.value
+      |> Jason.decode!()
+      |> DataWriter.write()
```
jeffgrunewald: have we verified that DataWriter will batch when it writes into Presto for you?
per Matt's previous updates, it sounds like we are expecting this pipeline refactor to solve that problem of estuary only writing a single record at a time so unless DataWriter is doing that for you out of the box, this isn't going to allow anything to get deployed to prod.

apps/estuary/test/unit/data_writer_test.exs
```diff
@@ -0,0 +1,45 @@
...
+      %{
+        "author" => author,
+        "create_ts" => time_stamp,
+        "data" => dataset,
+        "forwarded" => false,
+        "type" => "data:ingest:start"
+      }
+      |> DataWriter.write()
+
+    assert :ok == actual_value
```
NathanielScottStevens: Both of these tests should probably verify that the appropriate pipeline writer functions were called with the expected args. (Tony)

apps/estuary/lib/estuary/application.ex
jeffgrunewald: i don't think i see anything using this dynamic supervisor. why is it here in your supervision tree?

apps/estuary/lib/estuary/application.ex
jeffgrunewald: this isn't necessary if you don't have a function here that _could potentially_ generate its own list.

apps/estuary/lib/estuary/message_handler.ex
sukaant: Our current understanding is yes. We'll bring it up in exit to confirm it's the right decision.

apps/estuary/lib/estuary/init_server.ex
jeffgrunewald: pipeline will get restarted but the pipeline supervisor is a dynamic supervisor so if it goes down, it will come back up with no memory of the things it was supposed to be running so we need some other process to reschedule our readers/writers to come back up under it.

apps/estuary/test/unit/event_table_test.exs
```diff
@@ -1,48 +0,0 @@
-defmodule Estuary.InsertTest do
```
dunson062786: Why did you delete this file?

apps/estuary/config/config.exs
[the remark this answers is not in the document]
sukaant: Nothing, lol. We updated it.

apps/estuary/test/unit/data_reader_test.exs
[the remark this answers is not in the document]
sukaant: We have deleted this test.

apps/estuary/test/unit/test_helper.exs
sukaant: Yes, they are necessary, since we are doing the unit test with no start but we still require mox and dead_letter

apps/estuary/lib/estuary/message_handler.ex
sukaant: This guy is still writing only 1 message at a time to presto, despite handling/getting batches of messages from elsa. The next card we play will be WRITING those messages as a batch to presto.

apps/estuary/lib/estuary/application.ex
sukaant: Deleted.

apps/estuary/lib/estuary/application.ex
sukaant: We removed it.

apps/estuary/test/unit/event_table_test.exs
sukaant: These tests are redundant now. Either from other unit tests, the E2E test, or the pipeline tests.

apps/estuary/mix.exs
```diff
@@ -28,13 +28,16 @@ defmodule Estuary.MixProject do
 
   defp deps do
     [
-      {:credo, "~> 1.1", only: :dev, runtime: false},
+      {:brook, "~> 0.4.0"},
```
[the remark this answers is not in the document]
sukaant: We were using it for testing with Brook.Events, but we've deleted those tests, so we can get rid of it again.

apps/estuary/lib/estuary/init_server.ex
```diff
@@ -0,0 +1,35 @@
...
+    :ok = create_table()
+    :ok = DataReader.init()
+    {:noreply, %{pipeline: Process.whereis(Pipeline.DynamicSupervisor)}}
```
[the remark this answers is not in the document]
ManApart: Yup, you're right. We'll remove it from the handle_info call.

apps/estuary/config/test.exs
sukaant: Removed endpoints.

apps/estuary/config/test.exs
```diff
@@ -1,6 +1,17 @@
...
+  instance: :estuary,
+  handler: Estuary.MessageHandler,
+  connection: :estuary_elsa
 
-config :logger, level: :warn
+config :estuary, :dead_letter,
```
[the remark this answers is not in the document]
sukaant: Removed

apps/estuary/lib/estuary/data_reader.ex
```diff
@@ -0,0 +1,33 @@
...
+  @impl Pipeline.Reader
+  def init(_opts \\ []) do
+    :ok =
+      reader_args()
+      |> @topic_reader.init()
+  end
+
+  @impl Pipeline.Reader
+  def terminate(_opts \\ []) do
+    :ok
```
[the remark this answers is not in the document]
ManApart: My understanding at the moment (I may have missed something) is that terminate was implemented because the behavior requires it. That being said, we're never calling terminate and so it wasn't doing anything. We'll go ahead and have it call terminate on the writer just so it's consistent with the init / is setup for if/when we do end up using it..

apps/estuary/lib/estuary/init_server.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  def start_link(opts) do
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, [], name: name)
+  end
+
+  def init(_) do
+    with :ok <- create_table(),
+         :ok <- DataReader.init(),
+         pid <- Process.whereis(Pipeline.DynamicSupervisor) do
```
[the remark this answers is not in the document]
ManApart: I didn't realize that Process.whereis could return something other than a pid, but both port and nil make a lot of sense as alternatives. I see an example in Valkyrie, though it looks like forklift also needs a guard.

I'd like to better understand what we're doing here. It seems like we want to get a pid so when we get a down message we can match against our pid and restart if needed. That being said, wouldn't a guard clause mean we'd kill our init process if we got a port back (because the supervisor was on another node etc)? I guess that's fine since we NEED the exact pid, I just feel like I could be missing some pieces.

apps/estuary/lib/estuary/init_server.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def handle_info({:DOWN, _, _, pid, _} = msg, %{pipeline: pid} = state) do
+    with :ok <- DataReader.init(),
+         pid when is_pid(pid) <- Process.whereis(Pipeline.DynamicSupervisor) do
+      Process.monitor(pid)
+      {:noreply, %{pipeline: pid}}
+    else
+      _ ->
+        Process.sleep(100)
```
ManApart: I'm not a fan of how we did this, but it technically passes the tests.
I could see us hanging here indefinitly retrying to connect to pipeline, unless another part of the app/otp says we've tried for too long. Would like to talk with Jeff or someone about this.

apps/estuary/mix.exs
```diff
@@ -28,13 +28,15 @@ defmodule Estuary.MixProject do
 
   defp deps do
     [
-      {:credo, "~> 1.1", only: :dev, runtime: false},
+      {:credo, "~> 1.1", only: [:dev, :test], runtime: false},
```
jeffgrunewald: you shouldn't need credo in the test environment. if you're getting a warning about that because of another app in the pipeline, try deleting it from test in that app as well.

apps/estuary/mix.exs
```diff
@@ -28,13 +28,15 @@ defmodule Estuary.MixProject do
...
       {:dead_letter, in_umbrella: true},
       {:distillery, "~> 2.1"},
       {:divo, "~> 1.1", only: [:dev, :integration]},
       {:divo_kafka, "~> 0.1.5", only: [:dev, :integration]},
       {:elsa, "~> 0.10.0"},
-      {:mock, "~> 0.3", only: [:test, :integration], runtime: false},
+      {:jason, "~> 1.1"},
```
jeffgrunewald: where is jason being used? if it's just in tests can that be scoped to just those tests or does that cause warnings about other umbrella apps that use it in their prod code?

apps/estuary/test/unit/data_writer_test.exs
```diff
@@ -0,0 +1,39 @@
+defmodule Estuary.DataWriterTest do
+  use ExUnit.Case
+  use Placebo
+
+  alias Estuary.Datasets.DatasetSchema
+  alias Estuary.DataWriter
+  alias SmartCity.TestDataGenerator, as: TDG
+
+  test "should insert event to history table" do
+    allow(MockTable.write(any(), any()), return: :ok)
```
jeffgrunewald: if you're using placebo, you shouldn't need to be mocking the mock. the table writer should be testing it's ability to write to presto so your implementation of table writer should be able us Mox to mock the table writer module and assert that the right things were passed to the table writer `write` function here.

apps/estuary/test/unit/init_server_test.exs
```diff
@@ -0,0 +1,35 @@
...
+  alias Estuary.InitServer
+
+  setup :set_mox_global
+  setup :verify_on_exit!
+
+  test "should initialize topic reader and table writer on the application startup" do
+    expect(MockTable, :init, fn _ -> :ok end)
+    expect(MockReader, :init, fn _ -> :ok end)
+
+    assert {:ok, _} = InitServer.start_link(name: :foo)
```
jeffgrunewald: i'd remove the name since it's being ignored now.

apps/estuary/test/unit/init_server_test.exs
```diff
@@ -0,0 +1,35 @@
+defmodule Estuary.InitServerTest do
+  use ExUnit.Case
+  use Placebo
```
jeffgrunewald: you can get rid of this `use Placebo` here.

apps/estuary/test/unit/message_handler_test.exs
```diff
@@ -0,0 +1,84 @@
...
+
+  import Mox
+  import Assertions
+
+  alias Estuary.MessageHandler
+  alias SmartCity.TestDataGenerator, as: TDG
+  alias DeadLetter.Carrier.Test, as: Carrier
+
+  setup :set_mox_global
+  setup :verify_on_exit!
```
jeffgrunewald: i think you can remove all these `Mox` things (including the import above) since you don't seem to be using them anywhere in this file.

apps/estuary/mix.exs
ManApart: We're using it in the data_writer_test and message_handler_test. However, I removed it from the mix and did a mix deps.get and they're still passing, so I'll remove it from the mix.

apps/estuary/mix.exs
ManApart: Good call

apps/estuary/mix.exs
jeffgrunewald: You’d still have it unless you removed it from the lock file and cleaned the deps with `mix deps.unlock —unused` and `mix deps.clean —unused`

PR #437, opened 2020-01-10
Use Default Schema

ManApart:


apps/estuary/config/config.exs
```diff
@@ -2,8 +2,8 @@ use Mix.Config
 
 config :estuary,
   topic: "event-stream",
-  schema_name: "event_stream",
-  table_name: "history",
+  schema_name: "default",
+  table_name: "event_stream",
```
smillardHNTB: assuming this is being moved to the default namespace - should the table name maybe be something like os_internal__event_stream_history so its title format matches other tables?

apps/estuary/config/integration.exs
```diff
@@ -7,7 +7,7 @@ config :prestige,
   headers: [
     user: "estuary",
     catalog: "hive",
-    schema: "event_stream"
+    schema: "default"
```
jeffgrunewald: if you've got this setting in `config.exs` and you _dont_ override it anywhere, you shouldn't need to define it anywhere else, including in `runtime.exs`

apps/estuary/config/config.exs
jmorris-pillar: I think that could make it confusing and look like a dataset.  The flair table, operational_stats is named as written in this sentence.

ManApart: I'm good either way. We should do the proper process and take this to a committee of at least 5 people.

Seriously, I'm cool with whatever. Sounds like this is what it was before we decided to do the other schema. Unless there is strong push back I'm just going with `event_stream`.

smillardHNTB: Good point - but maybe this could be a private dataset so it could be queried by the API? That could be useful in the future.

PR #444, opened 2020-01-13
Batch commits

ManApart:


apps/estuary/lib/estuary/data_writer.ex
```diff
@@ -25,34 +25,48 @@ defmodule Estuary.DataWriter do
...
-  defp make_datawriter_payload(event) do
-    [
-      %{
-        payload: %{
-          "author" => event["author"],
-          "create_ts" => event["create_ts"],
-          "data" => event["data"],
-          "type" => event["type"]
-        }
+  defp make_payload(%{
```
kcarra: would it make sense to make this a struct and just `new` it up? Might make the pattern match a bit cleaner as well

ManApart: I'd prefer to just use the map here since we're not 're-using' the struct and we'd need to define fields etc. We discussed a bit so I'm going to resolve.

PR #459, opened 2020-01-15
Version Bump

sukaant:


apps/estuary/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Estuary.MixProject do
   def project do
     [
       app: :estuary,
-      version: "0.4.0",
+      version: "0.5.1",
```
ManApart: Straight from 4.0 to 5.1?

sukaant: Already released 5.0 and missed to change this file

PR #460, opened 2020-01-16
Andi consolidate validation

mcsearchin:


apps/andi/lib/andi/input_schemas/dataset_input.ex
```diff
@@ -0,0 +1,142 @@
...
+  ]
+
+  @email_regex ~r/^[A-Za-z0-9._%+-+']+@[A-Za-z0-9.-]+\.[A-Za-z]+$/
+  @no_dashes_regex ~r/^[^\-]+$/
+
+  def business_keys(), do: Map.keys(@business_fields)
+  def technical_keys(), do: Map.keys(@technical_fields)
+  def all_keys(), do: Map.keys(@types)
+
+  def get_display_name(field_key), do: DisplayNames.get(field_key)
```
NathanielScottStevens: This seems like it will be confusing for errors that come through on the API. I'm wondering if there's a way to do this message conversion on the view (Maybe in the error_view?).

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -72,18 +69,29 @@ defmodule Andi.InputSchemas.InputConverter do
...
 
+  defp keywords_to_list(keywords) when is_list(keywords), do: keywords
+
   defp date_to_iso8601_datetime(date) do
     time_const = "00:00:00Z"
 
     "#{Date.to_iso8601(date)} #{time_const}"
   end
+
+  defp fix_modified_date(map) do
```
NathanielScottStevens: I still think it would be cleaner to make `modifiedDate` a string type and then add a validation that ensures it's date. But I could be missing something.

mcsearchin: The main issue we ran into is that I *think* the phoenix form is expecting this field to be a date, and we're basically converting everything down to the types required by the form in order to make it consistent.  We can verify that the date input on the form will/won't work with a string field in the changeset.

apps/andi/lib/andi/input_schemas/dataset_input.ex
mcsearchin: I guess I'd only argue that it won't be confusing because the format of the response will still look like this:
```
{ "errors": { "contactEmail": ["Please provide a valid maintainer email."] } }
```

apps/andi/lib/andi/input_schemas/input_converter.ex
NathanielScottStevens: Oh, I see

apps/andi/lib/andi/input_schemas/dataset_input.ex
NathanielScottStevens: Checkout AndiWeb.ErrorHelpers

mcsearchin: Trying to think about how to build these messages in the view... we'd need to create some kind of mapping from error messages from the changeset validation to error message on the UI.  Seems like a layer of abstraction I'd prefer to avoid for now.

apps/andi/lib/andi/input_schemas/dataset_input.ex
```diff
@@ -110,26 +96,17 @@ defmodule Andi.InputSchemas.DatasetInput do
...
-        add_error(changeset, :schema, "#{get_display_name(:schema)} cannot be empty.")
-
-      nil ->
-        add_error(changeset, :schema, "Please enter a valid #{get_downcased_display_name(:schema)}.",
-          validation: :required
-        )
-
-      _ ->
-        validate_schema_internals(changeset)
+      [] -> add_error(changeset, :schema, "cannot be empty")
```
NathanielScottStevens: Looks like the formatter changed these back?

mcsearchin: No, that was probably me.  😝 We ran the formatter again.

PR #463, opened 2020-01-20
Validation performance adjustment

mcsearchin:


apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -156,11 +151,11 @@ defmodule AndiWeb.EditLiveView do
           {:noreply, assign(socket, changeset: changeset)}
       end
     else
-      {:noreply, assign(socket, changeset: %{changeset | action: :save})}
+      {:noreply, assign(socket, changeset: %{changeset | action: :save}, validation_errors: changeset.errors)}
```
bbrewer-papa-x: Why are you passing in the errors, but not using them?

mcsearchin: Good point.  Changing this to a flag.

PR #470, opened 2020-01-22
Add discovery api

kcarra:
Squashed commits for all changes we made to integrate discovery_api into the umbrella excluding the files that were copied https://github.com/smartcitiesdata/smartcitiesdata/pull/470/commits/7bc2c3798b48177b7299ac7694472aae74569a0b


apps/odo/mix.exs
```diff
@@ -30,7 +30,7 @@ defmodule Odo.MixProject do
       {:credo, "~> 1.0", only: :dev, runtime: false},
       {:distillery, "~> 2.1"},
       {:ex_aws, "~> 2.1"},
-      {:ex_aws_s3, "~> 2.0"},
+      {:ex_aws_s3, "~> 2.0", [env: :prod, git: "https://github.com/ex-aws/ex_aws_s3", ref: "6b9fdac73b62dee14bffb939965742f2576f2a7b"]},
```
mcsearchin: What happened here?  Do we need this very specific version that isn't available from hex?

apps/discovery_api/config/config.exs
```diff
@@ -37,7 +37,7 @@ config :discovery_api, DiscoveryApi.Quantum.Scheduler,
 
 config :mime, :types, %{
   "application/zip" => ["zip", "shp", "shapefile"],
-  "application/gtfs" => ["pb"]
+  "application/gtfs+protobuf" => ["gtfs"]
```
mcsearchin: Do we know how this will impact Discovery UI?

apps/odo/mix.exs
kcarra: Yeah, it is needed by discovery_api in order to use streaming. Still pending release. I'll add a comment to make that clear.

apps/discovery_api/config/config.exs
bbrewer-papa-x: Yeah, I'm interested to know. Probably not your fault if it breaks stuff b/c our mime type stuff is a dumpster show

apps/discovery_api/mix.exs
```diff
@@ -58,7 +62,7 @@ defmodule DiscoveryApi.Mixfile do
       {:phoenix_pubsub, "~> 1.0"},
       {:nanoid, "~> 2.0"},
       {:placebo, "~> 1.2.2", only: [:dev, :test]},
-      {:plug_cowboy, "~> 1.0"},
+      {:plug_cowboy, "~> 2.1"},
```
bbrewer-papa-x: Have we confirmed if the hosted file download bug (memory overflow) is present with this? I don't think we have enough of them for it to matter if not.

apps/odo/mix.exs
bbrewer-papa-x: Yeah, also, this is the one that overflows cowboy when you try to stream hosted files, btw

apps/discovery_api/config/config.exs
kcarra: the issue is the TDG doesn't accept `application/gtfs` anymore so if we keep it how it is then the TDG returns `application/octet-stream` or whatever the default sourceFormat is. I did grep for any datasets in prod, staging and dev with the `application/gtfs` source format and it doesn't look like there are any.

apps/discovery_api/test/unit/discovery_api/data/mapper_test.exs
```diff
@@ -28,7 +28,7 @@ defmodule DiscoveryApi.Data.MapperTest do
         [%{business: %{conformsToUri: nil}}, [:conformsToUri], "https://project-open-data.cio.gov/v1.1/schema/"],
         [%{business: %{conformsToUri: "overridden"}}, [:conformsToUri], "https://project-open-data.cio.gov/v1.1/schema/"],
         [%{technical: %{sourceFormat: "application/json"}}, [:fileTypes], ["JSON"]],
-        [%{technical: %{sourceFormat: "application/gtfs"}}, [:fileTypes], ["JSON"]],
+        [%{technical: %{sourceFormat: "application/gtfs+protobuf"}}, [:fileTypes], ["GTFS"]],
```
LtChae: I would expect fileTypes here to have "JSON", not GTFS. I believe this was covered with an override somewhere in the bit that generates the file types.

PR #471, opened 2020-01-23
Url edit/test section for Andi

LtChae:
Where possible, I have tried to maintain the css patterns ~~@NathanielScottStevens~~ Mr. President has left us.

The body display AC was removed as HEAD requests do not return a body, even on a bad http code.


LtChae:
Currently working on additional url test tests.

apps/andi/assets/css/app.scss
```diff
@@ -76,6 +76,14 @@ body {
   margin-right: 0;
 }
 
+.btn--test {
+  @extend .btn--cancel;
```
NathanielScottStevens: This `extend` violates the pattern of BEM, though I've noticed it's been used in other places. I think the inheritance going on here will just become increasingly confusing vs just having blocks and modifiers. Your call on whether to clean this up or not.

apps/andi/assets/css/edit.scss
```diff
@@ -6,11 +6,28 @@
...
+}
+
+.edit-button-group {
+  @extend .form-grid;
   grid-template-areas:
+    "cancel-btn save-btn";
+}
+
+.metadata-form {
+  @extend .form-grid;
```
NathanielScottStevens: Same thing with the `extends` here

apps/andi/lib/andi/input_schemas/display_names.ex
```diff
@@ -23,6 +23,7 @@ defmodule Andi.InputSchemas.DisplayNames do
     schema: "Schema",
     sourceFormat: "Source Format",
     sourceType: "Source Type",
+    sourceUrl: "Source Url",
```
mcsearchin: Not sure this is visible anywhere yet, but it probably should be `Source URL`.

apps/andi/assets/css/edit.scss
```diff
@@ -35,6 +59,29 @@
   color: $color-success;
 }
 
+#test-status {
```
NathanielScottStevens: Using ids feels wrong to me but I won't fight in that war.

apps/andi/assets/css/edit.scss
```diff
@@ -35,6 +59,29 @@
...
+  display: inline-block;
+  margin-left: 1rem;
+}
+
+#test-status__code {
+  margin-right: 1rem;
+}
+
+.test-status__good {
```
NathanielScottStevens: This looks like a modifier so it should be `--good`

apps/andi/assets/css/edit.scss
```diff
@@ -35,6 +59,29 @@
...
+  color: $color-success
+}
+
+.test-status__bad {
+  color: $color-error
```
NathanielScottStevens: `--bad`

apps/andi/assets/css/edit.scss
```diff
@@ -51,5 +98,19 @@
...
+
+@include area(edit-button-group, cancel-btn);
+@include area(edit-button-group, save-btn);
+@include area(edit-button-group, test-btn);
+
+@include area(url-form, source-url);
+@include area(url-form, test-section);
+
+
+.box-header {
```
NathanielScottStevens: This belongs in the app.scss file if it's meant to be generic, otherwise I would make it an element of the edit-page block (`edit-page__box-header`)

apps/andi/lib/andi/input_schemas/dataset_input.ex
```diff
@@ -51,7 +52,8 @@ defmodule Andi.InputSchemas.DatasetInput do
     :private,
     :publishFrequency,
     :sourceFormat,
-    :sourceType
+    :sourceType,
+    :sourceUrl
```
mcsearchin: Can you add to the `data_test` that we have for required fields in `dataset_input_test`?

LtChae: I believe I did. Can you point out the specific test?

mcsearchin:
I think we talked about this when we defined the card, but just to be clear... do we want to allow the user to save their dataset when the URL test fails?

apps/andi/assets/css/edit.scss
LtChae: Not sure how BEM works in this case. It's a modifier to a "generic" element that's only ever part of a form group.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -140,11 +168,12 @@ defmodule AndiWeb.EditLiveView do
 
     if changeset.valid? do
       changes = Ecto.Changeset.apply_changes(changeset)
+      dataset = InputConverter.restruct(changes, original_dataset)
```
NathanielScottStevens: Was this just a boy scout change?

LtChae:
> I think we talked about this when we defined the card, but just to be clear... do we want to allow the user to save their dataset when the URL test fails?

Yup. There are a variety of reasons why that might be, such as APIs not being ready yet or the file not uploaded. Now, it might be a reason to block starting the ingest when we get to that point.

apps/andi/lib/andi_web/live/edit_live_view.ex
LtChae: Yeah, Credo was upset about it so I went ahead and changed it.

apps/andi/assets/css/edit.scss
LtChae: They were used elsewhere so I continued the trend. Is there a problem using ids to identify styles for unique elements on a page?

I assume that BEM does not provide for using ids like this.

apps/andi/lib/andi/input_schemas/display_names.ex
LtChae: Changed

apps/andi/assets/css/edit.scss
NathanielScottStevens: If it's only ever going to be part of the form group I would do `edit-page__box-header`

apps/andi/lib/andi/input_schemas/dataset_input.ex
mcsearchin: Oh yeah, you did.  I apparently don't know how to read diffs.  Sorry!

apps/andi/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "0.14.4",
+      version: "0.15.0",
```
mcsearchin: You'll need to bump this again (sorry).  We also used version 0.15.0 for our change to bring in changes to smart_city.

apps/andi/lib/andi/services/url_test.ex
```diff
@@ -0,0 +1,25 @@
...
+  def test(url) do
+    case :timer.tc(&head/1, [url]) do
+      {time, {:ok, %{status: status}}} ->
+        timed_status(time, status)
+
+      {time, {:error, :nxdomain}} ->
+        timed_status(time, "Domain not found")
+
+      {time, _} ->
+        timed_status(time, "Could not complete request")
```
ManApart: If there is any possibility that timer can return anything else other than a tuple that starts with time, it'd be nice to have an even broader catch all. But if it ALWAYS returns {time, _}, than that works for me.

apps/andi/test/unit/andi/input_schemas/dataset_schema_validator_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule Andi.InputSchemas.DatasetSchemaValidatorTest do
       schema = [%{name: "field_name"}, %{name: "other_field", selector: "selector"}]
 
       errors = DatasetSchemaValidator.validate(schema, "application/json")
-      assert length(errors) == 0
+      assert Enum.empty?(errors)
```
ManApart: Good catch.

apps/andi/lib/andi/services/url_test.ex
LtChae: Yup, that's the way the erlang library works.

PR #472, opened 2020-01-23
UI to view events

sukaant:


apps/estuary/assets/static/robots.txt
```diff
@@ -0,0 +1,5 @@
+# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
```
dunson062786: What is this file for?

apps/estuary/assets/static/robots.txt
```diff
@@ -0,0 +1,5 @@
...
+#
+# To ban all spiders from the entire site uncomment the next two lines:
+# User-agent: *
```
dunson062786: Looks like you're supposed to uncomment these lines.

apps/estuary/config/config.exs
```diff
@@ -8,4 +8,15 @@ config :estuary,
   table_writer: Pipeline.Writer.TableWriter,
   connection: :estuary_elsa
 
+# Configures the endpoint
+config :estuary, EstuaryWeb.Endpoint,
+  url: [host: "localhost"],
+  # it should be overwriten as part of deploying the platform.
```
dunson062786: I don't know what secret_key_base is supposed to be, but it doesn't sound like it should be in our src code and on github.

apps/estuary/config/integration.exs
```diff
@@ -14,3 +14,34 @@ config :estuary,
...
+    patterns: [
+      ~r{priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$},
+      ~r{priv/gettext/.*(po)$},
+      ~r{lib/estuary_web/controllers/.*(ex)$},
+      ~r{lib/estuary_web/live/.*(ex)$},
+      ~r{lib/estuary_web/views/.*(ex)$},
+      ~r{lib/estuary_web/templates/.*(eex)$}
+    ]
+  ],
+  live_view: [
```
dunson062786: Is this being used? If so should it be an environment variable?

apps/estuary/assets/css/app.scss
```diff
@@ -0,0 +1,134 @@
+@import "./variables.scss";
```
bbrewer-papa-x: There is a lot of unused styling in here. At some point, we should probably share this styling somehow between this and ANDI. Until then, I'm fine having these extras here for consistency and future use.

apps/estuary/assets/static/robots.txt
bbrewer-papa-x: We don't, but that's a story for another day.

apps/estuary/lib/estuary_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,11 @@
+defmodule EstuaryWeb.HealthCheckController do
```
bbrewer-papa-x: I would consider swapping this out for `PlugHeartbeat` See https://github.com/smartcitiesdata/discovery_api/blob/a919fbdc16dbe1e6d8d1f5835f7ed9b88794c299/lib/discovery_api_web/endpoint.ex#L11

apps/estuary/config/prod.exs
```diff
@@ -1,3 +1,27 @@
...
+
 config :logger, level: :info
+
+config :estuary, EstuaryWeb.Endpoint,
+  http: [port: {:system, "PORT"}],
+  server: true,
+  root: ".",
+  cache_static_manifest: "priv/static/cache_manifest.json",
+  version: Application.spec(:estuary, :vsn),
+  live_view: [
```
smillardHNTB: i dont think you need this.

apps/estuary/test/unit/estuary/query/select_test.exs
```diff
@@ -0,0 +1,92 @@
...
+        "type" => "Type-2020-01-21 23:29:20.171543Z"
+      },
+      %{
+        "author" => "Author-2020-01-21 23:25:52.522084Z",
+        "create_ts" => 1_579_649_152,
+        "data" => "Data-2020-01-21 23:25:52.522107Z",
+        "type" => "Type-2020-01-21 23:25:52.522111Z"
+      }
+    ]
+
```
smillardHNTB: You might want a test here for when limit < # total items

smillardHNTB: Maybe also a test for order by ASC for a column other than create_ts

apps/estuary/lib/estuary/application.ex
```diff
@@ -4,10 +4,25 @@ defmodule Estuary.Application do
...
   end
+
+  # Tell Phoenix to update the endpoint configuration
+  # whenever the application is updated.
+  def config_change(changed, _new, removed) do
+    EstuaryWeb.Endpoint.config_change(changed, removed)
+    :ok
+  end
+
+  defp config_child(child) do
```
mcsearchin: Is there a better name for this function?  I think I can see what it's doing, but `config_child` doesn't seem quite right... maybe just `init`?

apps/estuary/lib/estuary_web/views/error_helpers.ex
```diff
@@ -0,0 +1,65 @@
...
+  https://elixirforum.com/t/liveview-phx-change-attribute-does-not-emit-event-on-input-text/21280
+  """
+  def error_tag_live(form, field) do
+    Enum.map(Keyword.get_values(form.errors, field), fn error ->
+      content_tag(:span, translate_error(error),
+        class: "error-msg",
+        id: "#{field}-error-msg"
+      )
+    end)
+  end
```
mcsearchin: You don't need this function.  We only had to add it to Andi once we started using the non text-input fields.  I don't think you need `error_tag` either unless you're doing form validation.

apps/estuary/lib/estuary_web/controllers/api/event_controller.ex
```diff
@@ -0,0 +1,29 @@
...
+        respond(conn, :not_found, "Unable to process your request")
+    end
+  end
+
+  defp respond(conn, status, body) do
+    conn
+    |> put_status(status)
+    |> json(body)
+  end
+end
```
mcsearchin: Verify with your PO whether this API is needed.  Seems like overkill for your card IMHO.

bbrewer-papa-x:
> I don't see any of the functions in your live view templates that actually make them update live (`handle_info` or `handle_event`).



> I don't see any of the functions in your live view templates that actually make them update live (`handle_info` or `handle_event`).

Those assume the page will live/auto refresh. If so, @sukaant see https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-example for an example of a timer driver update using `handle_info`

jeffgrunewald:
Alternatively, live view should be automatically configured to start a Phoenix.PubSub service that can broadcast messages, such as when events are added, deleted, or updated. With this volume and timer on the order of every 5 seconds might have the least negative impact on performance.

apps/estuary/config/integration.exs
```diff
@@ -14,3 +14,34 @@ config :estuary,
...
+    signing_salt: "SUPER VERY TOP SECRET!!!"
```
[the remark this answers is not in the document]
jeffgrunewald: Phoenix requires it under the hood. Environment variable isn’t important unless it’s the prod/runtime config.

PR #473, opened 2020-01-24
Added Benefit and Risk ratings to data curator UI

mcsearchin:


apps/andi/lib/andi/input_schemas/options.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Andi.InputSchemas.Options do
+  @moduledoc false
+
+  def ratings() do
+    %{
```
smillardHNTB: shouldn't these be defined in smartcitiesdata so they can be used everywhere as the same?

apps/andi/test/unit/andi/input_schemas/dataset_input_test.exs
```diff
@@ -165,6 +169,21 @@ defmodule Andi.InputSchemas.DatasetInputTest do
       assert changeset.errors
              |> Enum.any?(fn {:schema, {error, _}} -> String.match?(error, ~r/selector.+another_field/) end)
     end
+
+    data_test "requires #{field} to have an acceptable value" do
```
smillardHNTB: This test is doing the opposite - it's testing that they can't have an invalid value

apps/andi/test/unit/andi_web/live/edit_live_view_test.exs
```diff
@@ -222,86 +249,61 @@ defmodule AndiWeb.EditLiveViewTest do
...
-        :license,
-        "Please enter a valid license."
-      )
+    data_test "required #{field} field displays proper error message", %{conn: conn} do
+      html = save_form_for_dataset(conn, TDG.create_dataset(dataset_override))
 
+      assert get_text(html, "##{field}-error-msg") == expected_error_message
+
+      where([
+        [:field, :dataset_override, :expected_error_message],
```
smillardHNTB: should some of these be using the display names from within apps/andi/lib/andi/input_schemas/display_names.ex?

apps/reaper/config/test.exs
```diff
@@ -1,7 +1,7 @@
 use Mix.Config
 
 config :logger,
-  level: :info
+  level: :warn
```
smillardHNTB: Was this accidentally included?

apps/andi/test/unit/andi/input_schemas/dataset_input_test.exs
mcsearchin: Okay, we can fix the name of the test.

apps/andi/lib/andi/input_schemas/options.ex
mcsearchin: I don't know of a requirement to do so *yet*.  Right now it's Andi's job to validate incoming data, and I'd like to keep it that way.  These ratings values are currently not exposed externally, so only Andi needs to know about them.  

We boy-scouted these other options into this file just to make things consistent, but I think we need to have a larger discussion about fields like 'language' and what the allowed values should be.

apps/andi/test/unit/andi_web/live/edit_live_view_test.exs
mcsearchin: They are using those names.  We just don't want to reference that module in the test so changes to these messages still has to be intentional.

apps/reaper/config/test.exs
mcsearchin: Yes.  Unfortunately, reaper's test logging was very noisy and this bit of configuration was overriding the configs for Andi.  So even though Andi had warn level logging for unit tests, the alphabet made it so that Reapers info logging won every time.  😦 

Unit test logging is much quieter now.  😄

PR #488, opened 2020-01-29
make reaper discard dataset updates that raise

bbrewer-papa-x:
smartcitiesdata/smartcitiesdata#405

co-authored-by: Scott Millard <smillard@hntb.com>


LtChae:
Do you need to update the version in mix.exs?

PR #498, opened 2020-02-04
Update select query

sukaant:


apps/estuary/lib/estuary/services/event_retrieval_service.ex
```diff
@@ -4,18 +4,30 @@ defmodule Estuary.Services.EventRetrievalService do
...
 
   def get_all() do
-    make_select_table_schema()
-    |> Select.select_table()
+    make_select_table_sub_schema()
+    |> Select.create_select_statement()
+    |> make_select_table_schema
+    |> Select.create_select_statement()
+    |> PrestigeHelper.execute_query()
   end
```
mcsearchin: It's not clear to me what's happening here.  You're piping one create_select_statement into another... is there a way we can do this that makes it a little clearer what the final query is?

apps/estuary/test/unit/estuary/query/select_test.exs
```diff
@@ -2,126 +2,145 @@ defmodule Estuary.Query.SelectTest do
...
+  test "should return error, when table name is missing" do
+    expected_error = %RuntimeError{message: "Table name missing"}
+    table_schema = %{}
+    actual_error = Select.create_select_statement(table_schema)
+    assert expected_error == actual_error
+  end
+
+  @tag capture_log: true
+  test "should create the query with * with table name without any condition with limit 1000, when limit is passed" do
+    expected_select_statement = "SELECT *\n      FROM any_table\n      \n      \n      LIMIT 1000"
```
mcsearchin: I don't think we should expect line breaks and tabs in these generated statements.

apps/estuary/test/unit/estuary/services/event_retrieval_service_test.exs
```diff
@@ -3,26 +3,28 @@ defmodule Estuary.Services.EventRetrievalServiceTest do
...
+           "author" => "Author-2020-01-21 23:25:52.522084Z",
+           "create_ts" => 1_579_649_152,
+           "data" => "Data-2020-01-21 23:25:52.522107Z",
+           "type" => "Type-2020-01-21 23:25:52.522111Z"
+         }
+       ]}
 
-    allow(Select.select_table(any()), return: {:ok, expected_events})
+    allow(PrestigeHelper.execute_query(any()), return: expected_events)
     assert expected_events = EventRetrievalService.get_all()
```
mcsearchin: This test doesn't seem to add much value.  You're basically just testing that the value that you have mocked is returned.  Maybe this code would be covered better by an integration test?

apps/estuary/lib/estuary/services/event_retrieval_service.ex
```diff
@@ -2,21 +2,15 @@ defmodule Estuary.Services.EventRetrievalService do
...
-      "order" => "DESC",
-      "limit" => 1000
-    }
+  def get_all do
+    "SELECT events.author, events.create_ts, events.data, events.type
+    FROM (SELECT author, create_ts, data, type
+    FROM #{DatasetSchema.table_name()}
+    WHERE create_ts >= to_unixtime(now() - interval '3' hour) LIMIT 1000) events
+    ORDER BY events.create_ts DESC"
+    |> PrestigeHelper.execute_query()
```
mcsearchin: I think this might read better as a multiline string:

```
    query = """
            SELECT events.author, events.create_ts, events.data, events.type
            FROM (SELECT author, create_ts, data, type
            FROM #{DatasetSchema.table_name()}
            WHERE create_ts >= to_unixtime(now() - interval '3' hour) LIMIT 1000) events
            ORDER BY events.create_ts DESC
            """
    PrestigeHelper.execute_query(query)
```

apps/estuary/test/unit/estuary/services/event_retrieval_service_test.exs
```diff
@@ -1,28 +0,0 @@
...
-        "create_ts" => 1_579_649_152,
-        "data" => "Data-2020-01-21 23:25:52.522107Z",
-        "type" => "Type-2020-01-21 23:25:52.522111Z"
-      }
-    ]
-
-    allow(Select.select_table(any()), return: {:ok, expected_events})
-    assert expected_events = EventRetrievalService.get_all()
-  end
-end
```
mcsearchin: Possible to update this test to verify the query that is being executed (at least so we know it's getting built with the right table name)?

PR #512, opened 2020-02-07
remove registry from ANDI

bbrewer-papa-x:


bbrewer-papa-x:
> Will this allow us to repost/update orgs now?

Nope. Is there a card for that?!

PR #520, opened 2020-02-11
untested runtime

LtChae:
Due to the runtime.exs not being run during anything, this bug didn't get caught.

### DIFF
--- a/apps/discovery_api/runtime.exs
+++ b/apps/discovery_api/runtime.exs
@@ -1,12 +1,14 @@
 use Mix.Config
 
-redis_host = System.get_env("REDIS_HOST")
-redis_password = System.get_env("REDIS_PASSWORD")
-all_redis_args = [host: redis_host, password: redis_password]
-redix_args = Enum.filter(all_redis_args, fn
-  {_, nil} -> false
-  _ -> true
-end)
+get_redix_args = fn (host, password) ->
+  [host: host, password: password]
+  |> Enum.filter(fn
+    {_, nil} -> false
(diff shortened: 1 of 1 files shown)


apps/discovery_api/runtime.exs
```diff
@@ -1,10 +1,10 @@
 use Mix.Config
 
 redis_host = System.get_env("REDIS_HOST")
-redis_password = System.get_env("REDIS_PASSWORD", "")
+redis_password = System.get_env("REDIS_PASSWORD")
 all_redis_args = [host: redis_host, password: redis_password]
 redix_args = Enum.filter(all_redis_args, fn
-  {_, ""} -> false
+  {_, nil} -> false
```
mcsearchin: Should we handle both?  Or would it be safer to check for a string that has some length?

bbrewer-papa-x: I think this same pattern is all over the place too.

bbrewer-papa-x: Example from ANDI
```
get_redix_args = fn host, password ->
  [host: host, password: password]
  |> Enum.filter(fn
    {_, nil} -> false
    {_, ""} -> false
    _ -> true
  end)
end

redix_args = get_redix_args.(System.get_env("REDIS_HOST"), System.get_env("REDIS_PASSWORD"))
```

LtChae: I was apparently missing a fair bit of context on this one. Had no idea that this had already been done to the umbrella pre-discovery_api

PR #524, opened 2020-02-11
And params and headers

mcsearchin:


apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -13,7 +13,7 @@ defmodule Andi.InputSchemas.InputConverter do
     %{id: id, business: business, technical: technical} = AtomicMap.convert(dataset, safe: false, underscore: false)
```
mcsearchin: TODO: make sure we don't atomize maps with string keys here.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -216,11 +220,42 @@ defmodule AndiWeb.EditLiveView do
     {:noreply, assign(socket, test_results: results, testing: false)}
   end
 
+  def handle_info({:validate, %{"form_data" => form_data}}, socket), do: handle_validation(form_data, socket)
+
+  def handle_info({:add_key_value, %{"field" => field}}, socket) do
```
LtChae: It would be nice if it were more clear where this message is coming from. We figured it out, as it was coming from the Key Value Editor

apps/andi/lib/andi_web/views/error_helpers.ex
```diff
@@ -69,7 +59,13 @@ defmodule AndiWeb.ErrorHelpers do
 
   defp interpret_error(error, field) do
     {_, opts} = error
-    updated_message = "Please enter a valid #{get_downcased_display_name(field)}."
+
+    updated_message =
+      case field do
+        field when field in [:sourceHeaders, :sourceQueryParams] -> "Please enter valid key(s)."
```
LtChae: It would be nice if these generic error helpers weren't coupled to fields that are very specific to the dataset editor, but it's not the end of the wordl < Ben

apps/andi/lib/andi_web/live/edit_live_view.ex
mcsearchin: The "add_key_value" message is sent from the event handler in the KeyValueEditor... this matching string should make it relatively clear how these are linked.  Will continue to think about how to make this better.

apps/andi/lib/andi_web/views/error_helpers.ex
mcsearchin: I agree.  My thinking was that the next card with custom error messaging would be the one to have to move this into a more sophisticated 'lookup' function.

PR #526, opened 2020-02-12
Delete dataset

sukaant:


apps/valkyrie/lib/valkyrie/dataset_processor.ex
```diff
@@ -18,4 +18,10 @@ defmodule Valkyrie.DatasetProcessor do
   def stop(dataset_id) do
     Valkyrie.DatasetSupervisor.ensure_stopped(dataset_id)
   end
+
+  def delete(dataset_id) do
+    Logger.debug("#{__MODULE__}: Deleting Datatset: #{dataset_id}")
+    Valkyrie.DatasetSupervisor.ensure_stopped(dataset_id)
```
kcarra: This would be the same as just calling stop()

apps/valkyrie/test/integration/dataset_mutation_test.exs
```diff
@@ -58,4 +60,27 @@ defmodule Valkyrie.DatasetMutationTest do
...
   end
+
+  test "should delete all view state for the dataset and the input and output topics when dataset:delete is called" do
+    dataset_id = Faker.UUID.v4()
+    input_topic = "#{@input_topic_prefix}-#{dataset_id}"
+    output_topic = "#{@output_topic_prefix}-#{dataset_id}"
+    dataset = TDG.create_dataset(id: dataset_id, technical: %{sourceType: "ingest"})
+    Brook.Event.send(@instance, data_ingest_start(), :author, dataset)
+    Process.sleep(2_000)
+    topic_list_before_delete = Elsa.list_topics(@endpoints)
```
LtChae: You could make this eventually if you don't like the sleep. I'm honestly not sure why there were sleeps already

PR #529, opened 2020-02-13
Direct json

LtChae:


apps/e2e/test/e2e_test.exs
```diff
@@ -354,7 +366,7 @@ defmodule E2ETest do
 
           [coordinates] = result["geometry"]["coordinates"]
 
-          assert 253 == Enum.count(coordinates)
```
LtChae: This was very flaky before and was probably only as flaky as before due to our change. Making it less flaky with this < Ben

apps/pipeline/lib/pipeline/writer/s3_writer.ex
```diff
@@ -0,0 +1,144 @@
...
+  @doc """
+  Ensures PrestoDB tables exist for JSON and ORC formats.
+  """
+  def init(options) do
+    json_config = json_config(options)
+    orc_config = orc_config(options)
+
+    with {:ok, statement} <- Statement.create(json_config),
+         {:ok, _} <- execute(statement),
+         {:ok, statement} <- Statement.create(orc_config),
```
JarredOlson: Please name these `json_statement` and `orc_statement` or `statement_<type>` ?

apps/pipeline/lib/pipeline/writer/s3_writer.ex
```diff
@@ -0,0 +1,144 @@
...
+    }
+  end
+
+  defp json_config(options) do
+    %{
+      table: json_table_name(options),
+      schema: Keyword.fetch!(options, :schema),
+      format: "JSON"
+    }
+  end
```
JarredOlson: These could be unified, by passing in format for the "ORC" version as well.  I know that ORC is the default, but we reduce the risk of these diverging in the future.

apps/e2e/test/e2e_test.exs
mcsearchin: 👍

apps/pipeline/lib/pipeline/writer/s3_writer/compaction.ex
```diff
@@ -0,0 +1,124 @@
...
+         {:ok, orig_results} <- Task.await(count_task, :infinity),
+         _ <- Task.await(compaction_task, :infinity),
+         {:ok, new_results} <- execute("select count(1) from #{orc_table}_compact") do
+      [[new_row_count]] = new_results.rows
+      [[old_row_count]] = orig_results.rows
+      {new_row_count, old_row_count}
+    end
+  end
+
+  def complete({new, old}, options) when new == old do
```
mcsearchin: `{new_count, old_count}`?

mcsearchin:
Perhaps not necessary for this card, but should we consider adding a behavior for the Compaction module?  Does the S3 compaction module have enough in common with the ones for DataWriter or TableWriter to justify this?

PR #534, opened 2020-02-17
Delete dataset

sukaant:


apps/discovery_streams/lib/discovery-streams/event_handler.ex
```diff
@@ -31,6 +31,25 @@ defmodule DiscoveryStreams.EventHandler do
...
+      :ok ->
+        Logger.debug("#{__MODULE__}: Deleted dataset for dataset id: #{id} and system name: #{system_name}")
+
+      {:error, error} ->
+        Logger.error(
+          "#{__MODULE__}: Failed to delete dataset for dataset id: #{id} and system name: #{system_name}, Reason: #{
+            inspect(error)
+          }"
+        )
+    end
```
mcsearchin: Consider moving this whole block into its own block and calling it `delete_dataset_topic`?

apps/discovery_streams/lib/discovery-streams/topic_helper.ex
```diff
@@ -10,7 +10,19 @@ defmodule DiscoveryStreams.TopicHelper do
...
 
+  def get_endpoints() do
+    Application.get_env(:kaffe, :consumer)[:endpoints]
+    |> Enum.map(fn {host, port} -> {to_charlist(host), port} end)
+  end
+
+  def delete_topic(dataset_id) do
+    output_topic = output_topic(dataset_id)
+    Elsa.delete_topic(get_endpoints(), output_topic)
+  end
```
mcsearchin: This function should probably be called `delete_output_topic` since it is using the output_topic prefix.

PR #535, opened 2020-02-18
Data json updates

ScottMillard:


LtChae:
Looks good, but don't forget to bump the version
^ Ben

apps/discovery_api/test/unit/discovery_api_web/controllers/multiple_metadata_controller_datajson_test.exs
```diff
@@ -54,6 +54,24 @@ defmodule DiscoveryApiWeb.MultipleMetadataController.DataJsonTest do
       {:ok, %{model: public_model, results: results}}
     end
 
+    test "language is a list", %{results: [result | _]} do
+      assert is_list(result["language"]) == true
+    end
```
mcsearchin: Can we add a test to prove that when language is nil that it is not included in the list?

ScottMillard: That's covered by another test that was already writter

PR #536, opened 2020-02-18
make compaction not duplicate data

LtChae:
smartcitiesdata/smartcitiesdata#507

### DIFF
--- a/apps/forklift/lib/forklift/message_handler.ex
+++ b/apps/forklift/lib/forklift/message_handler.ex
@@ -28,7 +28,10 @@ defmodule Forklift.MessageHandler do
       |> Enum.reject(&error_tuple?/1)
       |> Forklift.DataWriter.write(dataset: dataset)
 
-    Task.start(fn -> Forklift.DataWriter.write_to_topic(timed_messages) end)
+    Task.start(fn ->
+      result = Forklift.DataWriter.write_to_topic(timed_messages)
+      Logger.debug(fn -> "Finished writing #{Enum.count(timed_messages)} timed messages for dataset #{dataset.id} with result #{inspect(result)}" end)
+    end)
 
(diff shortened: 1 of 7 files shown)


apps/forklift/lib/forklift/message_handler.ex
```diff
@@ -21,11 +21,14 @@ defmodule Forklift.MessageHandler do
...
-    |> Enum.reject(&error_tuple?/1)
-    |> Forklift.DataWriter.write(dataset: dataset)
+    timed_messages =
+      messages
+      |> Enum.map(&parse/1)
+      |> Enum.map(&yeet_error/1)
+      |> Enum.reject(&error_tuple?/1)
+      |> Forklift.DataWriter.write(dataset: dataset)
+
+    Task.start(fn -> Forklift.DataWriter.write_to_topic(timed_messages) end)
```
mcsearchin: Is there anything we should be doing once this task is complete? Even just debug logging the end result?

mcsearchin:
It's not clear from looking at the diff... are there any tests here that verify that we don't duplicate data during compaction?  I'm guessing this is the kind of test that wouldn't have a strong JSR - just curious.

bbrewer-papa-x:
> It's not clear from looking at the diff... are there any tests here that verify that we don't duplicate data during compaction? I'm guessing this is the kind of test that wouldn't have a strong JSR - just curious.

We considered putting a count check into the geojson part of the e2e test. I can give that a shot and see if it's stable. That test has always been a bit flaky though

apps/forklift/lib/forklift/message_handler.ex
bbrewer-papa-x: We're putting a debugge message in there

PR #542, opened 2020-02-20
Updates to andi regex to conform to podms

kcarra:
smartcitiesdata/migration#506

co-authored-by: Scott Millard <wannafly37@gmail.com>

### DIFF
--- a/apps/andi/lib/andi/input_schemas/dataset_input.ex
+++ b/apps/andi/lib/andi/input_schemas/dataset_input.ex
@@ -68,7 +68,7 @@ defmodule Andi.InputSchemas.DatasetInput do
     :sourceUrl
   ]
 
-  @email_regex ~r/^[A-Za-z0-9._%+-+']+@[A-Za-z0-9.-]+\.[A-Za-z]+$/
+  @email_regex ~r/^[\w\_\~\!\$\&\'\(\)\*\+\,\;\=\:.-]+@[\w.-]+\.[\w.-]+?$/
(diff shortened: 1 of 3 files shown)


apps/andi/lib/andi/input_schemas/dataset_input.ex
```diff
@@ -68,7 +68,7 @@ defmodule Andi.InputSchemas.DatasetInput do
     :sourceUrl
   ]
 
-  @email_regex ~r/^[A-Za-z0-9._%+-+']+@[A-Za-z0-9.-]+\.[A-Za-z]+$/
+  @email_regex ~r/^[\w\_\~\!\$\&\'\(\)\*\+\,\;\=\:.-]+@[\w.-]+\.[\w.-]+?$/
```
mcsearchin: JMS 😝

apps/andi/test/unit/andi_web/live/edit_live_view_test.exs
```diff
@@ -133,7 +133,8 @@ defmodule AndiWeb.EditLiveViewTest do
         [:email],
         ["foo@mail.com"],
         ["kevin@space.org"],
-        ["my@little.gov"]
+        ["my@little.gov"],
+        ["test-email@email.com"]
```
mcsearchin: I take it hyphens didn't match the existing regex?

kcarra: Correct, the regex we replaced it with is the regex that PODM's uses to validate emails

PR #543, opened 2020-02-20
Delete Dataset

sukaant:


apps/reaper/lib/reaper/event/handlers/dataset_delete.ex
```diff
@@ -0,0 +1,19 @@
...
+
+  require Logger
+
+  alias Reaper.Event.Handlers.Helper.DatasetHelper
+  alias Reaper.Topic.TopicManager
+
+  def handle(%SmartCity.Dataset{id: dataset_id}) do
+    Logger.debug("#{__MODULE__}: Deleting Datatset: #{dataset_id}")
+
+    with :ok <- DatasetHelper.deactivate_quantum_job(dataset_id),
```
kcarra: Would it be possible to have the dataset disabler do everything but the topic delete, so we don't have duplicate code? - Ben

Example:
```elixir
def handle(%SmartCity.Dataset{id: dataset_id} = dataset) do
...
with :ok <- Reaper.Event.Handlers.DatasetDisable.handle(dataset),
        :ok <- TopicManager.delete_topic(dataset_id) do
...
```

kcarra:
Might be worth throwing in an integration test to make sure you are actually deleting the dataset from view state and kafka

PR #547, opened 2020-02-25
Data json service

kcarra:


apps/discovery_api/test/integration/discovery_api/data/data_json_test.exs
```diff
@@ -8,13 +8,15 @@ defmodule DiscoveryApi.Data.DataJsonTest do
...
 
-  setup do
+  import SmartCity.TestHelper, only: [eventually: 1, eventually: 3]
+
+  setup_all do
     Helper.wait_for_brook_to_be_ready()
     Redix.command!(:redix, ["FLUSHALL"])
-    :ok
-  end
+    # on_exit(fn ->
```
LtChae: Should this still be here?

apps/discovery_api/test/integration/discovery_api/data/data_json_test.exs
```diff
@@ -45,22 +49,25 @@ defmodule DiscoveryApi.Data.DataJsonTest do
...
-    else
-      false
-    end
+    eventually(fn ->
+      assert Enum.count(get_data_json_datasets()) == 3
+    end)
+  end
+
+  defp get_data_json_datasets() do
+    "http://localhost:4000/api/v1/data_json"
```
LtChae: Should this get the port from the config?

apps/discovery_api/test/integration/discovery_api/data/data_json_test.exs
```diff
@@ -45,22 +49,25 @@ defmodule DiscoveryApi.Data.DataJsonTest do
...
+    |> get_map_from_url()
+    |> Map.get("dataset", [])
   end
 
   defp get_map_from_url(url) do
     url
-    |> HTTPoison.get!()
+    |> HTTPoison.get!([], follow_redirect: true)
+    |> IO.inspect(label: "response")
```
LtChae: Credo

apps/discovery_api/config/config.exs
```diff
@@ -45,4 +45,11 @@ config :mime, :types, %{
 config :discovery_api,
   user_visualization_limit: 1_000
 
+
+# data_json_service options are :local or s3
+  # :local sends the local data.json
+  # :s3 redirects to the data.json file hosted in s3
```
LtChae: Where do we upload the data.json to s3?

apps/discovery_api/lib/discovery_api/event_handler.ex
```diff
@@ -46,12 +47,15 @@ defmodule DiscoveryApi.EventHandler do
...
          {:ok, _cached} <- SystemNameCache.put(dataset.id, organization.name, dataset.technical.dataName),
          model <- Mapper.to_data_model(dataset, organization) do
+
       DiscoveryApi.Search.Storage.index(model)
       save_dataset_to_recommendation_engine(dataset)
       ResponseCache.invalidate()
       Logger.debug(fn -> "Successfully handled message: `#{dataset.technical.systemName}`" end)
-
       merge(:models, model.id, model)
+      DataJson.delete_data_json()
```
LtChae: Why do this after the update to view state?  Shouldn't the merge be last?  (Tony)

apps/discovery_api/lib/discovery_api_web/controllers/data_json_controller.ex
```diff
@@ -0,0 +1,9 @@
+defmodule DiscoveryApiWeb.DataJsonController do
+  use DiscoveryApiWeb, :controller
+
+  plug DiscoveryApiWeb.Plugs.DataJson
+
+  def show(conn, _params) do
```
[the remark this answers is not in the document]
kcarra: We merged the plug into here

apps/discovery_api/test/integration/discovery_api/data/data_json_test.exs
kcarra: We probably should, though this is pretty much everywhere at the moment and we're not in a good spot to iterate on our integration tests :( (we have all the prod datasets loaded locally...) - Ben

apps/discovery_api/config/config.exs
kcarra: We don't do this anymore :) - Ben

apps/discovery_api/lib/discovery_api/event_handler.ex
kcarra: This is to deal with a possible race condition where:
- data.json is deleted
- someone requests it via API and it gets rebuilt (without the new/updated model from the merge)
- new/updated model is merged, but not included in the now cached data.json file

Merging first ensures that the model is available for any rebuilds of the data.json file that might occur after the delete.

- Ben and Kevin

PR #561, opened 2020-03-05
update login link to reflect when someone successfully logs in

LtChae:

### DIFF
--- a/apps/discovery_api/priv/static/tableau/connector.js
+++ b/apps/discovery_api/priv/static/tableau/connector.js
@@ -79,6 +79,11 @@ function _setupConnector() {
       .then(function(refreshToken) { tableau.password = refreshToken; })
       .catch(function(error) { _displayLoginError(error) })
       .then(function() { initCallback() })
+      .then(function() {
+        document.getElementById('login-text').textContent = 'Logged In';
+        document.getElementById('login-button').onclick = undefined;
+        document.getElementById('login-button').className = 'login-link';
+       })
     } else {
       initCallback();
     }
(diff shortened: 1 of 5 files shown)


apps/discovery_api/priv/static/tableau/connector.js
```diff
@@ -79,6 +79,11 @@ function _setupConnector() {
       .then(function(refreshToken) { tableau.password = refreshToken; })
       .catch(function(error) { _displayLoginError(error) })
       .then(function() { initCallback() })
+      .then(function() {
+        document.getElementById('login-text').textContent = 'Logged In';
+        document.getElementById('login-button').onclick = undefined;
+        document.getElementById('login-button').className = 'login-link';
+       })
     } else {
```
dunson062786: what about $('#login-text'), $('#login-button')?

apps/discovery_api/priv/static/tableau/connector.js
```diff
@@ -79,6 +79,11 @@ function _setupConnector() {
       .then(function(refreshToken) { tableau.password = refreshToken; })
       .catch(function(error) { _displayLoginError(error) })
       .then(function() { initCallback() })
+      .then(function() {
+        document.getElementById('login-text').textContent = 'Logged In';
+        document.getElementById('login-button').onclick = undefined;
+        document.getElementById('login-button').className = 'login-link';
+       })
```
mcsearchin: Should this go before the catch block?  
Also, how's about a test for this behavior?  😁

apps/discovery_api/priv/static/tableau/connector.js
LtChae: We don't have jquery here.

PR #563, opened 2020-03-06
Adding dataset delete

sukaant:


apps/andi/lib/andi/dataset_util.ex
```diff
@@ -0,0 +1,13 @@
...
+  """
+
+  import Andi
+
+  @collection :dataset
+
+  def delete(id) do
+    Brook.ViewState.delete(@collection, id)
+  end
+end
```
mcsearchin: Can we put this under Andi.Services?  Maybe consolidate this with the DatasetRetrieval?

mcsearchin: After discussion - we'll address this in a separate card.

PR #565, opened 2020-03-06
Filter out remote datasets for data_json

kcarra:


apps/discovery_api/test/integration/discovery_api/data/data_json_test.exs
```diff
@@ -64,6 +64,15 @@ defmodule DiscoveryApi.Data.DataJsonTest do
...
 
+  test "Remote Datasets are idempotent", %{organization_id: organization_id} do
+    remote_dataset = TDG.create_dataset(%{technical: %{orgId: organization_id, sourceType: "remote"}, business: %{dataTitle: "Im remote fools"}})
+    Brook.Event.send(DiscoveryApi.instance(), dataset_update(), "integration", remote_dataset)
+
+    eventually(fn ->
+      assert Enum.count(get_data_json_datasets()) == 3
+    end)
+  end
+
```
mcsearchin: Could this test possibly pass due to it being asynchronous? Thinking it might make sense for us to also send through a non remote dataset and see that the count only goes up by one.

PR #568, opened 2020-03-09
Refactor for Brook

sukaant:


apps/andi/test/unit/andi/migration/migrations_test.exs
```diff
@@ -10,9 +10,10 @@ defmodule Andi.Migration.MigrationsTest do
   @modified_date_event "migration:modified_date:start"
 
   alias Andi.Migration.Migrations
+  alias Andi.Services.DatasetRetrieval
 
   test "send the modified date migration event if it has not succeeded yet" do
-    allow(Brook.get!(@instance, :migration, "modified_date_migration_completed"), return: nil)
+    allow(DatasetRetrieval.get!("modified_date_migration_completed", :migration), return: nil)
```
mcsearchin: This doesn't seem like it should have changed.  Seems like we should only be wrapping calls to brook that look at the `:dataset` or `:ingested_time` collections.

apps/andi/lib/andi/services/dataset_retrieval_service.ex
```diff
@@ -14,4 +16,24 @@ defmodule Andi.Services.DatasetRetrieval do
...
+    Brook.get_all_values!(instance_name(), @collection)
+  end
+
+  def get_all_ingested_time!() do
+    Brook.get_all_values!(instance_name(), :ingested_time)
+  end
+
+  def delete(id) do
+    Brook.ViewState.delete(@collection, id)
+  end
```
mcsearchin: Why do these functions on Brook.ViewState NOT take in an instance of Brook?  I'd like to understand how this works to make sure we're calling these correctly.

mcsearchin: Also, can we make all of these function signatures consistent?  Either have ALL of them take in the optional instance of Brook, or have NONE of them take in this arg.

apps/andi/test/unit/andi/dataset_store_test.exs
```diff
@@ -0,0 +1,44 @@
...
+
+  @brook_instance :andi
+
+  describe "get_all/0" do
+    test "retrieves all events from Brook" do
+      dataset1 = TDG.create_dataset(%{})
+      dataset2 = TDG.create_dataset(%{})
+      expected_datasets = {:ok, [dataset1, dataset2]}
+      Brook.Event.send(@brook_instance, dataset_update(), :dataset, dataset1)
+      Brook.Event.send(@brook_instance, dataset_update(), :dataset, dataset2)
```
mcsearchin: I don't think you need these calls to Brook.Event.send anymore (since you're mocking out get_all_values).

apps/andi/lib/andi/dataset_store.ex
```diff
@@ -7,11 +7,20 @@ defmodule Andi.DatasetStore do
...
     Brook.ViewState.merge(@collection_dataset, dataset.id, dataset)
   end
 
+  def get(id) do
+    Brook.get(instance_name(), @collection_dataset, id)
+  end
+
+  def get_org(id) do
+    Brook.get(instance_name(), @collection_org, id)
+  end
```
mcsearchin: These functions for org feel like they should be in a separate module.

PR #573, opened 2020-03-11
Return allowed actions with visualization responses

cameronmarsh:


apps/discovery_api/lib/discovery_api_web/controllers/visualization_controller.ex
```diff
@@ -5,20 +5,23 @@ defmodule DiscoveryApiWeb.VisualizationController do
...
          visualizations <- Visualizations.get_visualizations_by_owner_id(user.id) do
-      render(conn, :visualizations, %{visualizations: visualizations})
+      render(conn, :visualizations, %{visualizations: visualizations, allowed_actions: @owner_allowed_actions})
     end
   end
 
   def show(conn, %{"id" => id}) do
     with {:ok, %{query: query} = visualization} <- Visualizations.get_visualization_by_id(id),
-         user <- Map.get(conn.assigns, :current_user),
+         user = Map.get(conn.assigns, :current_user),
```
kcarra: Perfect usage of this - Ben

apps/discovery_api/lib/discovery_api_web/controllers/visualization_controller.ex
```diff
@@ -59,4 +65,13 @@ defmodule DiscoveryApiWeb.VisualizationController do
   defp under_visualizations_limit?(visualizations) do
     Enum.count(visualizations) <= Application.get_env(:discovery_api, :user_visualization_limit)
   end
+
+  defp get_allowed_actions(_visualization, nil), do: []
```
bbrewer-papa-x: Is there any more appropriate layer for this kind of thing than here? I think it's probably fine, and I'm not sure if it really belongs in the BL layer either.

apps/discovery_api/lib/discovery_api_web/controllers/visualization_controller.ex
bbrewer-papa-x: Perfect usage of this

apps/discovery_api/test/unit/discovery_api_web/controllers/visualization_controller_test.exs
```diff
@@ -121,7 +114,8 @@ defmodule DiscoveryApiWeb.VisualizationControllerTest do
                "query" => @query,
                "title" => @title,
                "id" => @id,
-               "chart" => @decoded_chart
+               "chart" => @decoded_chart,
+               "allowed_actions" => [%{"name" => "update"}, %{"name" => "create_copy"}]
```
bbrewer-papa-x: Thank you for choosing this form for the return. We've chosen things like arrays of strings and maps before and it has locked us in pretty hard.

apps/discovery_api/lib/discovery_api_web/controllers/visualization_controller.ex
mcsearchin: Good call out.  I'm thinking we wait and see how much more complicated this gets.  We could move it into its own plug or a helper module, but probably wouldn't do so until we have more rules.

apps/discovery_api/test/unit/discovery_api_web/controllers/visualization_controller_test.exs
mcsearchin: 👍 most of the credit to @JarredOlson

PR #576, opened 2020-03-13
Delete dataset

sukaant:


apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
```diff
@@ -65,4 +68,20 @@ defmodule DiscoveryApi.EventHandlerTest do
...
+      allow(StatsCalculator.delete_completeness(any()), return: :whatever_seriously)
+      allow(ResponseCache.invalidate(), return: :seriously_do_not_care)
+      allow(SystemNameCache.delete(any(), any()), return: :its_very_serious)
+      allow(DataJsonService.delete_data_json(), return: :ok)
+
+      dataset = TDG.create_dataset(%{})
+
+      Brook.Event.process(:discovery_api, Brook.Event.new(type: dataset_delete(), data: dataset, author: :author))
+
+      assert_called(DataJsonService.delete_data_json())
```
kcarra: It would be nice to have some more assertions in here, as well as a test for the failure case (for the rescue you have) for this and then also an integration test.

- Ben and Kevin

apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
```diff
@@ -65,4 +68,20 @@ defmodule DiscoveryApi.EventHandlerTest do
       assert_called(DataJsonService.delete_data_json())
     end
   end
+
+  describe "handle_event/1 #{dataset_delete()}" do
```
LtChae: Is there value in testing the individual parts of this process? Should we at least `expect` these calls with the appropriate parameters?

apps/discovery_api/lib/discovery_api/event_handler.ex
```diff
@@ -62,6 +66,19 @@ defmodule DiscoveryApi.EventHandler do
...
 
+  def handle_event(%Brook.Event{type: dataset_delete(), data: %Dataset{} = dataset}) do
+    DiscoveryApi.RecommendationEngine.delete(dataset.id)
+    SystemNameCache.delete(dataset.technical.orgName, dataset.technical.dataName)
+    DiscoveryApi.Search.Storage.delete(dataset)
+    StatsCalculator.delete_completeness(dataset.id)
+    ResponseCache.invalidate()
+    DataJsonService.delete_data_json()
+    Logger.debug("#{__MODULE__}: Deleted dataset for dataset: #{dataset.id}")
+  rescue
```
mcsearchin: Where is the actual call to `Model.delete`?

apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
mcsearchin: 👍

sukaant:
@JarredOlson Shall we add integration test? Please share your thoughts

apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
sukaant: @JarredOlson Please share your thoughts

apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
```diff
@@ -65,4 +68,30 @@ defmodule DiscoveryApi.EventHandlerTest do
...
+    end
+
+    test "should return ok if it throws error when dataset:delete is called", %{dataset: dataset} do
+      expect(DiscoveryApi.RecommendationEngine.delete(dataset.id), return: {:error, "ERR value is not an integer or out of range"})
+      expect(StatsCalculator.delete_completeness(dataset.id), return: {:error, "ERR value is not an integer or out of range"})
+      expect(ResponseCache.invalidate(), return: {:ok, true})
+      expect(SystemNameCache.delete(dataset.technical.orgName, dataset.technical.dataName), return: {:ok, true})
+      expect(Model.delete(dataset.id), return: {:error, "ERR value is not an integer or out of range"})
+      Brook.Event.process(:discovery_api, Brook.Event.new(type: dataset_delete(), data: dataset, author: :author))
+      assert {:ok, []} == DataJsonService.delete_data_json()
```
mcsearchin: At least one of these above should actually mock out raising an error.  Otherwise you're not really testing the rescue.  Try something like this:

```
allow(Model.delete(dataset.id), exec: fn _ -> raise {:error, "error"} end)
```

apps/discovery_api/lib/discovery_api/event_handler.ex
```diff
@@ -62,6 +66,20 @@ defmodule DiscoveryApi.EventHandler do
...
+    DiscoveryApi.Search.Storage.delete(dataset)
+    StatsCalculator.delete_completeness(dataset.id)
+    Model.delete(dataset.id)
+    ResponseCache.invalidate()
+    DataJsonService.delete_data_json()
+    Logger.debug("#{__MODULE__}: Deleted dataset for dataset: #{dataset.id}")
+  rescue
+    error ->
+      Logger.error("#{__MODULE__}: Failed to delete dataset for dataset: #{dataset.id}, Reason: #{inspect(error)}")
+  end
```
mcsearchin: Seems like both the happy and sad path should return `:discard`.  Right now they're both returning the result of the Logger function.

apps/discovery_api/test/unit/discovery_api/event_handler_test.exs
```diff
@@ -65,4 +68,30 @@ defmodule DiscoveryApi.EventHandlerTest do
...
+      %{dataset: TDG.create_dataset(%{id: Faker.UUID.v4()})}
+    end
+
+    test "should delete the dataset and return ok when dataset:delete is called", %{dataset: dataset} do
+      expect(DiscoveryApi.RecommendationEngine.delete(dataset.id), return: :ok)
+      expect(StatsCalculator.delete_completeness(dataset.id), return: :ok)
+      expect(ResponseCache.invalidate(), return: {:ok, true})
+      expect(SystemNameCache.delete(dataset.technical.orgName, dataset.technical.dataName), return: {:ok, true})
+      expect(Model.delete(dataset.id), return: :ok)
+      Brook.Event.process(:discovery_api, Brook.Event.new(type: dataset_delete(), data: dataset, author: :author))
```
mcsearchin: There are not assertions that you are calling `DiscoveryApi.Search.Storage.delete` or `DataJsonService.delete_data_json`.  I think this would be clearer if you used `assert_called` for each of the function calls you expect:

```
assert_called DiscoveryApi.RecommendationEngine.delete(dataset.id)
assert_called StatsCalculator.delete_completeness(dataset.id)
assert_called ResponseCache.invalidate()
# ... etc.
```

PR #581, opened 2020-03-13
place tighter controls around how a dataset is handled so we avoid re-ingesting old datasets

kcarra:
- extractions and ingestions are not triggered if a dataset has been explicitly disabled
- extractions and ingestions ONLY update the start timestamp for a dataset
- dataset update ONLY updates the dataset definition in view state
- datasets, if their definition is present in view state, are considered enabled until they are explicitly disabled

smartcitiesdata/smartcitiesdata#447

### DIFF
--- a/apps/reaper/test/integration/reaper/reaper_full_test.exs
+++ b/apps/reaper/test/integration/reaper/reaper_full_test.exs
@@ -571,6 +571,45 @@ defmodule Reaper.FullTest do
     ])
   end
 
+  data_test "dataset:disable followed by a ingest or extract end or file ingest end", %{bypass: bypass} do
+    dataset =
+      TDG.create_dataset(%{
+        technical: %{
+          cadence: "once",
+          sourceUrl: "http://localhost:#{bypass.port}/#{@csv_file_name}",
+          sourceFormat: "csv",
+          sourceType: source_type,
+          schema: [%{name: "id"}, %{name: "name"}, %{name: "pet"}]
+        }
+      })
+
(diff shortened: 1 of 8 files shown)


apps/reaper/test/integration/reaper/reaper_full_test.exs
```diff
@@ -419,6 +431,194 @@ defmodule Reaper.FullTest do
...
+      })
+
+    Brook.Event.send(@instance, dataset_update(), :reaper, dataset)
+
+    eventually(fn ->
+      assert Reaper.Collections.Extractions.get_dataset!(dataset.id) == dataset
+    end)
+  end
+
+  data_test "extracts and ingests update started_timestamp in view state", %{bypass: bypass} do
```
mcsearchin: Might be nice to name these data_tests so they are reflective of the parameters being used (using `start_event_type ` or `source_type` for instance).

kcarra: Having issues running the integration tests from home but i remember this was an issue with `data_test` not being able to interpolate functions.

bbrewer-papa-x: Yeah, Balser himself noted the data_test caveat here - the output of functions called in the where clause (including macros, somehow) cannot be used in the test name. We could just use the string versions in the where clause and hope it doesn't bite us when/if we update the event type macros. I'm 90% that we won't and that this will be the least of our problems if we do.

PR #583, opened 2020-03-16
Defaults and providers

LtChae:
This has three layers:
+ Providers
  + These are behavior driven structures within a map that can be executed to generate a value for a field at runtime.
  + These are found in the Providers "app". There is currently only one: `Timestamp`
  + They are executed via the Provisioner module
+ The new schema `default` field
  + This field is used as the value of a column when a record is missing that column or it is nil.
+ The Reaper Processor changes
  + Before Processing a dataset:extract event, Reaper will run the Providers in the dataset to populate any fields it needs. At the moment, that's only going to be `default` fields.

Notes:
* Providers are not run before ingest yet, so they cannot be used to set query params or retrieve secrets for the moment.
* Version is provided for future use as its expected to be annoying to do retroactively.


LtChae:
Working on cleaning up some of the boilerplate

apps/providers/lib/provider.ex
```diff
@@ -0,0 +1,48 @@
...
+        provider: "Echo",
+        opts: %{value: "Assistant Sales Rep"},
+        version: "1"
+      },
+      department: %{
+        name: "Sales",
+        type: %{
+          provider: "Echo",
+          opts: %{value: "Supremely Important"},
+          version: "1"
```
bbrewer-papa-x: I like this way of doing the versioning

apps/providers/lib/provider.ex
```diff
@@ -0,0 +1,48 @@
...
+  + Opts
+    + A map used to pass parameters to the provider. i.e. `%{value: "Supremely Important"}`.
+    + This field is __optional__. Some providers will not require input, such as `Providers.Timestamp`
+  + Version
+    + Allows for the versioning of providers. A major change to the functionality of a provider should be added to a new function head under a new version.
+    + This should avoid the need for migrations.
+
+  Providers provides the `Providers.Helpers.Provisioner` module for provisioning (executing within a map) providers.
+  """
+  @callback provide(String.t(), map()) :: any()
```
bbrewer-papa-x: Seems clear enough

apps/providers/lib/providers/timestamp.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Providers.Timestamp do
+  @moduledoc """
+  This provider implementation returns the current timestamp in UTC as an ISO8601 compliant date string.
+  """
+  @behaviour Providers.Provider
+  def provide("1", _opts) do
```
bbrewer-papa-x: I like it

apps/providers/test/helpers/provisioner_test.exs
```diff
@@ -0,0 +1,195 @@
...
+          version: "1"
+        },
+        rank: %{
+          provider: "Echo",
+          opts: %{value: "Assistant to the Manager"},
+          version: "1"
+        }
+      }
+
+      provisioned_map = Providers.Helpers.Provisioner.provision(map)
```
bbrewer-papa-x: I'm very glad we're mostly able to test this with inputs and outputs

apps/reaper/lib/reaper/data_extract/schema_filler.ex
```diff
@@ -32,6 +32,13 @@ defmodule Reaper.DataExtract.SchemaFiller do
     end
   end
 
+  defp fill_missing_fields(%{name: name, default: default}, payload) do
```
bbrewer-papa-x: noice

PR #597, opened 2020-03-19
Parse Jaxon errors for selecting empty arrays

kcarra:
smartcitiesdata/reaper#511

### DIFF
--- a/apps/reaper/lib/reaper/decoder/json.ex
+++ b/apps/reaper/lib/reaper/decoder/json.ex
@@ -47,10 +47,18 @@ defmodule Reaper.Decoder.Json do
 
     {:ok, data}
   rescue
-    error -> {:error, error}
+    error ->
+      parse_json_file_query_errors(error)
   end
 
   def truncate_file_for_logging(filename) do
     File.stream!(filename, [], 1000) |> Enum.at(0)
+  end
+
+  defp parse_json_file_query_errors(error) do
+    case error do
+      %Jaxon.ParseError{unexpected: :end_array} -> {:ok, []}
+      _ -> {:error, error}
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


apps/reaper/lib/reaper/decoder/json.ex
```diff
@@ -47,10 +47,18 @@ defmodule Reaper.Decoder.Json do
...
   rescue
-    error -> {:error, error}
+    error ->
+      parse_json_file_query_errors(error)
   end
 
   def truncate_file_for_logging(filename) do
     File.stream!(filename, [], 1000) |> Enum.at(0)
   end
+
```
dunson062786: So basically you're just matching on Jaxon parse errors on unexpected end_array and returning an ok tuple with an empty array.

apps/reaper/test/unit/reaper/decoder/json_test.exs
```diff
@@ -63,6 +63,24 @@ defmodule Reaper.Decoder.JsonTest do
...
+
+      {:ok, result} = Decoder.Json.decode({:file, @filename}, dataset_with_selector)
+
+      assert result == []
+
+      where([
+        [:body, :selector],
+        [%{data: []}, "$.data[*]"],
+        [%{data: %{nested_data: []}}, "$.data.nested_data[*]"],
+        [%{data: [%{nested_data: []}]}, "$.data[*].nested_data"]
```
bbrewer-papa-x: I'd be interested to know what the positive case for this kind of nesting is, but I "think" that even then, a single empty list works for that

bbrewer-papa-x: It "might?" need to be tweaked to be `$.data[*].nested_data[*]` to be more accurate as to the intent, but once again, not sure if it really matters

PR #599, opened 2020-03-19
Tableau connector api

LtChae:
Taking suggestions for the router paths

### DIFF
--- a/apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
+++ b/apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
@@ -11,6 +11,7 @@ defmodule DiscoveryApi.Data.TableInfoCache do
 
   def put(data) do
     Cachex.put(cache_name(), "table_info", data)
+    data
   end
 
   def get() do
--- a/apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
+++ b/apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
@@ -1,0 +1,27 @@
+defmodule DiscoveryApi.Data.TableInfoCache do
+  @moduledoc """
+  Simple module to cache systemName to dataset_id mapping
+  """
+  require Logger
+
+  def child_spec([]) do
+    Supervisor.child_spec({Cachex, cache_name()}, id: __MODULE__)
+
+  end
+
+  def put(data) do
+    Cachex.put(cache_name(), "table_info", data)
+  end
+
+  def get() do
+    Cachex.get!(cache_name(), "table_info")
+  end
+
+  def invalidate() do
+    Cachex.clear(cache_name())
+  end
+
+  defp cache_name() do
+    :table_info
+  end
+end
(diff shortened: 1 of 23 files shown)


apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def invalidate() do
+    Cachex.clear(cache_name())
+  end
+
+  defp cache_name() do
+    :table_info
+  end
+end
```
mcsearchin: Not having looked at the other cache(s), would it make sense to consolidate any of these cachex boilerplate into a module that can be extended?

kcarra:
I'm cool with the router path names :sungla

apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
```diff
@@ -51,6 +53,41 @@ defmodule DiscoveryApiWeb.MultipleMetadataController do
...
+        filtered_models ->
+          filtered_models
+      end
+
+    render(
+      conn,
+      :fetch_table_info,
+      models: filtered_models
+    )
+  end
```
mcsearchin: Is there any merit to the idea of putting tableau-related stuff into its own controller?  Not saying you need to make a bunch of changes - just a thought.

apps/discovery_api/lib/discovery_api_web/views/multiple_metadata_view.ex
```diff
@@ -31,6 +31,10 @@ defmodule DiscoveryApiWeb.MultipleMetadataView do
     }
   end
 
+  def render("fetch_table_info.json", %{models: models}) do
+    models
+  end
```
mcsearchin: Should this be `%{table_infos: table_infos}` - just to alleviate confusion?

apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
```diff
@@ -51,6 +53,41 @@ defmodule DiscoveryApiWeb.MultipleMetadataController do
...
 
+  def fetch_table_info(conn, _params) do
+    user_id = get_user_id(conn)
+
+    filtered_models =
+      case TableInfoCache.get(user_id) do
+        nil ->
+          remove_unauthorized_models(conn, Model.get_all())
+          |> get_filtered_table_info()
+          |> TableInfoCache.put(user_id)
```
mcsearchin: So, are we just caching with `nil` as the key if the user is anonymous?

LtChae: Yup

apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
kcarra: @mcsearchin I know that system_name_cache does something very similar to this.

apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
cameronmarsh: done

apps/discovery_api/lib/discovery_api/data/table_info_cache.ex
cameronmarsh: Good idea. We decided to punt on this though, maybe we should write a card for it?

PR #600, opened 2020-03-19
adjust validations and mapping to deal with schema as a struct

bbrewer-papa-x:
smartcitiesdata/smartcitiesdata/issues/577

co-authored by: Tony Brock <abrock17@gmail.com>


apps/andi/test/support/floki_helpers.ex
```diff
@@ -39,4 +49,9 @@ defmodule FlokiHelpers do
     |> Floki.parse_fragment!()
     |> Floki.find(selector)
   end
+
```
bbrewer-papa-x: we'll remove this

apps/discovery_api/lib/discovery_api/application.ex
```diff
@@ -27,8 +27,7 @@ defmodule DiscoveryApi.Application do
         {Brook, Application.get_env(:discovery_api, :brook)},
         DiscoveryApi.Data.CachePopulator,
         supervisor(DiscoveryApiWeb.Endpoint, []),
-        DiscoveryApi.Quantum.Scheduler,
```
cameronmarsh: Pretty much all of the changes from here down are reverting the changes made for the tableau connector endpoint. We are going to fix this too

PR #602, opened 2020-03-21
updating readme to reflect more recent local dev process

jeffgrunewald:
Removes readme references to deprecated docker compose instructions.


README.md
```diff
@@ -28,15 +28,11 @@ Finally, user access, discovery, and analysis is facilitated by a ReactJS web ap
...
-
-### stopping resources
-```bash
-docker-compose down
-```
+### starting the entire stack in minikube or Docker Desktop embedded Kubernetes cluster
+The current best approach to locally running the stack is in a local instance of Kubernetes, either in the minikube virtual machine or in the Kubernetes instance that can be run natively from Docker Desktop for Mac or Docker Desktop for Windows. Both options are viable, although in recent versions of the Docker version, exposing services via a `LoadBalancer` type allow the service to be reachable from the host machine without additional network manipulation.
+
+Once you have a Kubernetes cluster running, check out the `smartcitiesdata/charts` repo and the `platform` chart for standing up the complete platform or any ad hoc components you'd like to enable.
+
```
LtChae: Is it worth noting that individual apps are best started using their divo setups?

jeffgrunewald: maybe but not necessarily in the context of the open issue re: "how to start all the things?"

PR #604, opened 2020-03-24
Ldap purge

LtChae:
+ Removes LDAP and Paddle references from Discovery API


apps/discovery_api/lib/discovery_api_web/router.ex
```diff
@@ -30,25 +30,16 @@ defmodule DiscoveryApiWeb.Router do
...
-
-    get("/login", LoginController, :login)
-  end
-
   scope "/api/v1", DiscoveryApiWeb do
     pipe_through([:reject_cookies_from_ajax, :verify_token, :ensure_authenticated, :global_headers])
 
     post("/logged-in", UserController, :logged_in)
-
-    get("/logout", LoginController, :logout)
```
cameronmarsh: Doesn't this logout endpoint work for auth0 as well? And if it does, don't we need this or is there another way to logout aside from clicking the login zone link?

apps/discovery_api/lib/discovery_api_web/controllers/login_controller.ex
```diff
@@ -1,71 +0,0 @@
...
-        |> Guardian.Plug.sign_in(username)
-        |> Guardian.Plug.remember_me(username)
-        |> text("#{username} logged in.")
-
-      {:error, :invalidCredentials} ->
-        render_error(conn, 401, "Not Authorized")
-    end
-  end
-
-  def logout(conn, _) do
```
JarredOlson: I do not believe that this is LDAP specific.  Do we still need it?

apps/discovery_api/lib/discovery_api_web/router.ex
cameronmarsh: sweet

Issue #609, opened 2020-03-27
Tableau WDC Displays Query on Edit


LtChae:
#### User Story
As a data consumer using tableau, I would like my connector to show the query I wrote when I open it for editing so that I do not lose work.

#### Acceptance Criteria: 
- [ ] When a connector in query mode is opened for editing in Tableau it should display the query in the query box.

#### Tech Notes: 
- A connector has to be created and used in tableau before it can be edited.

<img width="365" alt="Screen Shot 2020-03-27 at 9 01 44 AM" src="https://user-images.githubusercontent.com/5122881/77758616-a7c5dc00-7009-11ea-86fa-f3bfab253269.png">

#### Amigos: 
@LtChae @JarredOlson @carriebrinkley

PR #615, opened 2020-03-30
WDC Query persists on edit

LtChae:
This should make using the WDC much more user friendly in Query mode.

#609


LtChae:
There is no version bump here as it is going out with an already merged bug fix.

cameronmarsh:
sure looks like (gee) 2 pts to me

PR #617, opened 2020-03-30
Added Profiling

sukaant:


apps/valkyrie/config/integration.exs
```diff
@@ -51,3 +51,6 @@ config :valkyrie, :brook,
       namespace: "valkyrie:view"
     ]
   ]
+
+config :profiling,
```
LtChae: Elixir convention usually uses this like `config :app_or_library, key: value`. Since profiling isn't a library, we could add this to the configuration of valkyrie.

PR #624, opened 2020-04-01
Adding a v2 to the date provider

JarredOlson:
Enabling it to accept a format as a parameter

### DIFF
--- a/apps/providers/lib/providers/timestamp.ex
+++ b/apps/providers/lib/providers/timestamp.ex
@@ -3,7 +3,15 @@ defmodule Providers.Timestamp do
   This provider implementation returns the current timestamp in UTC as an ISO8601 compliant date string.
   """
   @behaviour Providers.Provider
+
   def provide("1", _opts) do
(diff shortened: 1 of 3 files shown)


apps/providers/lib/providers/timestamp.ex
```diff
@@ -3,7 +3,15 @@ defmodule Providers.Timestamp do
   This provider implementation returns the current timestamp in UTC as an ISO8601 compliant date string.
   """
   @behaviour Providers.Provider
+
   def provide("1", _opts) do
     NaiveDateTime.utc_now() |> NaiveDateTime.to_iso8601()
   end
+
+  def provide("2", opts) do
```
bbrewer-papa-x, line 11: I like this and approve. However, could the same thing not be accomplished via the existing `format` field?

bbrewer-papa-x: That is, defer formatting until valkyrie

PR #627, opened 2020-04-02
updates to url_builder to evaluate EEx in s3 sourceUrl

kcarra:
smartcitiesdata/smartcitiesdata#296

co-authored-by: Tim Regan <tregan@pillartechnology.com>


LtChae:
For context, nothing else uses this slurper at the moment

PR #628, opened 2020-04-02
Added profiling

sukaant:


apps/reaper/lib/reaper/data_extract/load_stage.ex
```diff
@@ -113,6 +109,17 @@ defmodule Reaper.DataExtract.LoadStage do
     end
   end
 
+  defp add_timing(state) do
+    start = format_date(state.start_time)
+    stop = format_date(DateTime.utc_now())
+    timing = %{app: "reaper", label: "Ingested", start_time: start, end_time: stop}
```
LtChae, line 115: Not sure how much the date math here costs in cpu, but it is done for each message. It might be worth putting this behind the profiling condition as well so its only run when needed.

PR #635, opened 2020-04-07
Added Profiling configured to false by default

sukaant:
smartcitiesdata/smartcitiesdata#626

### DIFF
--- a/apps/forklift/lib/forklift/data_writer.ex
+++ b/apps/forklift/lib/forklift/data_writer.ex
@@ -171,11 +171,23 @@ defmodule Forklift.DataWriter do
   end
 
   defp add_total_time(datum) do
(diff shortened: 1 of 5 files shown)


apps/forklift/lib/forklift/data_writer.ex
```diff
@@ -171,11 +171,23 @@ defmodule Forklift.DataWriter do
   end
 
   defp add_total_time(datum) do
+    case Application.get_env(:forklift, :profiling_enabled) do
+      true -> add_timing(datum)
+      false -> remove_timing(datum)
+    end
+    |> Forklift.Util.remove_from_metadata(:forklift_start_time)
```
bbrewer-papa-x, line 178: What is putting this on there? Should that be removed too?

PR #637, opened 2020-04-09
Reaper provision expansion

LtChae:
+ This is in support of a data ingest card
+ Extends the timestamp provider with a version 2 that takes format, offset, and timezone
+ Allows hosted file ingestions to use providers. (This was not required for the card but it will save us from ourselves later)

### DIFF
--- a/apps/providers/test/unit/providers/date_test.exs
+++ b/apps/providers/test/unit/providers/date_test.exs
@@ -23,7 +23,8 @@ defmodule Providers.DateTest do
 
       result = Providers.Date.provide("1", %{offset_in_days: offset_in_days})
 
-      assert offset_in_days >= Timex.diff(Timex.parse!(result, @default_format), Timex.now(), :day)
+      assert offset_in_days >=
+               Timex.diff(Timex.parse!(result, @default_format), Timex.now(), :day)
     end
 
     test "provides a date with offset applied before formatting" do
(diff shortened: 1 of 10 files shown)


apps/reaper/lib/reaper/file_ingest/processor.ex
```diff
@@ -17,7 +18,8 @@ defmodule Reaper.FileIngest.Processor do
   @doc """
   Process a hosted dataset
   """
-  def process(%SmartCity.Dataset{} = dataset) do
+  def process(%SmartCity.Dataset{} = unprovisioned_dataset) do
+    dataset = Provisioner.provision(unprovisioned_dataset)
```
dunson062786, line 22: What does Provisioner.provision do?

apps/providers/test/unit/providers/date_test.exs
```diff
@@ -32,7 +33,9 @@ defmodule Providers.DateTest do
 
       result = Providers.Date.provide("1", %{format: format, offset_in_days: offset_in_days})
 
-      assert Timex.now() |> Timex.add(Timex.Duration.from_days(offset_in_days)) |> Timex.format!(format) == result
+      assert Timex.now()
```
bbrewer-papa-x, line 36: what if this runs right at midnight? :D

LtChae: Dunno. I suspect its not worth trying to solve that particular edge case

PR #638, opened 2020-04-13
bumping elsa version to pull in several updates

bbrewer-papa-x:


bbrewer-papa-x:
Yeah, this aint happening in a day...

PR #642, opened 2020-04-13
bringing in updated help text

bbrewer-papa-x:
Will sit below the query box, like so:

image


LtChae:
An alternative could be to put this in its own html file behind a help link on the connector. Looks like the images haven't been updated for the interactive connector or the login process, we should probably take care of that.

PR #651, opened 2020-04-20
Fields don't get reordered and mismatched after editing

cameronmarsh:


apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -41,7 +41,16 @@ defmodule Andi.InputSchemas.InputConverter do
   end
 
   def form_data_to_ui_changeset(form_data \\ %{}) do
-    form_data_as_params = adjust_form_input(form_data)
+    form_data_as_params =
+      form_data
+      |> Map.update("technical", %{}, fn technical ->
```
LtChae, line 46: Is it worth breaking this out into a named function that describes what's happening here?

cameronmarsh: yeah probably

PR #652, opened 2020-04-22
Live update

sukaant:


apps/estuary/lib/estuary_web/live/event_live_view/table.ex
```diff
@@ -10,14 +10,14 @@ defmodule EstuaryWeb.EventLiveView.Table do
     <div id="<%= @id %>" class="events-index__table">
       <table class="events-table">
       <thead>
-        <th class="events-table__th events-table__cell phx-value-field="author">Author </th>
-        <th class="events-table__th events-table__cell phx-value-field="create_ts">Create Timestamp </th>
-        <th class="events-table__th events-table__cell phx-value-field="data">Data </th>
-        <th class="events-table__th events-table__cell phx-value-field="type">Type </th>
+        <th class="events-table__th events-table__cell" phx-value-field="author">Author </th>
```
bbrewer-papa-x, line 13: Good catch

apps/estuary/lib/estuary/message_handler.ex
```diff
@@ -5,18 +5,26 @@ defmodule Estuary.MessageHandler do
...
   alias Estuary.DataWriter
 
+  @updated_event_stream "updated_event_stream"
+
   def handle_messages(messages) do
     messages
     |> Enum.map(fn message ->
       message.value
       |> Jason.decode!()
+      |> broadcast_event()
```
bbrewer-papa-x, line 15: One alternative here would be to send the batch of events to the live view via the broadcast. However, I think that there will probably be few enough of them that one at a time is probably fine for now.

apps/estuary/lib/estuary_web/router.ex
```diff
@@ -26,6 +26,7 @@ defmodule EstuaryWeb.Router do
 
     get("/", Redirect, to: "/events")
     live("/events", EventLiveView)
+    live("/streaming-events", StreamingEventLiveView, layout: {EstuaryWeb.LayoutView, :root})
```
bbrewer-papa-x, line 29: You might consider adding the `layout` part to the events view too, as I "think" that will be missing its styling if you don't.

apps/estuary/lib/estuary_web/templates/layout/root.html.eex
```diff
@@ -0,0 +1,19 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    <%= csrf_meta_tag() %>
+    <title>Estuary · Data Curator Interface</title>
+    <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
+  </head>
+  <body>
+    <div class="root">
+      <main role="main" class="root__main">
+        <%= render @view_module, @view_template, assigns %>
```
bbrewer-papa-x, line 14: When you merge in Cam's stuff, this will need to change to `<%= @inner_content %>`
```suggestion
       <%= @inner_content %>
```

PR #659, opened 2020-04-23
Changing streaming events view

sukaant:


apps/estuary/lib/estuary_web/live/streaming_event_live_view.ex
```diff
@@ -7,23 +7,32 @@ defmodule EstuaryWeb.StreamingEventLiveView do
...
   def mount(_params, _session, socket) do
     EstuaryWeb.Endpoint.subscribe(@updated_event_stream)
-    {:ok, assign(socket, events: [], order: {"create_ts", "desc"}, params: %{})}
+    {:ok, assign(socket, events: nil, order: {"create_ts", "desc"}, params: %{})}
   end
 
   def handle_info(%{topic: @updated_event_stream, payload: %{events: events}}, socket) do
     updated_events =
-      (events ++ socket.assigns.events)
+      validate_events(events, socket.assigns.events)
```
LtChae, line 23: Using [List.wrap/1](https://hexdocs.pm/elixir/List.html#wrap/1) on socket events could help simplify this logic.

apps/estuary/lib/estuary_web/live/streaming_event_live_view.ex
```diff
@@ -7,23 +7,32 @@ defmodule EstuaryWeb.StreamingEventLiveView do
...
   end
+
+  defp validate_events(events, socket_events) do
+    cond do
+      events == nil and (socket_events == nil or socket_events == [nil]) -> nil
+      events == nil -> socket_events
+      socket_events == nil or socket_events == [nil] -> events
+      true -> events ++ socket_events
+    end
+  end
```
LtChae, line 37: We prefer to use multiple function heads to deal with this sort of conditional where possible.

LtChae: I also think this could be simplified. For example, lines 32 and 34 are functionally identical.

PR #660, opened 2020-04-24
Selected field as default parent

cameronmarsh:


apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -332,8 +333,11 @@ defmodule AndiWeb.EditLiveView do
     {:noreply, assign(socket, changeset: changeset, add_data_dictionary_field_visible: true)}
   end
 
-  def handle_info({:assign_editable_dictionary_field, field}, socket) do
-    {:noreply, assign(socket, current_data_dictionary_item: field, selected_field_id: input_value(field, :id))}
+  def handle_info({:assign_editable_dictionary_field, field_id, index, name, id}, socket) do
+    new_form = find_field_changeset(socket.assigns.changeset, field_id) |> form_for(nil)
```
bbrewer-papa-x, line 337: This is why changesets are not fun to work with :D It will do for now

PR #669, opened 2020-05-01
Filter events

sukaant:


apps/estuary/lib/estuary_web/live/live_view_helper.ex
```diff
@@ -0,0 +1,23 @@
...
+  def filter_events(events, nil), do: events
+
+  def filter_events(events, ""), do: events
+
+  def filter_events(events, value) do
+    Enum.filter(events, fn event ->
+      search_contains?(event["author"], value) ||
+        search_contains?(event["create_ts"], value) ||
+        search_contains?(event["data"], value) ||
+        search_contains?(event["type"], value)
```
cameronmarsh, line 11: Do all of these need to be true, or just one of them?

apps/estuary/lib/estuary_web/live/live_view_helper.ex
```diff
@@ -0,0 +1,23 @@
...
+    end)
+  end
+
+  defp search_contains?(str, search_str) when is_integer(str) do
```
cameronmarsh, line 15: I would probably rename `str` here to something more descriptive and related to the clause that this is an integer

apps/estuary/lib/estuary_web/live/streaming_event_live_view.ex
```diff
@@ -20,18 +40,41 @@ defmodule EstuaryWeb.StreamingEventLiveView do
...
-      |> Enum.filter(&(!is_nil(&1)))
-      |> Enum.take(1000)
-      |> Enum.sort(&(&1["create_ts"] >= &2["create_ts"]))
+      |> refresh_events(socket.assigns.search_text)
 
     {:noreply, assign(socket, :events, updated_events)}
   end
+
+  def handle_params(params, _uri, socket) do
+    filtered_events = refresh_events(socket.assigns.events, params["search"])
```
cameronmarsh, line 57: The way this is written, we are dropping events that come in through kafka that don't show up with the current search text. So if you delete whatever search text is there, there will be some events that will never show up. Im assuming we want to show all of the events that have come through regardless of what is in the search box since the previous reload?

apps/estuary/lib/estuary_web/live/live_view_helper.ex
sukaant: any one is fine

apps/estuary/lib/estuary_web/live/streaming_event_live_view.ex
sukaant: We don't care about that much, so I think it is fine if something is dropped

PR #670, opened 2020-05-04
Filter restriction

sukaant:


apps/estuary/lib/estuary_web/live/event_live_view.ex
```diff
@@ -56,4 +57,19 @@ defmodule EstuaryWeb.EventLiveView do
...
+        {:ok, events} = EventRetrievalService.get_all()
+        events
+
+      filter_param == "" ->
+        {:ok, events} = EventRetrievalService.get_all()
+        events
+
+      true ->
+        socket_events
+    end
```
LtChae, line 73: I'm a little confused at to what's happening here. There is no clear way to get to the `true` branch, and the other two are identical. Maybe clean it up with an `or` and put the check in the `cond`?

PR #674, opened 2020-05-13
Finalize form

cameronmarsh:


apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -203,18 +207,26 @@ defmodule AndiWeb.EditLiveView do
 
     Process.flag(:trap_exit, true)
 
+    crontab =
```
bbrewer-papa-x, line 210: Is there any way this could be put inside the live component itself? I'm guessing this is just for making it work with the form and we unmunge it on the way out?

bbrewer-papa-x: Or maybe just find a way to make it more clear that this conversion is not permanent?

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -398,6 +410,13 @@ defmodule AndiWeb.EditLiveView do
      )}
   end
 
+  def handle_info({:assign_crontab, crontab}, socket) do
```
bbrewer-papa-x, line 413: Dear liveview, I hate you

apps/andi/test/integration/andi_web/live/edit_live_view_test.exs
```diff
@@ -638,4 +638,93 @@ defmodule AndiWeb.EditLiveViewTest do
...
+      render_click([view, "finalize_form_editor"], "set_schedule")
+      html = render(view)
+
+      assert dataset.technical.cadence == get_crontab_from_html(html)
+      assert Enum.empty?(find_elements(html, ".finalize-form__error-msg--visible"))
+    end
+
+    test "handles five-character cronstrings", %{conn: conn} do
+      dataset = TDG.create_dataset(%{technical: %{cadence: "4 2 7 * *"}})
+      {:ok, _} = Datasets.update(dataset)
```
dunson062786, line 685: Interesting I never thought of asserting like this. Have you seen the test fail if it returns {:error, _}.

apps/andi/test/integration/andi_web/live/edit_live_view_test.exs
```diff
@@ -638,4 +638,93 @@ defmodule AndiWeb.EditLiveViewTest do
...
+
+      assert {:ok, view, html} = live(conn, @url_path <> dataset.id)
+
```
dunson062786, line 688: Why does this need an a assert, but the previous not need an assert?

apps/andi/lib/andi_web/live/edit_live_view.ex
cameronmarsh: Yeah, absolutely. Really no reason why it's there

apps/andi/lib/andi_web/live/edit_live_view.ex
cameronmarsh: 🍕

apps/andi/test/integration/andi_web/live/edit_live_view_test.exs
cameronmarsh: Yeah, it's not an assertion but there will be an error if it fails

apps/andi/test/integration/andi_web/live/edit_live_view_test.exs
cameronmarsh: This is part of what we are testing, so we would want the test to fail if it doesn't return what we'd expect. If the line above fails, that is not something wrong with the logic we are testing, so it's a real problem and not a test failure

PR #684, opened 2020-05-20
Finalizing buttons

cameronmarsh:


apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -406,7 +430,11 @@ defmodule AndiWeb.EditLiveView do
     socket = reset_save_success(socket)
 
     dataset = Datasets.get(socket.assigns.dataset.id)
-    changeset = InputConverter.andi_dataset_to_full_ui_changeset(dataset)
+
+    changeset =
+      dataset
+      |> InputConverter.andi_dataset_to_full_ui_changeset()
+      |> Map.put(:action, :update)
```
bbrewer-papa-x, line 437: I'd love to see us DRY up some of this weirdness with `:action` I think it's also in `finalize_validation`, unless that has changed

apps/andi/lib/andi_web/live/edit_live_view/finalize_form.ex
```diff
@@ -123,61 +112,39 @@ defmodule AndiWeb.EditLiveView.FinalizeForm do
...
     </div>
     """
   end
 
   def handle_event("set_schedule", _, socket) do
-    case validate_cron(socket.assigns.crontab_list) do
-      {:ok, new_cron} ->
-        Datasets.update_cadence(socket.assigns.dataset_id, new_cron)
-        send(self(), {:assign_crontab})
+    new_cron = cronlist_to_cronstring(socket.assigns.crontab_list)
```
bbrewer-papa-x, line 122: cool

apps/andi/test/integration/andi_web/controllers/dataset_controller_test.exs
```diff
@@ -88,7 +88,7 @@ defmodule Andi.CreateDatasetTest do
           "sourceUrl" => "https://example.com",
           "sourceType" => "stream",
           "sourceFormat" => "gtfs",
-          "cadence" => "9000",
+          "cadence" => "*/9000 * * * * *",
```
bbrewer-papa-x, line 91: beautiful

cameronmarsh:
I don't think it's a concern because saving as a draft is only saving the definition to ecto, which is andi specific. The api will only see those that have been sent through brook, and therefore have been "published".

PR #685, opened 2020-05-20
Advanced search endpoint

LtChae:
implements the mvp of advanced search, matching on text fields and excluding private datasets

### DIFF
--- a/apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
+++ b/apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
@@ -7,6 +7,7 @@ defmodule DiscoveryApiWeb.MultipleMetadataController do
   alias DiscoveryApi.Data.Model
   alias DiscoveryApi.Data.TableInfoCache
   alias DiscoveryApi.Search.{DataModelFilterator, DataModelFacinator, DataModelSearchinator}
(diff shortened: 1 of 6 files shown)


apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
```diff
@@ -53,6 +54,34 @@ defmodule DiscoveryApiWeb.MultipleMetadataController do
       reraise e, __STACKTRACE__
   end
 
+  def advanced_search(conn, params) do
+    sort_by = Map.get(params, "sort", "name_asc")
+    query = Map.get(params, "query", "")
+
+    with {:ok, models} <- DatasetSearchIndex.search(query) do
```
dunson062786, line 61: Would this be better as a case statement?

LtChae: It could, but I left it that way because this is only the first step and I didn't want to saddle the next person with turning it back into a with.

dunson062786: That's what I was wondering. Thanks for explaining.

LtChae: Turns out Credo demands I do so in order to let the build pass

PR #696, opened 2020-05-27
Advanced search facets

LtChae:


apps/discovery_api/config/integration.exs
```diff
@@ -49,6 +49,18 @@ config :discovery_api, :elasticsearch,
               type: "keyword",
               index: true
             },
+            keywordFacets: %{
```
bbrewer-papa-x, line 52: Did you add these to the runtime.exs as well? If you can find a way to make this suck less for DRYness, please do it :)

apps/discovery_api/lib/discovery_api/search/dataset_index.ex
```diff
@@ -222,8 +233,22 @@ defmodule DiscoveryApi.Search.DatasetIndex do
     |> Map.drop([:completeness])
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
     |> Map.new()
+    |> populate_org_facets()
```
bbrewer-papa-x, line 236: Just makes cents

apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
```diff
@@ -101,6 +103,29 @@ defmodule DiscoveryApiWeb.MultipleMetadataController do
...
+    api_accessible = parse_api_accessible(params)
+
+    case validate_facets(facets) do
+      {:ok, filter_facets} ->
+        opts =
+          [
+            query: query,
+            api_accessible: api_accessible,
+            keywords: Map.get(filter_facets, :keywords),
+            org_title: Map.get(filter_facets, :organization, []) |> List.first()
```
bbrewer-papa-x, line 118: I hope someday that this `org_title` name will flow downstream into the UI (in vX.0 of the API) someday so we're no longer filtering on the hyper ambiguous `organization`

apps/discovery_api/config/integration.exs
LtChae: Have now

apps/discovery_api/lib/discovery_api_web/controllers/multiple_metadata_controller.ex
LtChae: Might be a good enhancement once we get it wired up to the ui

PR #698, opened 2020-05-28
Removes server header from cowboy applications

dunson062786:
https://github.com/smartcolumbusos/scosopedia/issues/148

This code removes the server headers from our cowboy applications.


dunson062786:
sure thing @bbrewer-papa-x. Should I put the common file in https://github.com/smartcitiesdata/smart_city/tree/master/lib/smart_city?

apps/web/lib/web.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Web do
```
LtChae, line 1: Should probably delete this.

apps/web/lib/web/application.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Web.Application do
+  # See https://hexdocs.pm/elixir/Application.html
+  # for more information on OTP Applications
+  @moduledoc false
+
+  use Application
+
+  def start(_type, _args) do
```
LtChae, line 8: Does this need to be a supervised application or is it just a library?

apps/web/mix.exs
```diff
@@ -0,0 +1,34 @@
...
+    [
+      extra_applications: [:logger],
+      mod: {Web.Application, []}
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
```
LtChae, line 29: Clean up boilerplate where possible

apps/web/test/web_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule WebTest do
```
LtChae, line 1: Should probably have actual tests instead of a placeholder

apps/web/lib/web/application.ex
dunson062786: It's is just a library

LtChae: It doesn't need this application file then. See the Providers library in apps.

apps/web/test/web_test.exs
dunson062786: I'm not sure how I would test this. There is only one function in strip_server_header that has a non-default implementation and that is the info method. It does a Map.drop on the header field. I could create an implementation class that does a Map.drop() and then test that and have info call that function.

apps/web/mix.exs
dunson062786: How does it look now?

apps/web/lib/web/application.ex
dunson062786: I've deleted the application.ex file

apps/web/lib/web.ex
dunson062786: done

apps/web/mix.exs
dunson062786: I've deleted the boilerplate

apps/andi/test/integration/andi_web/controllers/dataset_controller_test.exs
```diff
@@ -356,6 +356,14 @@ defmodule Andi.CreateDatasetTest do
     end
   end
 
+  describe "dataset get" do
+    test "andi doesn't return server in response headers" do
+      {:ok, %Tesla.Env{headers: headers}} = get("/api/v1/datasets", headers: [{"content-type", "application/json"}])
+      IO.puts("headers is #{inspect(headers)}")
```
dunson062786, line 362: I'll remove this

apps/andi/test/integration/andi_web/controllers/dataset_controller_test.exs
```diff
@@ -356,6 +356,14 @@ defmodule Andi.CreateDatasetTest do
     end
   end
 
+  describe "dataset get" do
+    test "andi doesn't return server in response headers" do
+      {:ok, %Tesla.Env{headers: headers}} = get("/api/v1/datasets", headers: [{"content-type", "application/json"}])
```
bbrewer-papa-x, line 361: The test itself is fine, but I would suggest either pattern matching out the tuple you care about (assuming you know exactly what headers are coming back) or turning it into a map and checking if it has the `server` key:
```suggestion
      {:ok, %Tesla.Env{headers: headers}} = get("/api/v1/datasets", headers: [{"content-type", "application/json"}])
      refute headers |> Map.new() |> Map.has_key?("server")
```

apps/web/test/web_test.exs
dunson062786: I've written integration test for andi, discovery_api, and discovery_streams to test this functionality.

PR #721, opened 2020-06-10
Add data name

cameronmarsh:


apps/andi/assets/css/app.scss
```diff
@@ -48,7 +48,7 @@ body {
 
 .error-msg {
   color: red;
-  display: block;
+  display: block !important;
```
bbrewer-forge, line 51: wait, what?! is there a reason we have to do this for css we control? Or is this overriding a 3rd party component?

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -135,13 +135,16 @@ defmodule Andi.InputSchemas.Datasets do
   end
 
   def is_unique?(id, data_name, org_name) do
+    org_name = String.downcase(org_name)
+    data_name = String.downcase(data_name)
+
     get_all()
```
bbrewer-forge, line 141: I'd love to see this done in a query, as it's likely more performant, but it's not a show stopped

apps/andi/assets/css/app.scss
dunson062786, line 51: Why are you using !important?

apps/andi/lib/andi/input_schemas/datasets.ex
bbrewer-forge: Additionally, is there a test for this? That might make it slightly less painful to migrate this to a query later.

apps/andi/assets/css/metadata_form.scss
```diff
@@ -1,6 +1,7 @@
 .metadata-form-edit-section {
   grid-template-areas:
     "title last-updated"
+    "data-name ."
```
JarredOlson, line 4: do you want that space dot on the end?

apps/andi/assets/css/app.scss
cameronmarsh: yeah we kind of just gave up going through the cascade but we will change this back

cameronmarsh: it was more just a temporary fix but we will change this back

apps/andi/lib/andi/input_schemas/datasets.ex
cameronmarsh: good point, there are some tests so we can do this

apps/andi/assets/css/metadata_form.scss
bbrewer-forge: yeah, it's the goofy grid css way of adding a blank (or stretching it?) cell

cameronmarsh: yeah I think that is how you denote an empty space when defining a grid elsewhere in andi, but the styling is not currently lining up so I will look into it some more

PR #728, opened 2020-06-12
Fix more date issues

LtChae:


apps/discovery_api/config/integration.exs
```diff
@@ -54,6 +54,14 @@ config :discovery_api, :elasticsearch,
               type: "keyword",
               index: true
             },
+            modifiedDate: %{
```
cameronmarsh, line 57: I don't much about mix config, but is there any way to combine this with the runtime.exs equivalent? Kinda sucks having this in two places

LtChae: Might be possible with the new elixir config standard.

PR #748, opened 2020-06-22
add add dataset button

bbrewer-forge:


apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -71,6 +75,22 @@ defmodule AndiWeb.DatasetLiveView do
     {:noreply, assign(socket, search_text: search_text, datasets: view_models, order: %{order_by => order_dir}, params: params)}
   end
 
+  def handle_event("add-dataset", _, socket) do
+    new_dataset_id = UUID.uuid4()
+    new_dataset_title = "New Dataset - #{Date.utc_today()}"
+    new_dataset_name = Datasets.data_title_to_data_name(new_dataset_title)
+
+    new_changeset =
```
bbrewer-forge, line 83: Cam and I have a TODO tomorrow of breaking this out into its own function. Probably `Datasets.create` or `Datasets.new` or something like that. The business logic with the date would stay here, however.

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -134,11 +134,21 @@ defmodule Andi.InputSchemas.Datasets do
       {:ok, get(dataset_id)}
   end
 
+  def is_unique?(_id, data_name, org_name) when is_nil(data_name) or is_nil(org_name), do: true
```
bbrewer-forge, line 137: Another TODO to "unit" test this as part of the `Datasets` module integration test

PR #758, opened 2020-06-29
Used datasets

LtChae:
These changes allow Discovery API to keep track of datasets used as parts of visualizations. They include a migrator to ensure that all existing visualizations are updated with their used datasets.


apps/discovery_api/lib/discovery_api/data/visualization_migrator.ex
```diff
@@ -0,0 +1,56 @@
...
+        Logger.info("Visualization Migration Failed. Retrying.")
+        Process.sleep(2000)
+        {:noreply, nil, {:continue, :run}}
+    end
+  end
+
+  defp migrate() do
+    try do
+      Logger.info("Starting Migration of Visualizations")
+      query = from(v in Visualization, where: is_nil(v.valid_query))
```
bbrewer-forge, line 42: I'd love it if the relationship between valid_query and the stuff done in `Visulizations` was more clear, but I can't think of a good way off the top of my head.

bbrewer-forge: Maybe the query for non-updated visualizations could be owned by `Visualizations` ? That way, the field it's checking doesn't need to be spread across two modules?

apps/discovery_api/lib/discovery_api/schemas/visualizations.ex
```diff
@@ -42,11 +45,48 @@ defmodule DiscoveryApi.Schemas.Visualizations do
...
+  defp add_used_datasets(visualization, nil), do: visualization
+
+  defp add_used_datasets(visualization, query) do
+    DiscoveryApi.prestige_opts()
+    |> Prestige.new_session()
+    |> Prestige.query("EXPLAIN (TYPE IO, FORMAT JSON) #{query}")
+    |> handle_explain_response(visualization)
+  end
+
+  defp handle_explain_response({:ok, response}, visualization) do
```
bbrewer-forge, line 67: Are some of these operations already done, to some extent, by the stuff under the authorization? Also, does there need to be some sort of authorization here too, so we don't leak private datasets via this field?

PR #759, opened 2020-06-29
Alert modal for unsaved changes

kcarra:


apps/andi/assets/css/edit.scss
```diff
@@ -138,6 +138,23 @@ $component-header-height: 7.5rem;
     margin-left: 1rem;
 }
 
+.unsaved-changes-modal--visible {
```
bbrewer-forge, line 141: @cameronmarsh do you think it's worthwhile to do a sass mixin for the modal to dry this up?

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -16,7 +16,26 @@ defmodule AndiWeb.EditLiveView do
...
     dataset_id = assigns.dataset.id
 
+    modifier =
+      case assigns.show_unsaved_changes_modal do
+        true -> "visible"
+        false -> "hidden"
+      end
+
     ~L"""
+    <div class="unsaved-changes-modal unsaved-changes-modal--<%= modifier %>">
```
bbrewer-forge, line 26: How are we doing the other modal? Is it its own live component? Just curious if this is a similar setup to that one, to avoid bloating this file any more. If it's consistent though, that's fine.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -16,7 +16,26 @@ defmodule AndiWeb.EditLiveView do
...
+        <p class="unsaved-changes-modal__message">
+          You have unsaved changes within this<br> section. Do you wish to continue without saving?
+        </p>
+        <br>
+        <div class="button-container">
+          <button type="button" class="btn" phx-click="unsaved-changes-canceled">Cancel</a>
+          <button type="button" class="btn submit_button" phx-click="force-cancel-edit">Continue</a>
+        </div>
+      </div>
+    </div>
```
cameronmarsh, line 38: It might be worth pulling this out into its own component like the add/delete field modals

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -127,13 +148,15 @@ defmodule AndiWeb.EditLiveView do
...
     |> InputConverter.form_data_to_ui_changeset()
     |> complete_validation(socket)
+    |> mark_changes()
   end
 
   def handle_event("validate", %{"form_data" => form_data, "_target" => ["form_data", "technical", "sourceQueryParams" | _]}, socket) do
     form_data
     |> FormTools.adjust_source_url_for_query_params()
     |> InputConverter.form_data_to_ui_changeset()
     |> complete_validation(socket)
```
bbrewer-forge, line 158: Are there any cases of `complete_validation` that are not paired with `mark_changes`. If not, I'd strongly consider just calling it from that function, unless we feel like this is more explicit.

kcarra: It gets called on save as well. That's why i just made it separate

bbrewer-forge: That'll do. That'll do

apps/andi/assets/css/edit.scss
cameronmarsh: yeah I think that it would be worth it. Shouldn't be too hard

cameronmarsh: actually there's already a class so a mixin might not even be necessary

kcarra:
> I know this isn't a part of this card exactly, but my only question would be can we also somehow prevent the user from hitting the back button without warning them?

For the sake of getting this card across the finish line, let make this a separate effort

PR #780, opened 2020-07-08
Adding Telemetry to Valkyrie

sukaant:


apps/valkyrie/lib/valkyrie/dataset_handler.ex
```diff
@@ -40,4 +49,14 @@ defmodule Valkyrie.DatasetHandler do
 
     delete(:datasets, dataset.id)
   end
+
+  defp add_event_count(event_type, dataset_id) do
+    [
+      app: "valkyrie",
+      author: "valkyrie",
```
LtChae, line 56: Author should be the app that created the event, not the one that's reading it

apps/valkyrie/test/integration/dataset_mutation_test.exs
```diff
@@ -20,6 +20,10 @@ defmodule Valkyrie.DatasetMutationTest do
 
     data1 = TDG.create_data(dataset_id: @dataset_id, payload: %{"age" => "21"})
 
+    assert :ok ==
```
LtChae, line 23: What do these assertions test?

apps/valkyrie/test/integration/end_of_data_test.exs
```diff
@@ -35,6 +35,9 @@ defmodule Valkyrie.EndOfDataTest do
 
     eod_message = end_of_data()
 
+    eod_message
```
LtChae, line 38: There doesn't seem to be a test that uses this

PR #789, opened 2020-07-14
Schema order fix

LtChae:
These changes create a field to track the order of fields in a schema. They are autogenerated as an increasing integer whenever the schema is edited.

The sequence field itself is editable, so we can easily use it to implement a reordering flow on the ui in the future if needed.

### DIFF
--- a/apps/andi/lib/andi/input_schemas/datasets/data_dictionary.ex
+++ b/apps/andi/lib/andi/input_schemas/datasets/data_dictionary.ex
@@ -21,6 +21,7 @@ defmodule Andi.InputSchemas.Datasets.DataDictionary do
     field(:rationale, :string)
     field(:bread_crumb, :string)
     field(:format, :string)
+    field(:sequence, :integer, read_after_writes: true)
     has_many(:subSchema, __MODULE__, foreign_key: :parent_id, on_replace: :delete)
 
     belongs_to(:data_dictionary, __MODULE__, type: Ecto.UUID, foreign_key: :parent_id)
@@ -46,7 +47,8 @@ defmodule Andi.InputSchemas.Datasets.DataDictionary do
     :technical_id,
     :parent_id,
     :bread_crumb,
-    :format
+    :format,
+    :sequence
   ]
   @required_fields [
     :name,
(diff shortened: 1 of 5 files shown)


apps/andi/lib/andi/input_schemas/datasets/data_dictionary.ex
```diff
@@ -21,6 +21,7 @@ defmodule Andi.InputSchemas.Datasets.DataDictionary do
     field(:rationale, :string)
     field(:bread_crumb, :string)
     field(:format, :string)
+    field(:sequence, :integer, read_after_writes: true)
```
bbrewer-forge, line 24: nice!

apps/andi/lib/andi/input_schemas/struct_tools.ex
```diff
@@ -71,6 +71,12 @@ defmodule Andi.InputSchemas.StructTools do
     struct(struct_type, preloaded)
   end
 
+  defp sort_if_sequenced([%{sequence: _sequence} | _] = list) do
```
bbrewer-forge, line 74: Cool, nicely done

apps/andi/lib/andi/input_schemas/struct_tools.ex
```diff
@@ -71,6 +71,12 @@ defmodule Andi.InputSchemas.StructTools do
...
+    Enum.sort(list, fn %{sequence: sequence1}, %{sequence: sequence2} -> sequence1 < sequence2 end)
```
bbrewer-forge, line 75: Is there a reason why `Enum.sort_by(&Map.get(&1, :sequence))` doesn't work here?

apps/andi/priv/repo/migrations/20200709175229_add_sequence.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Andi.Repo.Migrations.AddSequence do
+  use Ecto.Migration
+
+  def change do
+    alter table(:data_dictionary) do
+      add :sequence, :serial
+    end
+    execute "select setval(pg_get_serial_sequence('data_dictionary', 'sequence'), 1)"
```
bbrewer-forge, line 8: wowowow

apps/andi/lib/andi/input_schemas/struct_tools.ex
LtChae: Probably would. I'll look into it

PR #794, opened 2020-07-17
Added dead_letter to telemetry

sukaant:


apps/dead_letter/test/unit/dead_letter_test.exs
```diff
@@ -95,6 +95,8 @@ defmodule DeadLetterTest do
     end
 
     test "returns formatted DLQ message with a reason" do
+      allow(TelemetryEvent.add_event_count(any(), [:dead_letters_handled]), return: :ok)
```
LtChae, line 98: We should have tests for these that `expect` rather than `allow` so we know that the metrics events are being created.

apps/dead_letter/lib/dead_letter/server.ex
```diff
@@ -108,4 +110,12 @@ defmodule DeadLetter.Server do
   defp get_stacktrace({_, stacktrace}) when is_list(stacktrace) do
     stacktrace
   end
+
+  defp add_dead_letter_count(dataset_id, reason) do
```
LtChae, line 114: Can this raise an exception? If so, will it kill the whole dead_letter process? If possible, I feel that we should tolerate an exception here and log it. Updating the metrics is not a critical function.

PR #804, opened 2020-07-20
version bump to mix.lock

kcarra:


apps/andi/lib/andi/harvest/data_json_dataset_mapper.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Andi.Harvest.DataJsonToDataset do
+  @moduledoc """
+  maps data json to %SmartCity.Dataset{}
+  """
+  alias Andi.InputSchemas.Datasets
+
+  @scos_data_json_uuid "1719bf64-38f5-40bf-9737-45e84f5c8419"
```
JarredOlson, line 7: I would probably just call this `seed` or something

apps/andi/lib/andi/harvest/data_json_dataset_mapper.ex
```diff
@@ -0,0 +1,83 @@
...
+        "sourceUrl" => source_url(data_json_dataset["distribution"]),
+        "systemName" => system_name(org.orgName, data_name(data_json_dataset["title"]))
+      }
+    }
+  end
+
+  defp access_level(access_level) do
+    case access_level do
+      "non-public" -> true
+      "public" -> false
```
sukaant, line 61: Do we need to check "public" -> false here. I think it can be handled in default case

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -0,0 +1,46 @@
...
+        {:ok, response.body}
+
+      {:error, error} ->
+        Logger.error("Failed to get data json from #{url}: #{inspect(error)}")
+        {:error, error}
+    end
+  end
+
+  def map_data_json_to_dataset(data_json, org) do
+    DataJsonToDataset.mapper(data_json, org)
```
JarredOlson, line 40: It will probably make sense to pull this into a shared library with the dataset -> data.json mapping we use in discovery, but that's another effort.

apps/andi/lib/andi/harvest/data_json_dataset_mapper.ex
```diff
@@ -0,0 +1,83 @@
...
+    access_url
+  end
+
+  defp source_url(_distribution), do: ""
+
+  defp data_name(data_title) do
+    Datasets.data_title_to_data_name(data_title)
+  end
+
+  defp system_name(org_name, data_name) do
```
JarredOlson, line 76: Is this duplicated somewhere else?  If so I'd like to dry up that logic

kcarra: This is used in another function in andiweb that does a bunch of other logic. This is just a line that was pulled out

PR #811, opened 2020-07-23
add tooltip to failed url tests

bbrewer-forge:


apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -5,7 +5,7 @@ defmodule Andi.InputSchemas.InputConverter do
 
   alias Andi.InputSchemas.Datasets.Dataset
   alias Andi.InputSchemas.StructTools
-  alias Andi.InputSchemas.FormTools
+  alias AndiWeb.Helpers.FormTools
```
cameronmarsh, line 8: noice

apps/andi/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "0.32.1",
+      version: "0.32.3",
       build_path: "../../_build",
```
dunson062786, line 8: Did you mean to skip 2

apps/andi/lib/andi_web/live/edit_live_view/url_form.ex
```diff
@@ -229,6 +232,16 @@ defmodule AndiWeb.EditLiveView.UrlForm do
...
+  defp status_tooltip(%{status: status}) when status in 200..399, do: status_tooltip(%{status: status}, "hidden")
+
+  defp status_tooltip(%{status: status}, modifier \\ "shown") do
+    assigns = %{
+      description: HttpStatusDescriptions.get(status),
+      modifier: modifier
+    }
+
+    ~E(<sup class="test-status__tooltip-wrapper"><i phx-hook="addTooltip" data-tooltip-content="<%= @description %>" class="material-icons-outlined test-status__tooltip--<%= @modifier %>">info</i></sup>)
+  end
```
cameronmarsh, line 244: this is a cool pattern, we could probably stand to do this where we have some checks (like for visibility) at the beginning of our render functions

apps/andi/mix.exs
bbrewer-forge: good catch, oops

apps/andi/lib/andi_web/views/options.ex
```diff
@@ -110,8 +108,8 @@ defmodule Andi.InputSchemas.Options do
     }
   end
 
-  def organizations() do
-    case OrgStore.get_all() do
```
cameronmarsh, line 114: 👍

apps/andi/lib/andi_web/live/edit_live_view/url_form.ex
bbrewer-forge: Yeah, true. The only gotcha there would be hidden fields for forms (possibly). I can't recall if non-hidden form fields inside a non-displayed parent get send to the backend or not.

PR #814, opened 2020-07-27
Replacing nil reason with empty string for dead letter reason

sukaant:


apps/telemetry_event/lib/telemetry_event/helper/telemetry_event_helper.ex
```diff
@@ -32,13 +31,11 @@ defmodule TelemetryEvent.Helper.TelemetryEventHelper do
...
-  defp reject_nil(keyword_name, value) do
-    if keyword_name == :dataset_id or value != nil do
-      true
-    else
-      raise "Keyword :#{keyword_name} cannot be nil"
-    end
-  end
+  defp replace_nil(value) when is_nil(value), do: "UNKNOWN"
+
+  defp replace_nil(value) when value == "", do: "UNKNOWN"
```
LtChae, line 36: Minor, but these can be combined with an `||` in the guard clause

apps/telemetry_event/test/unit/telemetry_event_test.exs
```diff
@@ -12,38 +12,23 @@ defmodule TelemetryEventTest do
...
+               event_type: ""
              ]
              |> TelemetryEvent.add_event_count([:any_event_name])
   end
 
-  test "should return `:error` and error message when `event_type` is `nil`" do
-    expected_error = {:error, %RuntimeError{message: "Keyword :event_type cannot be nil"}}
-
-    assert expected_error ==
+  test "should return `:ok` any value is nil" do
```
LtChae, line 26: This test should probably check to make sure that fields are set to UNKNOWN as intended

PR #815, opened 2020-07-27
Download file as stream and add cancel button

cameronmarsh:
smartcitiesdata/smartcitiesdata#790

### DIFF
--- a/apps/andi/assets/js/app.js
+++ b/apps/andi/assets/js/app.js
@@ -35,8 +35,8 @@ Hooks.showSnackbar = {
 Hooks.readFile = {
     mounted() {
         this.el.addEventListener("change", e => {
-            var file = this.el.files[0];
             this.pushEvent("file_upload_started");
+            var file = this.el.files[0];
 
             fileToText(this.el.files[0]).then(fileAsText => {
                 this.pushEvent("file_upload", {
@@ -44,6 +44,8 @@ Hooks.readFile = {
                     fileType: file["type"],
                     fileSize: file["size"]
                 });
+            }, reason => {
+                reason == "cancel" & this.pushEvent("file_upload_cancelled");
             });
         })
     }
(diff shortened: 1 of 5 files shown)


apps/andi/assets/js/app.js
```diff
@@ -35,15 +35,17 @@ Hooks.showSnackbar = {
...
+            var file = this.el.files[0];
 
             fileToText(this.el.files[0]).then(fileAsText => {
                 this.pushEvent("file_upload", {
                     file: fileAsText,
                     fileType: file["type"],
                     fileSize: file["size"]
                 });
+            }, reason => {
+                reason == "cancel" & this.pushEvent("file_upload_cancelled");
```
bbrewer-forge, line 48: wait, what does `&` do here? Is it an XOR?

cameronmarsh: no its basically a shorthand if statement

apps/andi/lib/andi_web/live/edit_live_view/data_dictionary_form.ex
```diff
@@ -80,12 +80,15 @@ defmodule AndiWeb.EditLiveView.DataDictionaryForm do
...
-                    <%= ErrorHelpers.error_tag(f, :schema_sample, bind_to_input: false) %>
+                  <div class="file-input-button--<%= loader_visibility %>">
+                    <div class="file-input-button">
+                      <%= label(f, :schema_sample, "Upload data sample", class: "label") %>
+                      <%= file_input(f, :schema_sample, phx_hook: "readFile", accept: "text/csv, application/json") %>
+                      <%= ErrorHelpers.error_tag(f, :schema_sample, bind_to_input: false) %>
+                    </div>
                   </div>
 
+                  <button type="button" id="reader-cancel" class="file-upload-cancel-button file-upload-cancel-button--<%= loader_visibility %> btn">Cancel</button>
```
bbrewer-forge, line 91: How does this flow down to the "change" event? Is it part of the `file_input` internals?

cameronmarsh: There is an event listener in the `app.js` file, where the FileReader is being defined

apps/andi/assets/js/app.js
cameronmarsh: yeah that should be `&&`

PR #818, opened 2020-07-28
License help

LtChae:


apps/andi/lib/andi_web/views/error_helpers.ex
```diff
@@ -60,23 +60,16 @@ defmodule AndiWeb.ErrorHelpers do
...
+  defp interpret_error_message(message, :schema_sample), do: message
+  defp interpret_error_message("is required", field), do: default_error_message(field)
+  defp interpret_error_message(message, :format), do: "Error: " <> get_format_error_message(message)
+  defp interpret_error_message(message, field) when field in [:topLevelSelector, :cadence, :dataName], do: "Error: #{message}"
+  defp interpret_error_message(_message, field) when field in [:sourceHeaders, :sourceQueryParams], do: "Please enter valid key(s)."
+  defp interpret_error_message(_message, field), do: default_error_message(field)
 
-    {updated_message, opts}
-  end
+  defp default_error_message(field), do: "Please enter a valid #{get_downcased_display_name(field)}."
```
cameronmarsh, line 72: Nice. Are you just using the default for the `license` errors?

LtChae: I should probably add it to the case that prints the original message. The text is important

bbrewer-forge, line 72: Thank you - this place was not good

PR #824, opened 2020-07-28
Adding telemetry integration test to valkyrie

sukaant:


apps/dead_letter/lib/dead_letter/server.ex
```diff
@@ -114,7 +114,7 @@ defmodule DeadLetter.Server do
   defp add_dead_letter_count(dataset_id, reason) do
     [
       dataset_id: dataset_id,
-      reason: reason
+      reason: Kernel.inspect(reason)
```
LtChae, line 117: Kernel isn't needed here. Anything in Kernel is available like `inspect(reason)`

PR #830, opened 2020-07-30
Adding Telemetry Forklift

sukaant:


apps/telemetry_event/lib/telemetry_event/helper/telemetry_event_helper.ex
```diff
@@ -10,20 +10,32 @@ defmodule TelemetryEvent.Helper.TelemetryEventHelper do
...
-      Metrics.counter(Keyword.fetch!(metrics_option, :metric_name),
-        tags: Keyword.fetch!(metrics_option, :tags)
-      )
+      Keyword.fetch!(metrics_option, :metric_type)
+      |> String.upcase()
+      |> metrics_event(metrics_option)
     end)
   end
 
+  defp metrics_event(metric_type, metrics_option) when metric_type === "COUNTER" do
```
LtChae, line 27: By convention this is done like `defp metrics_event("COUNTER", metrics_option) do` to take advantage of Elixir's pattern matching.

In addition, we try to avoid hard-coded strings duplicated across apps. Look into either making them methods like we do with event names so that `telemetry_event` controls them entirely, or use atoms like `:counter`

apps/telemetry_event/lib/telemetry_event/telemetry_event.ex
```diff
@@ -9,8 +9,12 @@ defmodule TelemetryEvent do
     end
   end
 
-  def add_event_count(event_measurements, event_name) do
-    :telemetry.execute(event_name, %{}, TelemetryEventHelper.measurements(event_measurements))
+  def add_event_metrics(event_tags_and_values, event_name, measurement \\ []) do
+    :telemetry.execute(
+      event_name,
+      Keyword.get(measurement, :value, %{}),
```
LtChae, line 15: Should this be defaulted to an empty map?

apps/forklift/runtime.exs
```diff
@@ -129,3 +127,23 @@ if System.get_env("RUN_IN_KUBERNETES") do
...
 end
+
+config :telemetry_event,
+  metrics_port: System.get_env("METRICS_PORT") |> String.to_integer(),
+  metrics_options: [
+    [
+      metric_name: "events_handled.count",
+      tags: [:app, :author, :dataset_id, :event_type],
+      metric_type: "COUNTER"
+    ],
```
LtChae, line 138: Since these are duplicated across apps, should we look into having `telemetry_events` define these as functions so they're standardized?

LtChae: Note: this doesn't need to be a part of this pr

apps/telemetry_event/lib/telemetry_event/telemetry_event.ex
sukaant: Yes

apps/reaper/runtime.exs
```diff
@@ -110,7 +110,7 @@ config :telemetry_event,
   metrics_port: System.get_env("METRICS_PORT") |> String.to_integer(),
   metrics_options: [
     [
-      metric_name: "events_handled.count",
+      metric_type_and_name: [:counter, :events_handled, :count],
```
LtChae, line 113: Please use a tuple for this rather than trying to extract it via index

apps/telemetry_event/lib/telemetry_event/helper/telemetry_event_helper.ex
```diff
@@ -10,20 +10,33 @@ defmodule TelemetryEvent.Helper.TelemetryEventHelper do
...
   defp metrics() do
     Application.get_env(:telemetry_event, :metrics_options)
     |> Enum.map(fn metrics_option ->
-      Metrics.counter(Keyword.fetch!(metrics_option, :metric_name),
-        tags: Keyword.fetch!(metrics_option, :tags)
-      )
+      metric_type_and_name = Keyword.fetch!(metrics_option, :metric_type_and_name)
+      |> Enum.split(1)
+      
+      metric_name = metric_type_and_name |> elem(1)
```
LtChae, line 24: These should either be keyword params or pattern-matched from a tuple

apps/reaper/runtime.exs
LtChae: Maybe make the type its own key in the keyword list

PR #848, opened 2020-08-11
Andi bugs

cameronmarsh:


apps/andi/assets/js/app.js
```diff
@@ -28,7 +28,7 @@ Hooks.showSnackbar = {
     updated() {
         let snackbar = document.getElementById("snackbar");
         snackbar.className += " show";
-        setTimeout(function() { snackbar.className = snackbar.className.replace("show", ""); }, 3000);
+        setTimeout(function() { snackbar.className = snackbar.className.replace(" show", ""); }, 3000);
```
bbrewer-forge, line 31: oh

apps/andi/lib/andi_web/live/helpers/form_section.ex
```diff
@@ -15,6 +15,7 @@ defmodule AndiWeb.FormSection do
           socket.assigns.changeset
           |> Map.put(:action, :update)
 
+        AndiWeb.Endpoint.broadcast_from(self(), "form-save", "save-all", %{})
```
bbrewer-forge, line 18: Are these synchronous enough to not be a problem?

cameronmarsh: I haven't run into any issues with it, but that's a good point

apps/andi/assets/js/app.js
cameronmarsh: :-|

PR #853, opened 2020-08-13
Data json tracking

kcarra:


apps/andi/lib/andi/harvest/data_json_dataset_mapper.ex
```diff
@@ -6,7 +6,7 @@ defmodule Andi.Harvest.DataJsonToDataset do
 
   @scos_data_json_seed "1719bf64-38f5-40bf-9737-45e84f5c8419"
```
jmorris-pillar, line 8: I wish this file followed naming conventions.  Its module name does not match  its file  name

apps/andi/lib/andi/input_schemas/organizations.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Andi.InputSchemas.Organizations do
```
cameronmarsh, line 1: I find this module name a little confusing as it's getting and manipulating `HarvestedDatasets` instead of `Organizations`. Maybe if we want to keep this module name, we can rename some of the functions to reflect that? I can also change that when I add the organization table to Ecto

apps/andi/priv/repo/migrations/20200810140819_add_harvested_datasets.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Andi.Repo.Migrations.AddHarvestedDatasets do
+  use Ecto.Migration
+
+  def change do
+    alter table(:datasets) do
```
cameronmarsh, line 5: I thought we weren't manipulating this table and just making a new one instead?

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -37,10 +39,22 @@ defmodule Andi.Harvest.Harvester do
...
   def dataset_update(datasets) do
-    Enum.each(datasets, &Brook.Event.send(instance_name(), dataset_update(), :andi, &1))
+    Enum.each(datasets, fn dataset ->
+      Brook.Event.send(instance_name(), dataset_update(), :andi, dataset)
+    end)
+  end
+
+  def harvested_dataset_update(harvested_datasets) do
+    Enum.each(harvested_datasets, fn harvested_dataset ->
+      Brook.Event.send(instance_name(), dataset_harvest_end(), :andi, harvested_dataset)
```
jmorris-pillar, line 57: Is it okay that  we drop  brook event send failures here?

apps/andi/priv/repo/migrations/20200331174805_create_datasets.exs
```diff
@@ -7,6 +7,16 @@ defmodule Andi.Repo.Migrations.CreateDatasets do
       add :ingestedTime, :utc_datetime
     end
 
+    create table(:harvested_datasets, primary_key: false) do
```
cameronmarsh, line 10: I actually don't know if you can edit an old migration file or if its customary to, but I feel like we've generally just made new files/migrations to add

apps/andi/lib/andi/input_schemas/organizations.ex
kcarra: This is something @bbrewer-forge and I discussed. All the harvesting stuff is going to live alongside the organization code. There was no point in calling this something else related to harvesting then going back and changing it when the organizations page gets developed

bbrewer-forge: Yeah, hopefully this will get more fleshed out when we move the organization to the repo

apps/andi/priv/repo/migrations/20200810140819_add_harvested_datasets.exs
kcarra: my B. forgot to push the deleted migration

PR #856, opened 2020-08-17
Org list page

cameronmarsh:


apps/andi/lib/andi/input_schemas/organizations.ex
```diff
@@ -1,20 +1,25 @@
...
+
+  def get_all(), do: Repo.all(Organization)
+
+  def get_harvested_dataset(id) do
     Repo.get(HarvestedDatasets, id)
     |> HarvestedDatasets.preload()
   end
 
-  def get_all() do
+  def get_all_harvested_datasets() do
```
bbrewer-forge, line 22: I'd like to see this be there with an `org_id` argument, as that seems like the most common case we'll have, but feel free to burn that bridge when we get to it.

apps/andi/lib/andi_web/live/organization_live_view.ex
```diff
@@ -0,0 +1,126 @@
+defmodule AndiWeb.OrganizationLiveView do
```
bbrewer-forge, line 1: I'd love to see us take a crack at making this not repeat some of the problems with the datasets page (Ex. use the Repo + Query to do searching as it's arguably faster or at least doesn't kill the client machine with javascript based filtering). However, I'm fine with pushing forward if that's too much to pack on there.

apps/andi/lib/andi/input_schemas/organizations.ex
cameronmarsh: yup makes sense

apps/andi/lib/andi/event_handler.ex
```diff
@@ -37,6 +37,10 @@ defmodule Andi.EventHandler do
     organization_update()
     |> add_event_count(data.id)
 
+    data
```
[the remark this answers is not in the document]
cameronmarsh: no

apps/andi/lib/andi_web/live/organization_live_view.ex
cameronmarsh: yeah good call

bbrewer-forge:
WoooOOOooow

PR #877, opened 2020-08-21
Adding dataset record count

sukaant:


apps/pipeline/lib/pipeline/writer/table_writer/compaction.ex
```diff
@@ -66,4 +67,17 @@ defmodule Pipeline.Writer.TableWriter.Compaction do
       results.rows
     end
   end
+
+  def add_dataset_record_event_count(table_name, old_row_count, new_row_count) do
+    if !String.ends_with?(table_name, "__json") do
```
LtChae, line 72: The convention we use is to do this sort of logic via pattern matching in the function head.

apps/pipeline/lib/pipeline/writer/table_writer/compaction.ex
```diff
@@ -66,4 +67,17 @@ defmodule Pipeline.Writer.TableWriter.Compaction do
...
+      [
+        table_name: table_name,
+        old_row_count: old_row_count
```
LtChae, line 75: This should be `row_count` right? We don't care about old vs new so long as they are the same. To that end, what happens when the compaction is rejected for row counts not matching? Won't this record the wrong one?

apps/pipeline/lib/pipeline/writer/s3_writer/compaction.ex
```diff
@@ -81,6 +82,7 @@ defmodule Pipeline.Writer.S3Writer.Compaction do
     orc_table = options[:orc_table]
     json_table = options[:json_table]
     compact_table = "#{orc_table}_compact"
+    TelemetryEventHelper.add_dataset_record_event_count(orc_table, new_count)
```
LtChae, line 85: I'd suggest doing this last in this function so that it won't record numbers for a failed compaction

apps/pipeline/lib/pipeline/writer/s3_writer/compaction.ex
```diff
@@ -55,9 +56,17 @@ defmodule Pipeline.Writer.S3Writer.Compaction do
     json_table = options[:json_table]
 
     case PrestigeHelper.execute_query("select count(1) from #{json_table}") do
-      {:ok, %{rows: [[0]]}} -> true
-      {:error, _} -> true
-      _ -> false
+      {:ok, %{rows: [[0]]}} ->
+        options[:orc_table]
+        |> TelemetryEventHelper.add_dataset_record_event_count(0)
```
LtChae, line 61: We don't care about the records in the json table, right? Especially since there aren't any in this case. We need the record from the main table known here as `:orc_table`

PR #891, opened 2020-08-26
Harvesting cadence

kcarra:


apps/andi/lib/andi/harvest/data_json_dataset_mapper.ex
```diff
@@ -49,14 +50,14 @@ defmodule Andi.Harvest.DataJsonDatasetMapper do
...
-        "modifiedDate" => data_json_dataset["modified"],
+        "modifiedDate" => modified_date(data_json_dataset["modified"]),
         "orgTitle" => org.orgTitle,
         "parentDataset" => data_json_dataset["isPartOf"],
         "referenceUrls" => data_json_dataset["references"],
         "rights" => data_json_dataset["rights"],
         "spatial" => data_json_dataset["spatial"]
       },
-      "id" => UUID.uuid5(@scos_data_json_seed, data_json_dataset["identifier"]),
+      "id" => generate_scos_dataset_id(data_json_dataset["identifier"]),
```
bbrewer-forge, line 60: noice

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -5,19 +5,25 @@ defmodule Andi.Harvest.Harvester do
...
 
-  def start_harvesting(org) do
+  def start_harvesting(%Organization{} = org) do
     url = org.dataJsonUrl
 
     with {:ok, data_json} <- get_data_json(url),
          {:ok, decoded_data_json} <- Jason.decode(data_json),
          datasets <- map_data_json_to_dataset(decoded_data_json, org),
          harvested_datasets <- map_data_json_to_harvested_dataset(decoded_data_json, org),
+         :ok <- remove_old_harvested_datasets(harvested_datasets, org.id),
```
bbrewer-forge, line 26: I, Ben Brewer, wholly approve of doing this as a separate step like this

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -48,13 +66,43 @@ defmodule Andi.Harvest.Harvester do
 
   def dataset_update(datasets) do
     Enum.each(datasets, fn dataset ->
-      Brook.Event.send(instance_name(), dataset_update(), :andi, dataset)
+      case Organizations.get_harvested_dataset(dataset.id) do
```
bbrewer-forge, line 69: wait, what's the difference between this and the function below?

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -27,6 +33,18 @@ defmodule Andi.Harvest.Harvester do
     end
   end
 
+  def start_harvesting() do
+    case OrgStore.get_all() do
```
cameronmarsh, line 37: This would be faster if we used a query instead. Idk how many orgs will be in the system but it's probably better to handle it now. 

If you want to keep it this way, I'm pretty sure OrgStore can return `{:ok, nil}` when there are no orgs in the system. Even though we probably won't run into that case, it would fail here on the Enum calls

apps/andi/lib/andi/harvest/harvester.ex
```diff
@@ -48,13 +66,43 @@ defmodule Andi.Harvest.Harvester do
...
+        %{include: false} -> Logger.info("Skipping dataset update for harvested dataset #{harvested_dataset["datasetId"]}")
+        _ -> Brook.Event.send(instance_name(), dataset_harvest_end(), :andi, harvested_dataset)
+      end
+    end)
+  end
+
+  defp remove_old_harvested_datasets(new_datasets, org_id) do
+    current_datasets = Organizations.get_all_harvested_datasets(org_id)
+
+    Enum.each(current_datasets, fn existing_dataset ->
```
cameronmarsh, line 88: Would it be easier to have a function that overwrites the current harvested dataset instead of deleting the old one? That may not work for the ViewState as easily, but it would clean this up a little. If not, maybe break this block up into some helper functions?

apps/andi/lib/andi/input_schemas/organization.ex
```diff
@@ -7,7 +7,7 @@ defmodule Andi.InputSchemas.Organization do
 
   alias Andi.InputSchemas.StructTools
 
-  @primary_key {:id, Ecto.UUID, autogenerate: false}
+  @primary_key {:id, :string, autogenerate: false}
```
cameronmarsh, line 10: I thought we were keeping the UUID and adding validation to the changeset function

apps/andi/lib/andi/input_schemas/organizations.ex
```diff
@@ -30,6 +30,13 @@ defmodule Andi.InputSchemas.Organizations do
     Repo.all(query)
   end
 
+  def delete_harvested_dataset(dataset_id) do
+    case get_harvested_dataset(dataset_id) do
+      %{id: id} -> Repo.delete(%HarvestedDatasets{id: id})
+      _ -> Logger.info("Unable to delete dataset: #{dataset_id} from harvested datasets")
```
cameronmarsh, line 36: You could also just check for `nil` and have the default case be a harvested dataset that you can delete instead of making a new one to delete

apps/andi/lib/andi/harvest/harvester.ex
kcarra: dataset_update sends the update dataset event with the smart city dataset and harvested dataset update sends the harvest end event and updated the harvested dataset table.

PR #898, opened 2020-08-31
Add table fields to harvested datasets table

cameronmarsh:


apps/andi/lib/andi_web/live/edit_organization_live_view.ex
```diff
@@ -225,22 +226,33 @@ defmodule AndiWeb.EditOrganizationLiveView do
...
     order_dir =
       case socket.assigns.order do
-        "asc" -> "desc"
+        %{^field => "asc"} -> "desc"
         _ -> "asc"
       end
 
-    sorted_datasets = socket.assigns.harvested_datasets |> Enum.reverse()
-    {:noreply, assign(socket, harvested_datasets: sorted_datasets, order: order_dir)}
+    sorted_datasets = sort_by_dir(socket.assigns.harvested_datasets, field, order_dir)
```
bbrewer-forge, line 236: I'd love to see this moved back into ecto query as well, but I digress...

cameronmarsh: Yeah there is some opportunities to make another macro for table live view pages that I will probably add soon, so I can do that then

PR #901, opened 2020-09-01
Track api

sukaant:


apps/discovery_api/lib/discovery_api_web/router.ex
```diff
@@ -36,20 +36,24 @@ defmodule DiscoveryApiWeb.Router do
     plug(DiscoveryApiWeb.Plugs.NoStore)
   end
 
+  pipeline :telemetry do
+    plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
+  end
+
   scope "/api/v2", DiscoveryApiWeb do
-    pipe_through([:reject_cookies_from_ajax, :verify_token, :add_user_details, :global_headers])
+    pipe_through([:reject_cookies_from_ajax, :verify_token, :add_user_details, :global_headers, :telemetry])
```
LtChae, line 44: See if you can add this to a pipeline that everything passes through

apps/andi/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "0.42.1",
+      version: "0.42.3",
```
cameronmarsh, line 7: nbd, but you bumped this two versions

apps/telemetry_event/lib/telemetry_event/helper/add_init_server.ex
```diff
@@ -0,0 +1,18 @@
+defmodule TelemetryEvent.Helper.AddInitServer do
+  @moduledoc false
+  alias TelemetryEvent.Helper.TelemetryEventHelper
+
+  def add_metrics_prometheus(child, app_name) do
```
LtChae, line 5: This probably shouldn't be part of the `AddInitServer` module

apps/andi/mix.exs
sukaant: Just noticed, Thanks!

apps/telemetry_event/lib/telemetry_event/helper/add_metrics.ex
```diff
@@ -2,14 +2,41 @@ defmodule TelemetryEvent.Helper.AddMetrics do
...
+        [
+          [
+            metric_name: "phoenix.endpoint.stop.duration",
+            tags: [:request_path, :method],
+            tag_values: fn %{conn: conn} ->
+              %{request_path: Map.get(conn, :request_path), method: Map.get(conn, :method)}
+            end,
+            metric_type: :distribution,
+            unit: {:native, :millisecond},
+            reporter_options: [buckets: [0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1]]
```
LtChae, line 33: New metric type here. This should make a histogram of some sort I think? Make sure it's outputting in a way we can consume in dev.

LtChae (approved):

PR #914, opened 2020-09-03
SPIKE: a simple setup for getting user info using an auth code

bbrewer-forge:


apps/andi/config/integration.exs
```diff
@@ -29,7 +29,23 @@ config :andi, Andi.Repo,
 config :andi, AndiWeb.Endpoint,
   http: [port: 4000],
   server: true,
-  check_origin: false
+  check_origin: false,
+  https: [
+    port: 4443,
+    otp_app: :andi,
+    keyfile: "priv/key.pem",
```
bbrewer-forge, line 36: Had to generate some self-signed ones for local testing

apps/andi/lib/andi_web/controllers/user_from_auth.ex
```diff
@@ -0,0 +1,62 @@
+defmodule UserFromAuth do
```
bbrewer-forge, line 1: 100% boilerplate from Auth0's elixir article

bbrewer-forge:
Closing, but this will still be here for reference

PR #924, opened 2020-09-08
Fix elsa configuration

cameronmarsh:


apps/andi/lib/andi/application.ex
```diff
@@ -5,6 +5,21 @@ defmodule Andi.Application do
...
+        true ->
+          [{"localhost", 9092}]
+
+        false ->
+          kafka_brokers
+          |> String.split(",")
+          |> Enum.map(&String.trim/1)
+          |> Enum.map(fn entry -> String.split(entry, ":") end)
+          |> Enum.map(fn [host, port] -> {String.to_atom(host), String.to_integer(port)} end)
+      end
```
LtChae, line 21: Why was this removed from the config?

LtChae (approved):

apps/andi/lib/andi/application.ex
cameronmarsh: yeah that was a mistake

PR #927, opened 2020-09-09
Include button

kcarra:


cameronmarsh (approved):

dunson062786 (approved):

apps/andi/assets/css/organizations.scss
```diff
@@ -113,6 +113,20 @@
   max-width: 30rem;
 }
 
+.organizations-table__checkbox {
+  padding-left: 35px;
```
bbrewer-forge, line 117: have we been using a lot of px lately? I would prefer we stick to rems for everything but borders if possible, unless the px usage is already widespread

PR #930, opened 2020-09-11
Handle dataset count

sukaant:


apps/discovery_api/lib/discovery_api/event_handler.ex
```diff
@@ -137,6 +138,8 @@ defmodule DiscoveryApi.EventHandler do
   end
 
   defp add_dataset_count() do
+    Process.sleep(20_000)
```
LtChae, line 141: This works, though I'd suggest putting a comment here to explain why. The delay might be too pessimistic, I'd think five seconds would be enough.

LtChae (approved):

PR #951, opened 2020-09-21
Join metrics

sukaant:


apps/andi/lib/andi/dataset_cache.ex
```diff
@@ -76,10 +76,9 @@ defmodule Andi.DatasetCache do
     [
       dataset_id: dataset[:id],
       dataset_title: dataset[:business][:dataTitle],
-      system_name: "#{dataset[:id]}#{dataset[:data_name]}#{dataset[:org_name]}",
+      system_name: "#{dataset[:technical][:orgName]}__#{dataset[:technical][:dataName]}",
```
LtChae, line 79: Datasets have a systemName field already. Use that if you can before recalculating it

LtChae (approved):

PR #954, opened 2020-09-22
S3 step

jmorris-pillar:


sukaant (approved):

apps/reaper/lib/reaper/data_extract/processor.ex
```diff
@@ -69,7 +69,11 @@ defmodule Reaper.DataExtract.Processor do
   end
 
   defp create_producer_stage(%SmartCity.Dataset{technical: %{extractSteps: steps}} = dataset) do
-    ExtractStep.execute_extract_steps(dataset, steps)
+    %{output_file: output_file} = ExtractStep.execute_extract_steps(dataset, steps)
```
bbrewer-forge, line 72: I'm fine with this convention as that seems like a legit end point for an extract

bbrewer-forge (approved):

PR #955, opened 2020-09-22
Improve streaming performance

bbrewer-forge:


apps/discovery_streams/mix.exs
```diff
@@ -74,7 +74,9 @@ defmodule DiscoveryStreams.Mixfile do
       {:tasks, in_umbrella: true, only: :dev},
       {:decimal, "~> 1.0"},
       {:tzdata, "~> 1.0"},
-      {:web, in_umbrella: true}
+      {:web, in_umbrella: true},
+      {:combinatorics, "~> 0.1.0"},
```
jmorris-pillar, line 78: do you need this outside of testing?

apps/discovery_streams/test/integration/performance/cve_test.exs
```diff
@@ -0,0 +1,337 @@
...
+      inputs: scenarios,
+      before_scenario: fn %SetupConfig{
+                            messages: messages,
+                            prefetch_count: prefetch_count,
+                            prefetch_bytes: prefetch_bytes,
+                            min_bytes: min_bytes,
+                            max_bytes: max_bytes,
+                            max_wait_time: max_wait_time
+                          } = _parameters_from_inputs ->
+        # existing_topic_config = Application.get_env(:discovery_streams, :topic_subscriber_config)
```
jmorris-pillar, line 115: whats all this for?

jmorris-pillar (approved):

apps/discovery_streams/mix.exs
bbrewer-forge: no, we need to re-mark that

apps/discovery_streams/test/integration/performance/cve_test.exs
bbrewer-forge: this is usually where we'd tune the kafka consumer config, per the scenario. So we'd tweak the prefetch, etc. However, that's hardcoded inside of `Definition.Kafka` for the moment so we have it commented out

apps/discovery_streams/lib/discovery-streams/stream.ex
```diff
@@ -47,6 +47,7 @@ defmodule DiscoveryStreams.Stream do
       {:error, reason} ->
         {:stop, reason, state}
     end
+
```
bbrewer-forge, line 50: ewps

apps/discovery_streams/lib/discovery-streams/stream/source_handler.ex
```diff
@@ -33,6 +33,7 @@ defmodule DiscoveryStreams.Stream.SourceHandler do
   end
 
   def handle_batch(batch, context) do
+
```
bbrewer-forge, line 36: ewps

LtChae (approved):

apps/discovery_streams/lib/discovery-streams/stream/source_handler.ex
```diff
@@ -14,20 +14,11 @@ defmodule DiscoveryStreams.Stream.SourceHandler do
   getter(:dlq, default: Dlq)
 
   def handle_message(message, context) do
-    dataset_id = context.dataset_id
+    system_name = context.assigns.system_name
```
bbrewer-forge, line 17: this is the real meat right here in this function

apps/discovery_streams/test/integration/discovery_streams_test.exs
```diff
@@ -69,42 +70,55 @@ defmodule DiscoveryStreams.DiscoveryStreamsTest do
...
+      TopicHelper.topic_name(dataset.id),
+      [create_message(%{foo: "bar"}, topic: dataset.id)],
       parition: 0
     )
 
     assert_push("update", %{"foo" => "bar"}, 15_000)
 
-    Brook.Test.send(@instance, dataset_delete(), :author, dataset1)
+    dataset =
+      if update_path do
```
bbrewer-forge, line 94: I tried

apps/performance/test/performance_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule PerformanceTest do
```
bbrewer-forge, line 1: eh

apps/discovery_streams/config/config.exs
```diff
@@ -15,7 +15,16 @@ config :logger,
...
-  topic_prefix: "transformed-"
+  topic_prefix: "transformed-",
+  topic_subscriber_config: [
+    begin_offset: :latest,
+    offset_reset_policy: :reset_to_latest,
+    max_bytes: 10_000_000,
+    min_bytes: 0,
+    max_wait_time: 10_000,
+    prefetch_count: 0,
+    prefetch_bytes: 100_000_000
```
jmorris-pillar, line 26: have we tested how these kinds of things perform for datasets which trickle a few bytes in a minute?

apps/discovery_streams/lib/discovery-streams/stream/source_handler.ex
```diff
@@ -38,7 +29,11 @@ defmodule DiscoveryStreams.Stream.SourceHandler do
     :ok
   end
 
-  def send_to_dlq(dead_letters, _context) do
+  defp log_message(message, context) do
```
jmorris-pillar, line 32: As long as this works good i say dont change it but i wonder, since functions are executed inside out, if its not better to wrap the log statement in a config level check instead of doing logger.debug.  Maybe this works diff since its a macro /shrug

apps/discovery_streams/lib/discovery-streams/stream/source_handler.ex
```diff
@@ -48,11 +43,6 @@ defmodule DiscoveryStreams.Stream.SourceHandler do
     message["payload"]
   end
 
-  defp log_message(message) do
```
jmorris-pillar, line 51: lol whoops, i cant believe i was doing this twice.  I blame whoever reviewed my code

apps/discovery_streams/config/config.exs
bbrewer-forge: For intermittent, small chunks of data, we have not. There is not much that can be done here to help prefetch with that. WRT to max bytes and wait time, the min_bytes of 0 has some special kafka magic that will still forward on trickles of data. It's like an internal threshold of 100ms or something. I'll go dig it up. But yeah, as long as `min_bytes` is set to 0, we don't end up having to hit the max wait time for trickles

apps/discovery_streams/test/integration/discovery_streams_test.exs
jmorris-pillar: why do u need this?

jmorris-pillar (approved):

PR #968, opened 2020-10-01
Adding Handle Info

sukaant:


apps/estuary/lib/estuary_web/live/event_live_view.ex
```diff
@@ -58,6 +58,19 @@ defmodule EstuaryWeb.EventLiveView do
     {:noreply, push_patch(socket, to: Routes.live_path(socket, __MODULE__, search_params))}
   end
 
+  def handle_info(%{search: params}, socket) do
```
bbrewer-forge, line 61: Did you remove the event subscription part of this in an earlier PR?

bbrewer-forge (approved):
Looks good assuming this is not wired up to do a search on every brook event still.

dunson062786 (approved):

apps/estuary/lib/estuary_web/live/event_live_view.ex
sukaant: No

PR #969, opened 2020-10-01
Using source and destination in Valkyrie

bbrewer-forge:


bbrewer-forge:
Shelving this for the moment until we can get perf figured out.

PR #971, opened 2020-10-01
Create org

kcarra:


apps/andi/lib/andi/input_schemas/organizations.ex
```diff
@@ -91,9 +105,17 @@ defmodule Andi.InputSchemas.Organizations do
 
   def is_unique?(id, org_name) do
     from(org in Andi.InputSchemas.Organization,
-      where: org.orgName == ^org_name and org.id != ^id
+      where: org.orgName == ^org_name or org.id == ^id
```
cameronmarsh, line 108: Am I insane, becuase it seems like this would be false every time?

apps/andi/lib/andi/input_schemas/organizations.ex
```diff
@@ -91,9 +105,17 @@ defmodule Andi.InputSchemas.Organizations do
...
     )
     |> Repo.all()
     |> Enum.empty?()
   end
+
+  def org_title_to_org_name(org_title) do
```
cameronmarsh, line 114: it might be nice to DRY this up a bit because the datasets module does the same thing

apps/andi/lib/andi/input_schemas/organizations.ex
kcarra: This works or making sure the id and org name are new when you create a new org. The only thing I just realized is it messes up the save when you are editing the org so I might have to put some thing into that

apps/andi/test/integration/andi_web/live/edit_organization_live_view_test.exs
```diff
@@ -107,6 +107,13 @@ defmodule AndiWeb.EditOrganizationLiveViewTest do
 
       where(title: ["", "!@#$%"])
     end
+
+    test "user has the ability to add an organization on the organizations page", %{conn: conn} do
```
cameronmarsh, line 111: this seems like the wrong test file, I would maybe break up the tests into a new module testing only the `organization_live_view` stuff

cameronmarsh: Also, are there any tests around the uniqueness of the orgName in this test file? I feel like that error will always come up now because of the `or` in the uniqueness check above. So it might be worth putting one in here

kcarra: This is no `organization_live_view_test` so i figured it would be ok to put this here

cameronmarsh (approved):

PR #973, opened 2020-10-02
Require authentication in andi

cameronmarsh:


apps/andi/lib/andi_web/auth/error_handler.ex
```diff
@@ -0,0 +1,19 @@
+defmodule AndiWeb.Auth.ErrorHandler do
+  @moduledoc false
+  @behaviour Guardian.Plug.ErrorHandler
+
+  require Logger
+
+  @impl Guardian.Plug.ErrorHandler
+  def auth_error(conn, {:unauthenticated, "https://andi.smartcolumbusos.com/roles"}, _opts) do
```
LtChae, line 8: Any chance we could keep metrics of successful vs failed logins?

apps/auth/README.md
```diff
@@ -0,0 +1,19 @@
+# Auth
+
+## Installation
```
LtChae, line 3: We should probably remove the boilerplate

apps/auth/lib/auth.ex
```diff
@@ -0,0 +1,3 @@
+defmodule Auth do
```
LtChae, line 1: What does this do?

LtChae (approved):

apps/auth/README.md
cameronmarsh: okay

apps/auth/lib/auth.ex
cameronmarsh: nothing

PR #975, opened 2020-10-06
Instance name

sukaant:


apps/discovery_streams/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule DiscoveryStreams.Mixfile do
   def project do
     [
       app: :discovery_streams,
-      version: "2.10.0",
+      version: "2.10.1",
```
bbrewer-forge, line 7: There are definitely some cases in this app that I know of where `:discovery_streams` is hardcoded in the `Brook.Event.send` and other functions. It feels like we should at least take care of those too.

apps/odo/runtime.exs
```diff
@@ -32,7 +32,6 @@ config :ex_aws,
   region: System.get_env("AWS_REGION") || "us-west-2"
 
 config :odo, :brook,
-  instance: :odo_brook,
```
bbrewer-forge, line 35: are we positive brook still is able to receive events with this change? Looking at `application.ex` it's unclear how it would work.

bbrewer-forge (changes requested):
Please also get the places where we hardcoded the instance name (disco streams is one of them), and make sure that Odo still works. It should be failing the e2e test right now, at a minimum

bbrewer-forge (approved):

PR #999, opened 2020-10-13
User owned datasets

LtChae:
This includes the initial functionality of assigning a dataset to the user who creates it, as well as a migration function to initially assign all the datasets to the current curator (as a one-time manual process)

### DIFF
--- a/apps/andi/lib/andi/input_schemas/datasets/dataset.ex
+++ b/apps/andi/lib/andi/input_schemas/datasets/dataset.ex
@@ -5,6 +5,7 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
   use Ecto.Schema
   import Ecto.Changeset
 
+  alias Andi.Schemas.User
   alias Andi.InputSchemas.Datasets
   alias Andi.InputSchemas.Datasets.Business
   alias Andi.InputSchemas.Datasets.Technical
@@ -16,6 +17,7 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
     field(:dlq_message, :map)
     field(:ingestedTime, :utc_datetime, default: nil)
     field(:version, :string)
+    belongs_to(:owner, User, type: Ecto.UUID, foreign_key: :owner_id)
     has_many(:data_dictionaries, DataDictionary)
     has_one(:business, Business, on_replace: :update)
     has_one(:technical, Technical, on_replace: :update)
@@ -32,6 +34,14 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
     |> cast(changes, @cast_fields)
     |> cast_assoc(:technical, with: &Technical.changeset/2)
     |> cast_assoc(:business, with: &Business.changeset/2)
+  end
+
+  def changeset_for_draft(dataset, %{owner: owner} = changes) do
+    dataset
+    |> cast(changes, @cast_fields)
+    |> put_assoc(:owner, owner)
+    |> cast_assoc(:technical, with: &Technical.changeset_for_draft/2)
+    |> cast_assoc(:business, with: &Business.changeset_for_draft/2)
   end
 
   def changeset_for_draft(dataset, changes) do
--- a/apps/andi/lib/andi/migration/owner.ex
+++ b/apps/andi/lib/andi/migration/owner.ex
@@ -1,0 +1,25 @@
+defmodule Andi.Migration.Owner do
+  @moduledoc false
+  alias Andi.Repo
+  alias Andi.InputSchemas.Datasets.Dataset
+  alias Andi.InputSchemas.Datasets
+
+  def transfer_all_datasets_to_owner(owner) do
+    Datasets.get_all()
+    |> Enum.map(fn dataset -> Repo.preload(dataset, [:owner]) end)
+    |> Enum.each(fn dataset -> update_owner(dataset, owner) end)
+  end
+
+  defp update_owner(dataset, owner) do
+    new_changeset =
+      Dataset.changeset_for_draft(
+        dataset,
+        %{
+          owner: owner
+        }
+      )
+
+    {:ok, new_dataset} = Datasets.save(new_changeset)
+    new_dataset
+  end
+end
--- a/apps/andi/lib/andi/schemas/user.ex
+++ b/apps/andi/lib/andi/schemas/user.ex
@@ -21,11 +21,15 @@ defmodule Andi.Schemas.User do
   end
 
   def create_or_update(subject_id, changes \\ %{}) do
-    case Repo.get_by(__MODULE__, subject_id: subject_id) do
+    case get_by_subject_id(subject_id) do
       nil -> %__MODULE__{subject_id: subject_id}
       user -> user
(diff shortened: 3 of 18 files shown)


apps/andi/lib/andi/schemas/user.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Andi.Schemas.User do
```
LtChae, line 1: Not sure about this thing's location, but its definitely not an input schema.

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
```diff
@@ -34,6 +36,14 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
     |> cast_assoc(:business, with: &Business.changeset/2)
   end
 
+  def changeset_for_draft(dataset, %{owner: owner} = changes) do
```
LtChae, line 39: Not in love with having two heads for this, but I can't see a way around having one way put an owner in and the other not.

apps/andi/lib/andi_web/router.ex
```diff
@@ -32,7 +32,7 @@ defmodule AndiWeb.Router do
     pipe_through [:browser, :auth]
 
     get "/", Redirect, to: "/datasets"
-    live "/datasets", DatasetLiveView, layout: {AndiWeb.LayoutView, :root}
+    live "/datasets", DatasetLiveView, layout: {AndiWeb.LayoutView, :root}, session: {AndiWeb.Auth.TokenHandler.Plug, :current_resource, []}
```
bbrewer-forge, line 35: Is the session thing here even necessary since you are putting the userid in the session?

apps/andi/lib/andi_web/controllers/auth_controller.ex
```diff
@@ -16,8 +16,11 @@ defmodule AndiWeb.AuthController do
   end
 
   def callback(%{assigns: %{ueberauth_auth: auth}} = conn, _params) do
+    {:ok, user} = Andi.Schemas.User.create_or_update(auth.uid, %{email: auth.info.email})
+
     conn
     |> TokenHandler.put_session_token(auth.credentials.token)
+    |> Plug.Conn.put_session(:user_id, user.id)
```
bbrewer-forge, line 23: I would kind of prefer we do this through the resource loader, as that is a fairly standard pattern. As in, if you want user_id around, you can put it in the session with the `resource_from_claims` call

bbrewer-forge (changes requested):
I'd like to pair with you and do this a different way. As in, the token should be single source of truth, and adding a separate session variable with its own lifecycle could be a problem

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
```diff
@@ -34,6 +36,14 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
     |> cast_assoc(:business, with: &Business.changeset/2)
```
cameronmarsh, line 36: do we need to add a ` |> foreign_key_constraint(:owner_id)` here and in the changeset_for_draft function below? I think that would force a dataset to have a user, so idk if we want that or not

apps/andi/lib/andi/migration/owner.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Andi.Migration.Owner do
```
cameronmarsh, line 1: nice, I don't think we've used this pattern for past migrations but we should have

apps/andi/test/support/auth_helper.ex
```diff
@@ -34,6 +34,7 @@ defmodule Andi.Test.AuthHelper do
       |> Plug.Session.call(signing_opts)
       |> Plug.Conn.fetch_session()
       |> TokenHandler.put_session_token(jwt)
+      |> Plug.Conn.put_session(:user_id, "TEST")
```
bbrewer-forge, line 37: This here in particular highlights why this approach is not great. Any use could set their user_id to ANY value and we have no mechanisms in place to validate that. If it all stems from the JWT, then we can verify the signature, expiry, etc.

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -56,10 +56,11 @@ defmodule AndiWeb.DatasetLiveView do
     """
   end
 
-  def mount(_params, _session, socket) do
+  def mount(_params, %{"user_id" => user_id} = _session, socket) do
```
cameronmarsh, line 59: you can probably remove the `_session` here

apps/andi/test/support/auth_helper.ex
bbrewer-forge: LiveView does make this tricky with its use of session, so we may have to lock that down as well.

apps/andi/lib/andi_web/controllers/auth_controller.ex
cameronmarsh, line 23: Would it make sense to just put the `user` on the conn instead of the id so we wouldn't have to look the user up again on mount? Or is there some concern about that

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
cameronmarsh: nvm, just saw your comment in slack

apps/andi/lib/andi_web/controllers/auth_controller.ex
LtChae: The user object could easily get out of date. Looking it up should keep things fresh

apps/andi/priv/repo/migrations/20201012134252_add_user_id_to_dataset.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Andi.Repo.Migrations.AddUserIdToDataset do
+  use Ecto.Migration
+
+  def change do
+    create table(:users, primary_key: false) do
+      add :id, :uuid, null: false, primary_key: true
+      add :subject_id, :string, null: false
```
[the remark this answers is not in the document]
LtChae: The id is its identifier in our system, the subject_id is the identifier in Auth0. This is a copy of the pattern in Discovery API. Not sure if there's a ton of value or not, but it seems to be our convention and it doesn't hurt anything.

apps/andi/lib/andi_web/live/edit_live_view.ex
```diff
@@ -26,6 +26,7 @@ defmodule AndiWeb.EditLiveView do
         <% [business] = inputs_for(f, :business) %>
         <% [technical] = inputs_for(f, :technical) %>
         <%= hidden_input(f, :id) %>
+        <%= hidden_input(f, :owner_id) %>
```
bbrewer-forge, line 29: This is fine, but now it's got me thinking of all the different ways someone could just hack in values for this, the dataset_id, etc. We'll A&C that

bbrewer-forge (approved):

PR #1002, opened 2020-10-15
Owned dataset view

LtChae:
+ Public users cannot access unowned datasets and organizations
+ Curators can still access everything


sukaant (approved):

bbrewer-forge (approved):
kewl

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -12,10 +12,12 @@ defmodule AndiWeb.DatasetLiveView do
     <div class="datasets-view">
       <div class="page-header">
         <a href="/datasets">Dataset Ingestion Interface</a>
+        <%= if assigns.is_curator do %>
```
cameronmarsh, line 15: I think it's liveview convention to use `@is_curator`

cameronmarsh (approved):

PR #1005, opened 2020-10-16
Http extract step

cameronmarsh:


LtChae (approved):
Looks workable, though the amount of code needed to implement just the http step in schemas is worrying to me. We need to be able to make new steps fairly easily if this is how we add new ingestion functionality going forward, to say nothing of converting the steps we already have.

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -111,7 +113,7 @@ defmodule Andi.InputSchemas.Datasets do
       |> StructTools.to_map()
       |> Map.merge(form_changes)
 
-    new_changes = %{technical: technical_changes, business: business_changes, id: dataset_id} |> StructTools.to_map()
+    new_changes = %{technical: technical_changes, business: business_changes, id: dataset_id}
```
jessie-morris, line 116: Why no to_map anymore?

cameronmarsh:
In response to Tim's comment, unfortunately a lot of this is required to make new Ecto tables and liveview components, so there's not a lot we can do there. However, some of it is only temporary to allow both the sourceUrl and extract step forms to be shown at the same time. For future extract step cards at least, there are patterns that can be extracted out and shared (like the form_section file) that will result in less code.

apps/andi/lib/andi/input_schemas/datasets.ex
cameronmarsh: They are all maps already, this was doing nothing

PR #1011, opened 2020-10-19
Generic getter implementation (2)

msomji:


LtChae:
I believe the version checker issue is due to this being a fork pr.

apps/properties/lib/properties.ex
```diff
@@ -72,24 +72,40 @@ defmodule Properties do
...
-    case required do
-      true ->
-        quote do
-          defp unquote(key)() do
-            Application.get_env(@properties_otp_app, unquote(module), [])
-            |> Keyword.fetch!(unquote(key))
-          end
-        end
+    getter(generic, required, module, key, default)
+  end
```
bbrewer-forge, line 80: I like this. Looks good to me. Another pattern I sometimes do for opts like this is (NOTE: doesn't deal with defaults - usually I'll just Map.merge those in when I transform it):
```
defmacro getter(key, opts \\ []) do
  opts_as_map = Map.new(opts)
  module = __CALLER__.module

  getter(module, key, opts_as_map)
end

defp getter(module, key, %{generic: true, required: true}) do
...
defp getter(module, key, %{generic: true, required: false}) do
...
defp getter(module, key, %{generic: false, required: true}) do
...
defp getter(module, key, _opts) do
```

Not appreciably different, but leaves you open for future options without having to mess with the call signature too much. At the end of the day, it's not a huge deal either way, since it's a private function anyways

bbrewer-forge (approved):

PR #1028, opened 2020-10-22
Change ownership

kcarra:


apps/andi/README.md
```diff
@@ -18,8 +18,7 @@ Interactions with Redis are abstracted with `smartcitiesdata.smart_city*` functi
 - Install dependencies with `mix deps.get`
 - `cd assets` and `npm i`
 - `MIX_ENV=integration mix docker.start`
-- `MIX_ENV=integration mix ecto.create && MIX_ENV=integration mix ecto.migrate`
-- Start Phoenix endpoint locally with `AUTH0_CLIENT_SECRET="<auth_client_secret>" MIX_ENV=integration iex -S mix phx.server`
+- Start Phoenix endpoint locally with `AUTH0_CLIENT_SECRET="<auth_client_secret>" MIX_ENV=integration iex -S mix start`
```
cameronmarsh, line 21: the reason I changed this was because this command doesn't start andi interactively and is less clear about the ecto setup. I feel like it would be helpful to give people more info but I might be wrong

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -114,12 +114,18 @@ defmodule Andi.InputSchemas.Datasets do
       |> StructTools.to_map()
       |> Map.merge(form_changes)
 
-    new_changes = %{technical: technical_changes, business: business_changes, id: dataset_id}
+    user_id =
```
cameronmarsh, line 117: I feel like this function could use some refactoring. I think it would be cleaner to make a separate function head that pattern matches out the `user_id` and break this function up

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -227,6 +233,10 @@ defmodule Andi.InputSchemas.Datasets do
     |> String.downcase()
   end
 
+  defp extract_user_id(nil, %{userId: userId}), do: userId
```
cameronmarsh, line 236: should this be `user_id` instead of `userId`?

apps/andi/lib/andi_web/input_schemas/metadata_form_schema.ex
```diff
@@ -20,6 +20,7 @@ defmodule AndiWeb.InputSchemas.MetadataFormSchema do
     field(:dataName, :string)
     field(:dataTitle, :string)
     field(:datasetId, :string)
+    field(:userId, :string)
```
cameronmarsh, line 23: I would rename this to `user_id` to stay consistent with the dataset changeset association

cameronmarsh (approved):

apps/andi/README.md
kcarra: I don't really have a preference on this. I can change it back

PR #1030, opened 2020-10-23
Replacing Application.get_env with getter

sukaant:


bbrewer-forge (commented):
Looks good, but I'll let Jessie take a crack too

jessie-morris (approved):

PR #1031, opened 2020-10-23
Extract Step SFTP

### LINKED ISSUE #1023: Extract Step SFTP
As a data curator
I want to be able to ingest data through sftp
So I can capture cool data.

Create a new extract step definition and implementation to pull data using sftp using existing data slurpers etc

AC:

- I can ingest data from an sftp data source
- Fields such as the url, username, and password are templatable from previous assigns steps.
- This step produces an :output_file following the pattern of http and s3 steps.

msomji:
closes #1023


bbrewer-forge (approved):
Nice! Looks good to me

PR #1034, opened 2020-10-26
GitHub actions

bbrewer-forge:


kcarra:
:pray: :pray:

scripts/wdc-test.sh
```diff
@@ -1,17 +0,0 @@
-#!/usr/bin/env bash
-
-set -e
-
-source ./scripts/lib_common.sh
```
bbrewer-forge, line 5: will restore

.github/workflows/andi.yml
```diff
@@ -0,0 +1,64 @@
+name: Andi
+on:
+  push:
+    branches:
+      - master
+    paths:
+    - 'apps/andi/**'
```
bbrewer-forge, line 7: pros: allows us to truly skip steps (not spin up a vm for 45 seconds to check if it changed)
cons: lots of copy and paste for library based apps we want to check

LtChae: Adding new lines here isn't too horrible a cost to pay in my opinion

LtChae (approved):

PR #1076, opened 2020-11-10
Extract step fixes

cameronmarsh:


apps/andi/lib/andi/input_schemas/datasets/extract_steps/extract_date_step.ex
```diff
@@ -71,18 +73,34 @@ defmodule Andi.InputSchemas.Datasets.ExtractDateStep do
...
   defp validate_time_unit(changeset), do: changeset
 
+  defp validate_delta_change(%{changes: %{deltaTimeUnit: deltaTimeUnit}} = changeset) when deltaTimeUnit not in [nil, ""] do
+    case get_change(changeset, :deltaTimeValue) in [nil, ""] do
+      true -> add_error(changeset, :deltaTimeValue, "must be set when deltaTimeUnit is set")
+      false -> changeset
+    end
+  end
+
+  defp validate_delta_change(%{changes: %{deltaTimeValue: deltaTimeValue}} = changeset) when deltaTimeValue not in [nil, ""] do
```
bbrewer-forge, line 95: what's the difference between this and the above function?

cameronmarsh: the one above checks for `deltaTimeUnit` in the changeset and verifies that `deltaTimeValue` is also set. We are basically validating that if one is set, then the other has to be too, even though neither of them are required

bbrewer-forge: 🤦‍♂️

bbrewer-forge (approved):

PR #1077, opened 2020-11-11
submission status in dataset live view

bbrewer-forge:


apps/andi/lib/andi_web/live/dataset_live_view/table.ex
```diff
@@ -69,7 +69,7 @@ defmodule AndiWeb.DatasetLiveView.Table do
     end
   end
 
-  defp ingest_status(_), do: ""
+  defp ingest_status(%{""}), do:
```
bbrewer-forge, line 72: oops....

apps/andi/test/integration/test_helper.exs
```diff
@@ -1,2 +1,3 @@
-Divo.Suite.start()
+Application.ensure_all_started(:andi)
+# Divo.Suite.start()
```
cameronmarsh, line 2: 🍕

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -135,6 +135,12 @@ defmodule Andi.InputSchemas.Datasets do
     update(from_dataset, %{ingestedTime: iso_ingested_time})
   end
 
+  def update_submission_status(dataset_id, status) do
+    from_dataset = get(dataset_id)
+
+    update(from_dataset, %{submission_status: status})
```
bbrewer-forge, line 141: Some things like this could be optimized to not even do a `get` since all they need is the dataset id. However, it's not killing us AFAIK

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -158,24 +160,43 @@ defmodule AndiWeb.DatasetLiveView do
...
+      "status" => status(dataset),
+      "status_sort" => status_sort(dataset)
     }
   end
 
+  defp status(%{ingestedTime: nil, submission_status: status}), do: String.capitalize(Atom.to_string(status))
+  defp status(dataset), do: ingest_status(dataset)
+
+  defp status_sort(%{ingestedTime: nil, submission_status: status}), do: status
+  defp status_sort(%{ingestedTime: it}), do: it
```
cameronmarsh, line 177: do it

apps/andi/lib/andi_web/live/helpers/sorting_helpers.ex
```diff
@@ -0,0 +1,29 @@
+defmodule AndiWeb.Helpers.SortingHelpers do
+  @moduledoc """
+  Helpers for sorting a list of maps by a field. Includes the ability to mix structs into the sorting (at least DateTime)
+  """
+  def sort_list_by_field(list, field, dir \\ "asc") do
```
cameronmarsh, line 5: I like this, we might be able to move the StructTools sort by sequence function here too. We made it public in our PR but it's private in master so yaidontknow

cameronmarsh (approved):

apps/andi/lib/andi_web/live/helpers/sorting_helpers.ex
bbrewer-forge: either way - mostly I was wanting to do some more extensive testing and it was painful to have to mock up entire datasets to do it in the table_test

PR #1079, opened 2020-11-12
Dataset link section

kcarra:
Just waiting on preamble text.


apps/andi/lib/andi/input_schemas/datasets/dataset.ex
```diff
@@ -15,6 +15,7 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
   @primary_key {:id, :string, autogenerate: false}
   schema "datasets" do
     field(:dlq_message, :map)
+    field(:datasetLink, :string)
```
bbrewer-forge, line 18: At some point, as a compromise, if we have to add more fields to this top level struct, I vote we at least make an embedded `submission_details` schema to hold these fields. I'm adding one two, and I think we're on the boundary of it being unclear what these are for.

bbrewer-forge (approved):

cameronmarsh (approved):
looks good

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
kcarra: Yes, I completely agree!

PR #1080, opened 2020-11-13
Extract secret

cameronmarsh:


apps/andi/lib/andi_web/views/error_helpers.ex
```diff
@@ -14,15 +14,22 @@ defmodule AndiWeb.ErrorHelpers do
...
-      content_tag(:span, translated,
-        class: "error-msg",
-        id: "#{field}-error-msg",
-        data: get_additional_content_tag_data(form, field, options)
-      )
-    end)
+    form.errors
+    |> Map.new()
+    |> Map.get(field)
+    |> generate_error_tag(field, form, options)
```
cameronmarsh, line 20: I put this here to prevent us from showing multiple error messages for the same field. I don't know of a case where we would want more than one error message in the same form but if there is one I can change this back

apps/andi/assets/css/extract_steps.scss
```diff
@@ -23,7 +23,19 @@
...
 .example-output {
-    color: green;
+    color: $color-success;
+}
+
+.secret-status-msg {
+    margin-left: 1rem;
+}
+
+.secret-save-success {
```
bbrewer-forge, line 33: Could we try to still stick with BEM conventions on this? As in, something like `secret__save--success` If we don't want to do that anymore, I understand, but I'd rather we discuss that in A&C and try to not have all sorts of patterns going on.

apps/andi/lib/andi/input_schemas/datasets/extract_steps/extract_step.ex
```diff
@@ -60,7 +60,7 @@ defmodule Andi.InputSchemas.Datasets.ExtractStep do
 
   def preload(struct), do: StructTools.preload(struct, [])
 
-  defp validate_type(%{changes: %{type: type}} = changeset) when type in ["http", "date"] do
+  defp validate_type(%{changes: %{type: type}} = changeset) when type in ["http", "date", "secret"] do
```
bbrewer-forge, line 63: Worth noting that Ecto (as of my recent merge) has an `Ecto.Enum` type that works for these kinds of type selections. Not expecting us to have to move to that, but worth noting

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -236,12 +236,13 @@ defmodule Andi.InputSchemas.InputConverter do
 
   defp update_context(context) do
     context
+    |> IO.inspect(label: "our context")
```
bbrewer-forge, line 239: Yarp

apps/andi/lib/andi_web/views/error_helpers.ex
bbrewer-forge: That's probably fine

apps/andi/test/unit/andi/secret_service_test.exs
```diff
@@ -25,14 +25,26 @@ defmodule Andi.SecretRetrieverTest do
       allow Vault.read(values.vault, "secrets/smart_city/auth0/andi"),
         return: {:ok, values.credentials}
 
-      assert SecretRetriever.retrieve_auth0_credentials() == {:ok, values.credentials}
+      assert SecretService.retrieve_auth0_credentials() == {:ok, values.credentials}
```
bbrewer-forge, line 28: How are we handling failures for this? Do we expect them to crash the page? Or are we dealing with them with an error tuple?

bbrewer-forge: As in, errors other than the `read` ones we've dealt with

apps/andi/test/unit/andi_web/live/extract_secret_step_form_test.exs
```diff
@@ -0,0 +1,103 @@
...
+      expect(Andi.SecretService.write("#{extract_step_id}___bob", %{"bob" => "secret_value"}), return: {:ok, :na})
+
+      render_change([extract_steps_form_view, "#step-#{extract_step_id}"], "validate", %{"form_data" => %{"destination" => "bob"}})
+      html = render_change([extract_steps_form_view, "#step-#{extract_step_id}"], "save_secret", %{"secret" => "secret_value"})
+
+      success_text = get_text(html, ".secret-status-msg")
+      assert success_text == "Secret saved successfully!"
+    end
+
+    test "displays error message when secret cannot be saved", %{
```
bbrewer-forge, line 83: Good to see - just want to make sure we have a test in that module that shows this.

apps/andi/test/unit/andi/secret_service_test.exs
cameronmarsh: Right now they are crashing the app

apps/andi/lib/andi/input_schemas/datasets/extract_steps/extract_date_step.ex
```diff
@@ -85,23 +84,19 @@ defmodule Andi.InputSchemas.Datasets.ExtractDateStep do
 
   defp validate_time_unit(changeset), do: changeset
 
-  defp validate_delta_change(%{changes: %{deltaTimeUnit: deltaTimeUnit}} = changeset) when deltaTimeUnit not in [nil, ""] do
+  defp validate_delta_change(%{changes: %{deltaTimeUnit: delta_time_unit}} = changeset) when delta_time_unit not in [nil, ""] do
```
bbrewer-forge, line 87: Thank you :|

apps/andi/test/unit/andi/secret_service_test.exs
bbrewer-forge: Oh good

bbrewer-forge (approved):

PR #1086, opened 2020-11-18
refactor ssui

kcarra:


bbrewer-forge (commented):
This looks good to me. It's definitely clear how much extra stuff it keeps out of the way in each form schema. I'll defer to Tim and Cam for approval, if they absolutely hate it, but I think we can give this a try without painting ourselves into a corner.

cameronmarsh:
I personally still think that new schemas will be harder to keep track of in the long run, and if we keep adding new schemas it just makes it harder to clean andi up if we ever do. I don't hate this enough to request changes but I will also defer approval before Tim takes a look at it

LtChae (approved):
I think this will do the job. The split schemas feel a little redundant, especially the one with only a single field, but without knowing for sure how we're going to refactor the structure of all this I'd say it will do for now.

PR #1090, opened 2020-11-19
Submit section

kcarra:


apps/andi/assets/css/app.scss
```diff
@@ -117,6 +123,11 @@ body {
   color: white;
```
cameronmarsh, line 123: is there a way we can combine some of these? seems like unnecessary duplication

apps/andi/lib/andi_web/live/submit_live_view/review_submission.ex
```diff
@@ -0,0 +1,77 @@
...
+  """
+  use Phoenix.LiveView
+  require Logger
+
+  def mount(_, %{"dataset" => dataset}, socket) do
+    AndiWeb.Endpoint.subscribe("toggle-visibility")
+
+    {:ok,
+     assign(socket,
+       changeset: %{valid?: true, changes: %{}},
```
cameronmarsh, line 13: I think you can get rid of this I don't see it being used anywhere

bbrewer-forge (approved):
It's fine

apps/andi/lib/andi_web/live/submit_live_view/review_submission.ex
```diff
@@ -0,0 +1,77 @@
...
+            </div>
+          </div>
+        </div>
+
+        <div class="form-section">
+          <div class="component-edit-section--<%= @visibility %>">
+            <div class="review-form-header">
+              <h4>Please ensure that there are no errors or omitted fields before submitting. Upon submission, the Data Curator will review your dataset for completeness, efficacy, and accuracy. You may check the status of your submission by returning to your submission portal homepage. If you have any questions, please contact the <a href="https://www.smartcolumbusos.com/contact-us" target="_blank">Data Curator</a></h4>
+            </div>
+
```
cameronmarsh, line 48: Are there no buttons on this page or am I missing something?

apps/andi/lib/andi_web/live/submit_live_view/review_submission.ex
kcarra: yah, it was messing up something in the form_section, but since it's no longer needed it can be removed

apps/andi/lib/andi_web/live/submit_live_view/review_submission.ex
kcarra: they've been removed since we are going to play a card to put them at the bottom of the page

PR #1093, opened 2020-11-23
Extract step delete

cameronmarsh:


sukaant (approved):

apps/reaper/lib/reaper/event/event_handler.ex
```diff
@@ -90,7 +90,13 @@ defmodule Reaper.Event.EventHandler do
       | technical: %{
           shapefile_dataset.technical
           | sourceFormat: hosted_file.mime_type,
-            sourceUrl: "s3://#{hosted_file.bucket}/#{hosted_file.key}"
+            extractSteps: [
```
cameronmarsh, line 93: was unsure about replacing all extract steps here so lmk if this is a mistake

apps/reaper/lib/reaper/file_ingest/processor.ex
```diff
@@ -26,8 +27,13 @@ defmodule Reaper.FileIngest.Processor do
     dataset = Provisioner.provision(unprovisioned_dataset)
     filename = get_filename(dataset)
 
-    dataset
-    |> UrlBuilder.build()
+    http_extract_step =
```
cameronmarsh, line 30: was also unsure if there will be geojson datasets that have more than one extract step, so I can change this function if needed. It's currently only looking at the last step, which it assumes should be an http step

apps/andi/lib/andi_web/input_schemas/dataset_link_form_schema.ex
```diff
@@ -20,14 +20,14 @@ defmodule AndiWeb.InputSchemas.DatasetLinkFormSchema do
 
   def changeset(changes), do: changeset(%__MODULE__{}, changes)
 
-  def changeset(datasetLink, changes) do
-    datasetLink
+  def changeset(dataset_link, changes) do
```
bbrewer-forge, line 23: thank you for fixing the problem you introduced ;)

apps/andi/lib/andi_web/live/edit_live_view/extract_steps/extract_step_placeholder.ex
```diff
@@ -0,0 +1,24 @@
+defmodule AndiWeb.ExtractSteps.ExtractPlaceholderStepForm do
```
bbrewer-forge, line 1: I like this idea

apps/reaper/lib/reaper/event/event_handler.ex
bbrewer-forge: I think this is an exceptional enough case that you shouldn't worry too much about it. We should support more of them at some point though

bbrewer-forge (approved):

PR #1105, opened 2020-12-07
Andi fixes

cameronmarsh:


apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -390,4 +391,10 @@ defmodule Andi.InputSchemas.InputConverter do
     end)
     |> elem(1)
   end
+
+  defp remove_query_params_from_url(url) do
```
sukaant, line 395: Is there any library which can do this? otherwise it is fine.

sukaant (approved):

PR #1110, opened 2020-12-08
Move save buttons

kcarra:


apps/andi/lib/andi_web/live/edit_live_view/data_dictionary_form.ex
```diff
@@ -18,6 +18,8 @@ defmodule AndiWeb.EditLiveView.DataDictionaryForm do
 
   def mount(_, %{"dataset" => dataset, "is_curator" => is_curator}, socket) do
     new_changeset = DataDictionaryFormSchema.changeset_from_andi_dataset(dataset)
+    send(socket.parent_pid, {:update_data_dictionary_status, new_changeset.valid?})
```
cameronmarsh, line 21: what is this line for?

cameronmarsh: nvm I see it below

cameronmarsh (commented):
testing this out, it doesn't seem like the save triggers validation errors in the individual sections. I feel like we should address this before merging

cameronmarsh (approved):

PR #1124, opened 2020-12-15
Validate schema fields have valid presto characters

cameronmarsh:
Datastillery/smartcitiesdata#909


LtChae (approved):
Do we have any concerns about this making existing datasets invalid?

cameronmarsh:
Shouldn't be a problem until we need to repost them, in which case we would want valid names

PR #1130, opened 2020-12-17
Blocking xls file upload

sukaant:


apps/andi/lib/andi_web/live/edit_live_view/data_dictionary_form.ex
```diff
@@ -160,7 +160,7 @@ defmodule AndiWeb.EditLiveView.DataDictionaryForm do
   end
 
   def handle_event("file_upload", %{"fileType" => file_type}, socket)
-      when file_type not in ["text/csv", "application/json", "application/vnd.ms-excel"] do
```
cameronmarsh, line 163: I'm not sure exactly what your card is, but I don't think this does what you think. We actually had to put this mime type in here to allow `csv` files for windows machines. I am not sure if there is a different mime type for `xlsx` files, but I feel like we need to target the extension of the file

sukaant: The card is to block xlsx and xls file in windows and `application/vnd.ms-excel` allows xls file only.

JarredOlson (commented):
I thought we talked about doing this check in the front end and using this library: https://www.npmjs.com/package/file-type

PR #1149, opened 2021-01-18
Upload dataset sample

kcarra:


apps/andi/lib/andi_web/live/submit_live_view/upload_data_dictionary.ex
```diff
@@ -0,0 +1,248 @@
...
+            </div>
+            <div class="upload-section">
+
+              <div class="upload-data-dictionary-form__file-upload">
+                <div class="file-input-button">
+                  <%= label(f, :sample_dataset, "Select File", class: "file-upload-label") %>
+                  <%= file_input(f, :sample_dataset, phx_hook: "readFile", accept: "text/csv, application/json") %>
+                </div>
+                <div class="sample-file-display">
+                  <%= hidden_input(f, :datasetLink) %>
```
cameronmarsh, line 85: I don't think you need a hidden input if you are already referencing the value on the line below. maybe I'm wrong but what are you using this for?

apps/andi/lib/andi_web/live/submit_live_view/upload_data_dictionary.ex
```diff
@@ -0,0 +1,248 @@
...
+      when file_type in ["text/csv", "application/vnd.ms-excel"] do
+    dataset_id = socket.assigns.dataset_id
+    dataset_link = "#{@bucket_path}#{dataset_id}/#{file_name}"
+
+    with {:ok, presigned_url} <- presigned_url(dataset_id, file_name),
+         {:ok, _} <- upload_sample_dataset(presigned_url, file, "text/csv") do
+      %{datasetLink: dataset_link}
+      |> DatasetLinkFormSchema.changeset_from_form_data()
+      |> send_dataset_link_status(socket)
+      |> complete_validation(socket)
```
cameronmarsh, line 158: this might be better broken out into its own function, like the `complete_validation` functions in some of our other liveview pages

apps/andi/lib/andi_web/views/error_helpers.ex
```diff
@@ -89,9 +89,8 @@ defmodule AndiWeb.ErrorHelpers do
 
   defp interpret_error_message(_message, :schema, _), do: "Please add a field to continue"
 
-  defp interpret_error_message(_message, :datasetLink, _), do: "Please enter a valid URL to your externally-hosted dataset"
```
cameronmarsh, line 92: I added this error message as a part of a card that gave the user more descriptive error messages. I would check with Jake to see if we want a more descriptive message than the default

cameronmarsh (approved):

apps/andi/lib/andi_web/views/error_helpers.ex
kcarra: The error messaging is all done in the event handlers for this field. Essentially for the size being to large, the format, or an error uploading the file

apps/andi/lib/andi_web/live/submit_live_view/upload_data_dictionary.ex
kcarra: I use this so I can display the file name that was uploaded. I had to override the styling of the upload button so it no longer does this by default

PR #1167, opened 2021-02-01
Download sample link

cameronmarsh:


LtChae (approved):
I assume that for now the way someone would view these audit logs would be through the console or database?

cameronmarsh:
> I assume that for now the way someone would view these audit logs would be through the console or database?

Yeah

kcarra (approved):

PR #1174, opened 2021-02-04
Log dataset sample upload

cameronmarsh:


apps/andi/lib/andi/schemas/dataset_upload.ex
```diff
@@ -0,0 +1,35 @@
...
+  import Ecto.Changeset
+
+  alias Andi.InputSchemas.StructTools
+
+  @primary_key {:id, Ecto.UUID, autogenerate: true}
+  schema "dataset_upload" do
+    field(:dataset_id, Ecto.UUID)
+    field(:timestamp, :utc_datetime)
+    field(:user_uploading, :string)
+    field(:upload_success, :boolean)
```
cameronmarsh, line 15: Anything else anyone would want to save here?

LtChae (approved):

apps/andi/lib/andi/schemas/dataset_upload.ex
LtChae: File name?

LtChae: MD5 checksum?

PR #1188, opened 2021-02-17
Andi providers

cameronmarsh:


apps/andi/lib/andi_web/live/edit_live_view/data_dictionary_field_editor.ex
```diff
@@ -63,8 +63,8 @@ defmodule AndiWeb.EditLiveView.DataDictionaryFieldEditor do
 
             <div class="inline" style="align-items: baseline;">
               <%= checkbox(@form, :use_default, id: id <> "__use-default", value: using_default) %>
-              <%= label(@form, :default, "Default Offset", class: "label") %>
-              <div class="test-status__tooltip-wrapper"><p phx-hook="addTooltip" data-tooltip-content="If the data ingested does not have a <%= field_type %> value, the system can add this value during ingestion" class="add-default-tooltip">Help</p></div>
+              <%= label(@form, :default, "Set the Default Offset #{String.capitalize(field_type)}", class: "label") %>
```
LtChae, line 66: The use of the word Offset here is still weird but I can't think of anything better than removing it.

LtChae (approved):

apps/andi/lib/andi_web/live/edit_live_view/data_dictionary_field_editor.ex
ksmith-accenture: Maybe "Default date/time offset"?

cameronmarsh: can do

PR #1194, opened 2021-02-19
Change storage of provider offset in postgres 

cameronmarsh:


LtChae (approved):
Double check that the api works with the providers still as well, but looks good

PR #1203, opened 2021-03-08
Limit data org name length

LtChae:


cameronmarsh (approved):
Looks good as long as it doesn't significantly slow down the data name generation. I think it's on debounce anyway so I don't think it will be an issue

PR #1212, opened 2021-03-18
Reaper compression handling

LtChae:
Some test cleanup was done, with unused imports and variables cleaned.

### DIFF
--- a/apps/reaper/lib/reaper/auth_retriever.ex
+++ b/apps/reaper/lib/reaper/auth_retriever.ex
@@ -4,7 +4,7 @@ defmodule Reaper.AuthRetriever do
   """
   alias Reaper.Collections.Extractions
   alias Reaper.Cache.AuthCache
-  alias Reaper.UrlBuilder
+  alias Reaper.Util
 
   def authorize(dataset_id, url, body, encode_method, headers, cache_ttl) when is_list(headers) do
     authorize(dataset_id, url, body, encode_method, Enum.into(headers, %{}), cache_ttl)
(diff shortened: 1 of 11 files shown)


apps/reaper/lib/reaper/auth_retriever.ex
```diff
@@ -74,6 +75,9 @@ defmodule Reaper.AuthRetriever do
     end
   end
 
+  defp handle_content_encoding(body, "gzip"), do: :zlib.gunzip(body)
```
JarredOlson, line 78: does this have potential error cases?

apps/reaper/lib/reaper/util.ex
```diff
@@ -5,4 +5,11 @@ defmodule Reaper.Util do
   def deep_merge(left, right), do: Map.merge(left, right, &deep_resolve/3)
   defp deep_resolve(_key, %{} = left, %{} = right), do: deep_merge(left, right)
   defp deep_resolve(_key, _left, right), do: right
+
+  def get_header_value(headers, header_name) do
+    case Enum.find(headers, fn {name, _value} -> String.downcase(name) == header_name end) do
```
JarredOlson, line 10: I would downcase header_name as well just to be safe.  I know this is only being used in 1 place currently with lower case value but it can prevent issues down the line.

JarredOlson (approved):

apps/reaper/lib/reaper/auth_retriever.ex
LtChae: Yes, now handled

apps/reaper/lib/reaper/util.ex
LtChae: Done

PR #1242, opened 2021-05-18
Users table

kcarra:


LtChae (approved):
Is there any practical way to downgrade the integration tests to unit tests? The Andi suite is pretty hefty already.

kcarra:
> Is there any practical way to downgrade the integration tests to unit tests? The Andi suite is pretty hefty already.

hmmm definitely agree that it is pretty hefty, not sure if there's a way to downgrade it to a unit test since we need some of the rendering functionality. There's no integration tests for the organizations table, not sure what the reasoning for this was, but if we don't think it's worth having this test I'm totally fine with removing it.

PR #1250, opened 2021-06-03
Change user role

kcarra:


apps/andi/lib/andi/services/auth0_management_service.ex
```diff
@@ -29,14 +29,49 @@ defmodule Andi.Services.Auth0Management do
...
+             headers: [{"Authorization", "Bearer #{access_token}"}, {"content-type", "application/json"}]
+           ) do
+      {:ok, response |> Map.get(:body)}
+    else
+      {:error, reason} ->
+        Logger.error("Unable to assign user role: #{reason}")
+        {:error, :assign_auth0_role}
+    end
+  end
+
```
LtChae, line 74: Is there a good way to unify these? They seem identical except for the http method.

apps/andi/lib/andi_web/live/user_live_view/edit_user_live_view.ex
```diff
@@ -125,8 +164,35 @@ defmodule AndiWeb.UserLiveView.EditUserLiveView do
...
     {:ok, event_data} = UserOrganizationAssociate.new(%{subject_id: user.subject_id, org_id: org_id, email: user.email})
     Brook.Event.send(:andi, user_organization_associate(), :andi, event_data)
   end
+
+  defp parse_roles(roles) when length(roles) == 0, do: []
+
+  defp parse_roles(roles) do
+    roles = roles |> Enum.map(fn %{"id" => id, "description" => description} -> {description, id} end)
+    roles
+  end
```
LtChae, line 193: No need to assign to role here, implicit return will do it.

LtChae (approved):
Looks good. Minor refactor opportunities

PR #1263, opened 2021-07-06
Smrt-1255 Remediate Organization Tech Debt

christinepoydence:


apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -122,21 +128,72 @@ defmodule Andi.InputSchemas.Datasets do
...
+    }
 
-      owner_id ->
-        existing_dataset
-        |> update(%{technical: technical_changes, business: business_changes, id: dataset_id, owner_id: owner_id, datasetLink: dataset_link})
-    end
+    existing_dataset |> update_dataset(owner_id, org_changed, changes)
+  end
+
+  def update_dataset(existing_dataset, nil, false, changes) do
```
LtChae, line 155: Good use of function hydras. Is there a good way to make each of these heads smaller? I see some commonality between them, but I also don't want to make unnecessary repeat database calls for cleaner code.

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
```diff
@@ -77,18 +107,33 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
   def validate_unique_system_name(%{changes: %{technical: technical}} = changeset) do
     id = Ecto.Changeset.get_field(changeset, :id)
     data_name = Ecto.Changeset.get_change(technical, :dataName)
-    org_name = Ecto.Changeset.get_change(technical, :orgName)
+    org_id = Ecto.Changeset.get_field(changeset, :organization_id)
+
+    case is_nil(org_id) do
```
LtChae, line 112: Consider breaking this case into a function and passing the result to one copy of the duplicated lines below

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
```diff
@@ -77,18 +107,33 @@ defmodule Andi.InputSchemas.Datasets.Dataset do
...
   end
 
   def validate_unique_system_name(changeset) do
     id = Ecto.Changeset.get_field(changeset, :datasetId)
     data_name = Ecto.Changeset.get_change(changeset, :dataName)
-    org_name = Ecto.Changeset.get_change(changeset, :orgName)
+    org_id = Ecto.Changeset.get_field(changeset, :organization_id)
 
-    check_uniqueness(changeset, id, data_name, org_name)
+    case is_nil(org_id) do
```
LtChae, line 129: Ditto. A common "get orgName if it exists" method might be helpful here

apps/andi/lib/andi_web/live/user_live_view/edit_user_live_view.ex
```diff
@@ -34,7 +34,7 @@ defmodule AndiWeb.UserLiveView.EditUserLiveView do
 
               <div class="user-form__organizations">
                 <%= label(:organization, :org_id, class: "label") %>
-                <%= select(:organiation, :org_id, MetadataFormHelpers.get_org_options(), [class: "select", readonly: true]) %>
+                <%= select(:organization, :org_id, MetadataFormHelpers.get_org_options(), [class: "select", readonly: true]) %>
```
LtChae, line 37: Good typo catch

apps/andi/priv/repo/migrations/20210623201753_add_org_id.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Andi.Repo.Migrations.AddOrgAssociation do
+  use Ecto.Migration
+  alias Andi.InputSchemas.Datasets
+  alias Andi.InputSchemas.Datasets.Dataset
+  alias Andi.Repo
+
+  def change do
```
LtChae, line 7: Don't forget to plan how to do the migration

LtChae (approved):
A few style comments etc.

apps/andi/lib/andi/event/event_handler.ex
```diff
@@ -30,16 +30,22 @@ defmodule Andi.Event.EventHandler do
   @instance_name Andi.instance_name()
 
   def handle_event(%Brook.Event{type: dataset_update(), data: %Dataset{} = data, author: author}) do
-    dataset_update()
-    |> add_event_count(author, data.id)
+    case Organizations.get(data.organization_id) do
+      nil ->
+        Logger.error("Organization #{data.organization_id} does not exist. Discarding update event for dataset #{data.id}.")
```
ksmith-accenture, line 35: Would this error get propagated to the user? (@LtChae)

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -122,21 +128,72 @@ defmodule Andi.InputSchemas.Datasets do
...
+      organization_id: organization_id,
+      dataset_link: dataset_link
+    }
 
-      owner_id ->
-        existing_dataset
-        |> update(%{technical: technical_changes, business: business_changes, id: dataset_id, owner_id: owner_id, datasetLink: dataset_link})
-    end
+    existing_dataset |> update_dataset(owner_id, org_changed, changes)
+  end
```
ksmith-accenture, line 153: So Clean

apps/andi/lib/andi/input_schemas/datasets.ex
ksmith-accenture: Wouldn't only one of these functions call with each underlying data update due to pattern matching?

apps/andi/lib/andi/input_schemas/datasets.ex
```diff
@@ -274,6 +333,10 @@ defmodule Andi.InputSchemas.Datasets do
   defp extract_owner_id(_, %{ownerId: ownerId}), do: ownerId
   defp extract_owner_id(_, _), do: nil
 
+  defp extract_organization_id(nil, %{organization_id: organization_id}), do: organization_id
```
ksmith-accenture, line 336: Is this one necessary if the next one catches nil with the `_` and doesn't do anything functionally different?

apps/andi/lib/andi/input_schemas/organization.ex
```diff
@@ -73,6 +73,8 @@ defmodule Andi.InputSchemas.Organization do
     end
   end
 
+  def preload(struct), do: struct
```
ksmith-accenture, line 76: Is this to meet an interface's [sic] requirements?

apps/andi/lib/andi_web/live/dataset_live_view.ex
```diff
@@ -198,15 +198,29 @@ defmodule AndiWeb.DatasetLiveView do
...
+      "org_title" => dataset.organization.orgTitle,
       "data_title" => dataset.business.dataTitle,
       "status" => status(dataset)
     }
   end
 
+  defp to_view_model(dataset) do
+    org_id = dataset.organization_id
+    view_model(dataset, org_id)
+  end
```
ksmith-accenture, line 222: I believe this can be structurally cleaned up into two definitions of to_view_model using a guard to check if the organization id is null

apps/andi/lib/andi_web/views/options.ex
```diff
@@ -152,17 +152,11 @@ defmodule AndiWeb.Views.Options do
   end
 
   def organizations(stored_organizations) do
-    case stored_organizations do
```
ksmith-accenture, line 155: Does removing this guard leave the code vulnerable to error if nil is (somehow) passed or would it return an empty list?

apps/andi/lib/andi/event/event_handler.ex
LtChae: No, the user has no visibility into issues on the event stream

apps/andi/lib/andi/input_schemas/organization.ex
christinepoydence: Yeah- we made code changes to preload organization from dataset. To support that, organization needed to also have preload defined.

apps/andi/lib/andi/input_schemas/datasets.ex
christinepoydence: That's a good question- This code was directly modeled on the `extract_owner_id` function above it. We will do some testing to see if all three variations of the function are needed.

apps/andi/lib/andi_web/views/options.ex
christinepoydence: The code that calls this function was changed to use Postgres instead of the OrgStore to retrieve the list of stored organizations that is passed into this function. The Postgres Organizations.get_all() call returns an empty list if no orgs are present, and it returns a list if there are orgs. This method no longer accepts a tuple and will always receive a list, so the guard is no longer needed.

apps/andi/lib/andi_web/live/dataset_live_view.ex
christinepoydence: @ksmith-accenture, I tried to take this suggestion by doing something like:
` defp to_view_model(dataset) when is_nil(dataset.organization_id) do
    ...map here...
end`

However, that resulted in: 
`(CompileError) lib/andi_web/live/dataset_live_view.ex:201: cannot invoke remote function dataset.organization_id/0 inside guards`

Could you give an example of what you mean? (I think I may have misunderstood what you're looking for here.)

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
christinepoydence: Fixed!

apps/andi/lib/andi/input_schemas/datasets/dataset.ex
christinepoydence: Fixed!

apps/andi/lib/andi/input_schemas/datasets.ex
christinepoydence: Good catch, @ksmith-accenture. Fixed!

apps/andi/lib/andi/input_schemas/datasets.ex
christinepoydence: @ksmith-accenture - true!

@LtChae - we tried doing something like this:
`
 def update_from_form(dataset_id, form_changes) do
    org_changed = is_org_changed?(form_changes)
    existing_dataset = get(dataset_id)
    changeset = InputConverter.andi_dataset_to_full_ui_changeset(existing_dataset)

    technical_changes =
      changeset
      |> Changeset.get_change(:technical)
      |> Changeset.apply_changes()
      |> StructTools.to_map()
      |> Map.merge(form_changes)

    business_changes =
      changeset
      |> Changeset.get_change(:business)
      |> Changeset.apply_changes()
      |> StructTools.to_map()
      |> Map.merge(form_changes)

    owner_id =
      changeset
      |> Changeset.get_field(:owner_id, nil)
      |> extract_owner_id(form_changes)

    organization_id =
      changeset
      |> Changeset.get_field(:organization_id, nil)
      |> extract_organization_id(form_changes)

    dataset_link =
      changeset
      |> Changeset.apply_changes()
      |> StructTools.to_map()
      |> Map.merge(form_changes)
      |> Map.get(:datasetLink)

    updated_if_different_changes = %{
      owner_id: owner_id,
      organization_id: organization_id
    }

    always_updated_changes = %{
      dataset_id: dataset_id,
      technical_changes: technical_changes,
      business_changes: business_changes,
      dataset_link: dataset_link
    }

    existing_dataset |> update_dataset(owner_id, org_changed, always_updated_changes, updated_if_different_changes)
  end

  def update_dataset(existing_dataset, nil, false, always_updated_changes, _updated_if_different_changes) do
    existing_dataset
    |> update(always_updated_changes)
  end

  def update_dataset(existing_dataset, owner_id, false, always_updated_changes, updated_if_different_changes) do
    changes = Map.put(always_updated_changes, :owner_id, updated_if_different_changes.owner_id)
    existing_dataset
    |> update(changes)
  end

  def update_dataset(existing_dataset, nil, true, always_updated_changes, updated_if_different_changes) do
    changes = Map.put(always_updated_changes, :organization_id, updated_if_different_changes.organization_id)
    existing_dataset
    |> update(changes)
  end

  def update_dataset(existing_dataset, owner_id, true, always_updated_changes, updated_if_different_changes) do
    changes = Map.put(always_updated_changes, :organization_id, updated_if_different_changes.organization_id)
    changes = Map.put(always_updated_changes, :owner_id, updated_if_different_changes.owner_id)
    existing_dataset
    |> update(changes)
  end
`

However, after we started heading down this road, we found that it wasn't saving that many lines of code, and wasn't drastically improving readability. What would your thoughts be on leaving this as is?

apps/andi/lib/andi_web/live/dataset_live_view.ex
christinepoydence: Tim pointed out my error here- I forgot to use pattern matching! Fixed.

apps/andi/lib/andi/input_schemas/datasets.ex
christinepoydence: Spoke with Tim- decided to leave this as is!

PR #1294, opened 2021-09-15
Add Authorization to Raptor

christinepoydence:


LtChae (commented):
I'm surprised by the amount of code added to handle redis interactions. I was under the impression that Brook handled the majority of that. Was there a gap in the functionality that required us to make new modules to handle it?

https://hexdocs.pm/brook/Brook.Storage.html#content

christinepoydence:
> I'm surprised by the amount of code added to handle redis interactions. I was under the impression that Brook handled the majority of that. Was there a gap in the functionality that required us to make new modules to handle it?
> 
> https://hexdocs.pm/brook/Brook.Storage.html#content

@LtChae - we did this because it seems as though Brook only allows you to store the events as they were sent in the event stream. We wanted slightly different behavior!

LtChae (approved):

PR #1321, opened 2022-01-10
Raptor url config

bmitchinson:
## Description

A while ago we moved raptor logic to a shared library. Old config values were not removed when this happened.
The raptor URL was hardcoded, this change allows it to be configured with an environment variable.

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/raptor_service/lib/raptor_service.ex
+++ b/apps/raptor_service/lib/raptor_service.ex
@@ -2,8 +2,7 @@ defmodule RaptorService do
   use Properties, otp_app: :raptor_service
   require Logger
 
-  def is_authorized(api_key, system_name) do
-    raptor_url = System.get_env("RAPTOR_URL")
+  def is_authorized(raptor_url, api_key, system_name) do
     case HTTPoison.get(raptor_url_with_params(raptor_url, api_key, system_name)) do
       {:ok, %{body: body}} ->
         {:ok, is_authorized} = Jason.decode(body)
--- a/apps/raptor_service/lib/raptor_service.ex
+++ b/apps/raptor_service/lib/raptor_service.ex
@@ -2,10 +2,9 @@ defmodule RaptorService do
   use Properties, otp_app: :raptor_service
   require Logger
 
-  getter(:raptor_url, generic: true)
-
   def is_authorized(api_key, system_name) do
-    case HTTPoison.get(raptor_url_with_params(raptor_url(), api_key, system_name)) do
+    raptor_url = System.get_env("RAPTOR_URL")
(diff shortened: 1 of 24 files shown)


apps/raptor_service/lib/raptor_service.ex
```diff
@@ -2,10 +2,9 @@ defmodule RaptorService do
   use Properties, otp_app: :raptor_service
   require Logger
 
-  getter(:raptor_url, generic: true)
-
   def is_authorized(api_key, system_name) do
-    case HTTPoison.get(raptor_url_with_params(raptor_url(), api_key, system_name)) do
+    raptor_url = System.get_env("RAPTOR_URL")
+    case HTTPoison.get(raptor_url_with_params(raptor_url, api_key, system_name)) do
```
LtChae, line 7: Is there a reason we switched from the getter to a direct environment call?

We try to use the getter macros where possible.

scripts/build.sh
```diff
@@ -35,4 +37,11 @@ fi
 echo "Building $repository:$tag"
 docker build -t $repository:$tag apps/$app
 
-echo "${nl}Built ${repository}:${tag}${nl}"
\ No newline at end of file
+echo "${nl}Built ${repository}:${tag}$"
+
+if [[ $* == *--push* ]]; then
+  echo "Pushing ${repository}:${tag}"
+  docker push "${repository}:${tag}"
```
LtChae, line 44: Thanks for adding this.

LtChae (approved):

apps/raptor_service/lib/raptor_service.ex
bmitchinson: It's been difficult to find much info on elixir "getter macros"

Does using a getter macro allow us to read in the env variable at runtime?

It seemed like using .get_env in the exs config file loaded it in at compile time, causing it to be nil. Does using a getter macro help us get around that, and tell elixir to run .get_env while executing?

I moved it here so that it would pull it in at runtime. Could have sword I tried keeping the getter setup, but was still getting `nil` for the env value.

LtChae: Envs should be loaded into the config in runtime.exs. In lower configs it's set directly in the config.

The getter loads it from the config.

bmitchinson: Okay thanks, that clears things up.

Can I add a runtime.exs to the raptor_service? I thought since it was just a defined function without a running process, runtime.exs wasn't an option.

bmitchinson: This was resolved by having applications that utilize the raptor_service provide the raptor url themselves

PR #1322, opened 2022-01-14
Datastillery/internal#113

christinepoydence:
update tableau connector to use API Key


bmitchinson (approved):
Nice. Lots of old code removed here, looking forward to the exit Monday. 

I didn't realize it originally relied on Auth0? How were we able to remove that dependency?

bmitchinson:
Also just confirming, we don't have to update any of the Tableau readme right? Doesn't look like it since it seems to function the same but, thought I'd double check

LtChae (approved):

PR #1337, opened 2022-02-28
Type Conversion Transformation

bmitchinson:
## Reminders:

      in `mix.exs`?
- [x] ~If altering an API endpoint, was the relevant postman collection updated?~

### DIFF
--- a/apps/transformers/lib/transformations/type_conversion.ex
+++ b/apps/transformers/lib/transformations/type_conversion.ex
@@ -13,7 +13,7 @@ defmodule Transformers.TypeConversion do
       parse_or_error(conversion_function, payload, field, value, target_type)
     else
       {:error, reason} -> {:error, reason}
-      nil_payload -> nil_payload
+      nil_payload -> {:ok, nil_payload}
     end
   end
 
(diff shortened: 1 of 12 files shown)


apps/transformers/lib/transformations/type_conversion.ex
```diff
@@ -0,0 +1,66 @@
...
+      :ok
+    end
+  end
+
+  defp check_field_is_of_sourcetype(field, value, source_type) do
+    function =
+      case source_type do
+        "float" -> fn value -> is_float(value) end
+        "integer" -> fn value -> is_integer(value) end
+        "string" -> fn value -> is_bitstring(value) end
```
christinepoydence, line 47: Should we use `is_binary/1` for this? (https://stackoverflow.com/questions/54161232/use-is-bitstring-or-is-binary-in-elixir-guard-for-string)

christinepoydence (approved):
Very cool!

apps/transformers/lib/transformations/type_conversion.ex
bmitchinson: great find, thanks for the link.

PR #1338, opened 2022-03-02
Internal 549 Andi ingestions API

christinepoydence:
## Reminders:

      in `mix.exs`?


bmitchinson:
Reviewing this but, I did contribute to some of the earlier commits. Might be good to have an additional set of eyes on it.

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -0,0 +1,172 @@
...
+    changes_with_id = StructTools.ensure_id(ingestion, changes)
+
+    ingestion
+    |> cast(changes_with_id, @cast_fields, empty_values: [])
+    |> cast_assoc(:schema, with: &DataDictionary.changeset_for_draft/2)
+    |> cast_assoc(:extractSteps, with: &ExtractStep.changeset_for_draft/2)
+    |> foreign_key_constraint(:targetDataset)
+  end
+
+  def preload(struct), do: StructTools.preload(struct, [:schema, :extractSteps])
```
bmitchinson, line 103: Should "transformations" be here as well

Edit: On second thought, it looks like transformations is absent from other parts of this PR

I'm assuming now that we'll probably add transformations logic to ingestions once we're adding those for the first time. But for now, this PR restores the functionality of adding / removing datasets.

Disregard this comment then

bmitchinson (approved):

apps/andi/lib/andi/input_schemas/ingestion.ex
christinepoydence: Thanks, Ben! Exactly, transformations was out of scope for the initial ingestions work.

PR #1339, opened 2022-03-02
Add Datetime Transformation

bmitchinson:
## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transformations/date_time.ex
+++ b/apps/transformers/lib/transformations/date_time.ex
@@ -13,7 +13,7 @@ defmodule Transformers.DateTime do
          {:ok, payloadSourceValue} <- FieldFetcher.fetch_value(payload, sourceField),
          {:ok, sourceDatetime} <- parseTime(payloadSourceValue, sourceFormat, sourceField),
          {:ok, transformedDatetime} <- formatDateTime(sourceDatetime, targetFormat) do
-      {:ok, %{targetField => transformedDatetime}}
+      {:ok, payload |> Map.put(targetField, transformedDatetime)}
     else
       {:error, reason} ->
         {:error, reason}
--- a/apps/transformers/lib/transformations/date_time.ex
+++ b/apps/transformers/lib/transformations/date_time.ex
@@ -6,14 +6,15 @@ defmodule Transformers.DateTime do
   @impl Transformation
 
   def transform(payload, parameters) do
-    with {:ok, sourceField} <- FieldFetcher.fetch_parameter(parameters, :sourceField),
-         {:ok, sourceFormat} <- FieldFetcher.fetch_parameter(parameters, :sourceFormat),
-         {:ok, targetField} <- FieldFetcher.fetch_parameter(parameters, :targetField),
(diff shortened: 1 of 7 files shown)


apps/transformers/lib/transformations/date_time.ex
```diff
@@ -0,0 +1,46 @@
...
+    else
+      {:error, reason} ->
+        {:error, reason}
+
+      nil_payload ->
+        {:ok, nil_payload}
+    end
+  end
+
+  defp parseTime(dateString, dateFormat, sourceField) do
```
christinepoydence, line 26: This function is called `parseTime`, but it accepts a `dateString` and a `dateFormat`. Should it be renamed `parseDate`?

apps/transformers/lib/transformations/date_time.ex
```diff
@@ -0,0 +1,46 @@
...
+    else
+      {:error, timexReason} ->
+        {:error,
+         "Unable to parse datetime from \"#{sourceField}\" in format \"#{dateFormat}\": #{
+           timexReason
+         }"}
+    end
+  end
+
+  defp formatDateTime(dateTime, format) do
```
christinepoydence, line 38: super tiny thing - I think that elixir convention for functions is snake_case?

christinepoydence (approved):
This is great! General question about the datetime transformations - would it be worthwhile to document that the parsing and formatting uses the hex package Timex, so that users know where they can find documentation on acceptable values? (Maybe returned as part of the error message if there's a parsing failure?) Or are we planing to include that information on the UI?

bmitchinson:
> This is great! General question about the datetime transformations - would it be worthwhile to document that the parsing and formatting uses the hex package Timex, so that users know where they can find documentation on acceptable values? (Maybe returned as part of the error message if there's a parsing failure?) Or are we planing to include that information on the UI?

I believe when discussed using Timex we decided that information would be in the UI.

Open to document it in other places though. What were you thinking when you mentioned documenting it? The readme of Transformations?

apps/transformers/lib/transformations/date_time.ex
bmitchinson: good call 👍  changed

apps/transformers/lib/transformations/date_time.ex
bmitchinson: glad you mentioned this, my JS is showing. Fixed!

christinepoydence: Hah same, I find myself doing this all the time too!

christinepoydence:
> > This is great! General question about the datetime transformations - would it be worthwhile to document that the parsing and formatting uses the hex package Timex, so that users know where they can find documentation on acceptable values? (Maybe returned as part of the error message if there's a parsing failure?) Or are we planing to include that information on the UI?
> 
> I believe when discussed using Timex we decided that information would be in the UI.
> 
> Open to document it in other places though. What were you thinking when you mentioned documenting it? The readme of Transformations?

I'm good with putting the information in the UI when we get there! Yeah, I guess the README would be the other option, but it was pretty easy for me to just check the mix.exs file to find the package, so it probably isn't necessary.

PR #1341, opened 2022-03-07
Regex replace

jfolk2015:
## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transformations/regex_replace.ex
+++ b/apps/transformers/lib/transformations/regex_replace.ex
@@ -12,7 +12,9 @@ defmodule Transformers.RegexReplace do
          {:ok, value} <- FieldFetcher.fetch_value(payload, source_field),
          {:ok, regex} <- RegexUtils.regex_compile(regex_pattern),
          :ok <- abort_if_not_string(replacement) do
-           {:ok, payload}
+           transformed_value = Regex.replace(regex, value, replacement)
+           transformed_payload = Map.put(payload, source_field, transformed_value)
+           {:ok, transformed_payload}
     else
       {:error, reason} -> {:error, reason}
     end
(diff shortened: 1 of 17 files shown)


apps/transformers/lib/transformations/regex_replace.ex
```diff
@@ -0,0 +1,31 @@
...
+      transformed_value = Regex.replace(regex, value, replacement)
+      transformed_payload = Map.put(payload, source_field, transformed_value)
+      {:ok, transformed_payload}
+    else
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp abort_if_not_string(value, field_name) do
+    if is_bitstring(value) do
```
christinepoydence, line 25: I think we might prefer `is_binary` over `is_bitstring`? 

(https://stackoverflow.com/questions/54161232/use-is-bitstring-or-is-binary-in-elixir-guard-for-string)

apps/transformers/test/unit/transformations/regex_replace_test.exs
```diff
@@ -0,0 +1,162 @@
+defmodule Transformers.RegexReplaceTest do
+  use ExUnit.Case
+
+  alias Transformers.RegexReplace
+
+  test "when source field not specified, return error" do
```
christinepoydence, line 6: Do you see any value in using a data test for these three (very similar) unit tests? Something like:

```
data_test "when #{missing_field} field not specified, return error" do
     payload = %{
       "something" => "123"
     }

     parameters = %{
       "regex" => "^(\\w+)",
       "replacement" => "a",
       "source" => "something"
     } |> Map.delete(missing_field)

     {:error, reason} = RegexReplace.transform(payload, parameters)

     assert reason == "Missing transformation parameter: #{error_message}"

     where([
        [:missing_field, :error_message],
        ["source", "sourceField"],
        ["regex", "regex"],
        ["replacement", "replacement"],
     ])
   end
```

Warning: The above is untested code.

christinepoydence (approved):

apps/transformers/lib/transformations/regex_replace.ex
jfolk2015: :( That's a confusing name and I wish they would just implement "is_string". Fixed.

apps/transformers/test/unit/transformations/regex_replace_test.exs
jfolk2015: After some googling, I'm not finding documentation for data_test. Can you point me in the right direction?

jfolk2015: Nevermind, found it on the Chechov library. I'll try it out in this case and we can see how we like it.

PR #1344, opened 2022-03-08
Internal 562 Update data_ingest_start() to use ingestions

christinepoydence:
## Reminders:

      in `mix.exs`?


apps/forklift/lib/forklift/event/event_handler.ex
```diff
@@ -21,16 +22,17 @@ defmodule Forklift.Event.EventHandler do
...
-    |> add_event_count(author, dataset.id)
+    |> add_event_count(author, dataset_id)
 
-    :ok = Forklift.DataReaderHelper.init(dataset)
+    dataset = Forklift.Datasets.get!(dataset_id)
 
-    Forklift.Datasets.update(dataset)
+    if dataset != nil do
+      :ok = Forklift.DataReaderHelper.init(dataset)
+    end
```
bmitchinson, line 35: can we add a more explicit return in the event that the dataset is nil? Thinking about implicit returns and how previously, we were previously returning the result of `Forklift.Datasets.update(dataset)`

apps/e2e/mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule E2E.MixProject do
 
   defp deps do
     [
+      {:alchemist, in_umbrella: true},
```
bmitchinson, line 28: can we bump the e2e mix version? To `0.2` since it includes alchemist now? 

(maybe this doesn't matter because we don't release the e2e test)

jfolk2015 (approved):

apps/e2e/mix.exs
christinepoydence: Updated!

bmitchinson (approved):
LGTM 👍

apps/forklift/lib/forklift/event/event_handler.ex
christinepoydence: Good catch! Added an explicit :ok to the end of the event handler function.

PR #1349, opened 2022-03-11
Init Access Group page with table headers

bmitchinson:
## Description:

- Adds Access Group route
- Adds table to the view on that route with headers in table

## Reminders:

      in `mix.exs`?
- [x] NA: If altering an API endpoint, was the relevant postman collection updated?

image


apps/andi/assets/css/access-groups.scss
```diff
@@ -0,0 +1,168 @@
...
+  border-right: $table-border;
+  height: $table-height;
+  max-width: 30rem;
+}
+
+.access-groups-table__status-cell {
+  width: 10%;
+}
+
+// .input-container {
```
christinepoydence, line 50: Do we want to leave this commented out code here?

apps/andi/test/integration/andi_web/live/dataset_live_view_test.exs
```diff
@@ -27,12 +27,6 @@ defmodule AndiWeb.DatasetLiveViewTest do
   @url_path "/datasets"
 
   describe "non-curator view" do
-    test "organization button is not shown", %{public_conn: conn} do
```
christinepoydence, line 30: Why did we remove this test?

christinepoydence: Wait- never mind, got to the next file!

christinepoydence (approved):

apps/andi/assets/css/access-groups.scss
bmitchinson: Nah, it's in datasets.scss anyway.

Removed 👍

apps/andi/test/integration/andi_web/live/dataset_live_view_test.exs
bmitchinson: 👍

christinepoydence (commented):
Should there be a unit test file for access_group_live_view.ex?

bmitchinson:
> Should there be a unit test file for access_group_live_view.ex?

Good Q, we had one for a minute, and then moved all it's contents to the access group test file.

There's not really any functionality in the view to test, since everything is inside the table component.

There are files like `dataset_live_view_test`, but they test the controls located outside of the table, and access group doesn't have any of those to test.

apps/andi/test/integration/andi_web/live/header_live_view_test.exs
```diff
@@ -0,0 +1,57 @@
...
+  import FlokiHelpers,
+    only: [
+      find_elements: 2
+    ]
+
+  @endpoint AndiWeb.Endpoint
+  @url_path "/datasets"
+
+  # TODO: Could we make this a unit test instead?
+  describe "non-curator view" do
```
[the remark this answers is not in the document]
bmitchinson: Added

PR #1357, opened 2022-03-21
Add delete to ingestions edit page

bmitchinson:
## Description

- Added delete button to ingestion page
  - Confirmation ingestion delete modal
  - Deletes ingestion through ingestion_delete event
- Add delete ingestion to postman documentation

## Reminders:

      in `mix.exs`?


.github/pull_request_template.md
```diff
@@ -1,5 +1,13 @@
+## [Ticket Link #111](https://app.zenhub.com/workspaces/mdot-615b97c1a5fde400126174f8/issues/urbanos-public/internal/111)
```
nicole-arruda, line 1: Cool!

apps/andi/lib/andi/services/ingestion_delete_service.ex
```diff
@@ -12,8 +12,12 @@ defmodule Andi.Services.IngestionDelete do
   """
   @spec delete(term()) :: {:ok, SmartCity.Ingestion.t()} | {:error, any()} | {:not_found, any()}
   def delete(ingestion_id) do
+    delete(ingestion_id, :api)
+  end
+
+  def delete(ingestion_id, user_id) do
     with {:ok, ingestion} when not is_nil(ingestion) <- IngestionStore.get(ingestion_id),
-         _ <- Andi.Schemas.AuditEvents.log_audit_event(:api, ingestion_delete(), ingestion),
+         _ <- Andi.Schemas.AuditEvents.log_audit_event(user_id, ingestion_delete(), ingestion),
```
christinepoydence, line 20: If this is changed to user id, what happens to the logging when a delete ingestion event is sent from the api?

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -3,20 +3,66 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
+  def handle_event("delete-confirmed", _, socket) do
+    ingestion_id = socket.assigns.ingestion.id
+    user_id = socket.assigns.user_id
+
+    case IngestionStore.get(ingestion_id) do
+      {:ok, nil} ->
+        Ingestions.delete(ingestion_id)
+
+      {:ok, smrt_ingestion} ->
+        IngestionDelete.delete(smrt_ingestion.id, user_id)
```
christinepoydence, line 59: So, in other places in the code, we send the brook event directly from the live view, instead of using the services that were written for the api. I'm fine either way, was just curious what made you decide to change the pattern?

christinepoydence (approved):
Approved, assuming that the audit event functionality still works for ingestion_deletes via the api correctly.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
bmitchinson: We liked using `IngestionDelete` because we got the audit table population for free.

In other words: We encapsulated calling `save_audit_thing` by reusing `IngestionDelete`, instead of having to call both:

```elixir
save_audit_log...
brook.event.send...
```

apps/andi/lib/andi/services/ingestion_delete_service.ex
bmitchinson: Our understanding was that by calling "delete(ingestion_id)" (which is what the API does), it'd call our new function hydra head (where you left your comment).

Line 14/15 should ensure that it keeps the same behavior, might just look strange in the diff

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
bmitchinson: Just manually confirmed that this still logs as expected 👍

apps/andi/lib/andi/services/ingestion_delete_service.ex
bmitchinson: Just manually confirmed that this still logs as expected 👍

PR #1360, opened 2022-03-25
Save Access Groups with Associated Datasets

christinepoydence:
## Description

 - There is a save button on the dataset search page.
 - Clicking the save button navigates back to the edit access group page.
 - If datasets were selected when save was clicked, those datasets are listed under "Datasets assigned to this access group" with their name, organization, and keywords.
 - When the access group is saved from the edit page, a dataset_access_group_associate event is sent.

  
image
 
 
## Reminders:

      in `mix.exs`?
NA If altering an API endpoint, was the relevant postman collection updated?
NAIf a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?

### DIFF
--- a/apps/andi/lib/andi/event/event_handler.ex
+++ b/apps/andi/lib/andi/event/event_handler.ex
@@ -23,6 +23,7 @@ defmodule Andi.Event.EventHandler do
   alias SmartCity.{Dataset, Organization, Ingestion}
   alias SmartCity.UserOrganizationAssociate
   alias SmartCity.UserOrganizationDisassociate
+  alias SmartCity.DatasetAccessGroupRelation
 
   alias Andi.Services.DatasetStore
   alias Andi.Services.OrgStore
@@ -122,13 +123,11 @@ defmodule Andi.Event.EventHandler do
 
   def handle_event(%Brook.Event{
         type: dataset_access_group_associate(),
-        data: %SmartCity.DatasetAccessGroupRelation{dataset_id: dataset_id, access_group_id: access_group_id},
+        data: %DatasetAccessGroupRelation{dataset_id: dataset_id, access_group_id: access_group_id},
         author: author
       }) do
     dataset_access_group_associate()
     |> add_event_count(author, nil)
-
-    IO.inspect(dataset_id, label: "REACHED EVENT HANDLER")
 
     case Andi.InputSchemas.Datasets.Dataset.associate_with_access_group(access_group_id, dataset_id) do
       {:error, error} ->
@@ -143,7 +142,7 @@ defmodule Andi.Event.EventHandler do
 
   def handle_event(%Brook.Event{
         type: dataset_access_group_disassociate(),
-        data: %SmartCity.DatasetAccessGroupRelation{dataset_id: dataset_id, access_group_id: access_group_id},
+        data: %DatasetAccessGroupRelation{dataset_id: dataset_id, access_group_id: access_group_id},
         author: author
       }) do
     dataset_access_group_disassociate()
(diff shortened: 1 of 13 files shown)


apps/andi/lib/andi/event/event_handler.ex
```diff
@@ -118,6 +121,46 @@ defmodule Andi.Event.EventHandler do
...
+        type: dataset_access_group_associate(),
+        data: %DatasetAccessGroupRelation{dataset_id: dataset_id, access_group_id: access_group_id},
+        author: author
+      }) do
+    dataset_access_group_associate()
+    |> add_event_count(author, nil)
+
+    case Andi.InputSchemas.Datasets.Dataset.associate_with_access_group(access_group_id, dataset_id) do
+      {:error, error} ->
+        Logger.error("Unable to associate dataset with access group #{access_group_id}: #{inspect(error)}. This event has been discarded.")
```
jfolk2015, line 134: Is it useful to include the dataset id in the error message too?

jfolk2015 (approved):
A minor suggestion, non-blocking

apps/andi/test/unit/andi/event/event_handler_test.exs
```diff
@@ -17,9 +16,36 @@ defmodule Andi.Event.EventHandlerTest do
...
+    access_group_id = UUID.uuid4()
+    {:ok, assoc} = DatasetAccessGroupRelation.new(%{dataset_id: dataset_id, access_group_id: access_group_id})
+    allow(Andi.InputSchemas.Datasets.Dataset.associate_with_access_group(access_group_id, dataset_id), return: :ok)
+    expect(TelemetryEvent.add_event_metrics(any(), [:events_handled]), return: :ok)
+
+    event = Brook.Event.new(type: dataset_access_group_associate(), data: assoc, author: :author)
+    EventHandler.handle_event(event)
+
+    assert_called(Andi.InputSchemas.Datasets.Dataset.associate_with_access_group(access_group_id, dataset_id))
+  end
```
bmitchinson, line 34: Do we want the assertion here to confirm that the association ended up in AccessGroup table? Or was it a deliberate choice to not assert side effects.

I see how by asserting that `Dataset.associate_with...` was called, it's testing that the event handler fulfilled it's responsibility.

Previously @LtChae advised that we might as well test the full effect to get some more coverage, but I like how you've done it here, since it's a Unit test. 

I feel like unit tests should only test the specific piece they're testing, and not their effects elsewhere. 

(ex: if we asserted it ended up in the DB, and we broke the Datasets inputschema, the event handler test would break, which doesn't seem right)

---

So, not really requesting a change here. Just curious of your thought process on what to assert for this was. Would like to align the way I think about it for future tickets.

apps/andi/test/integration/andi_web/live/access_group_live_view/edit_access_group_live_view_test.exs
```diff
@@ -278,6 +279,86 @@ defmodule AndiWeb.AccessGroupLiveView.EditAccessGroupLiveViewTest do
...
+      # verfy that the selected datasets appear in the datasets table
+      assert get_text(html, ".access-groups-dataset-table") =~ dataset_a.business.orgTitle
+      assert get_text(html, ".access-groups-dataset-table") =~ dataset_a.business.dataTitle
+
+      # save the changes to the access group
+      save_button = element(view, ".save-edit", "Save")
+      html = render_click(save_button)
+
+      eventually(fn ->
+        access_group = AccessGroups.get(access_group.id) |> Andi.Repo.preload(:datasets) |> IO.inspect(label: "here")
```
bmitchinson, line 356: IO inspect removal

bmitchinson (approved):

PR #1362, opened 2022-03-28
599: Ingestion extract steps

narruda789:
## Description
Adding extract steps for ingestions into the Andi UI.


apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -21,28 +22,78 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
+  def handle_info({:update_save_message, status}, socket) do
+    message = save_message(status == "valid" && socket.assigns.changeset.valid?)
+
+    {:noreply, assign(socket, click_id: UUID.uuid4(), save_success: true, success_message: message)}
+  end
+
+  # This handle_info takes care of all exceptions in a generic way.
+  # Expected errors should be handled in specific handlers.
+  # Flags should be reset here.
+  def handle_info({:EXIT, _pid, {_error, _stacktrace}}, socket) do
```
jfolk2015, line 93: What does :EXIT do? it seems like it kills the process? Is that desired?

Are those flags for showing toasts or on-screen errors to the user?

jfolk2015 (approved):
I have some questions about the generic error handling, but nothing that should block.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
christinepoydence: @jfolk2015 - this is a great question. This handle_info event is boilerplate code that exists in several of the views. From what we can tell, this seems to show an error message on screen to the viewer if it's encountered. (Specifically, there's a small modal that pops up at the bottom of the page that says," A page error occurred.") @LtChae - do you have any additional thoughts here?

PR #1370, opened 2022-04-06
587 access groups users

narruda789:
## Description

Ability to add users to an access group.

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi_web/live/search/manage_users_modal.ex
+++ b/apps/andi/lib/andi_web/live/search/manage_users_modal.ex
@@ -52,11 +52,21 @@ defmodule AndiWeb.Search.ManageUsersModal do
                     <td class="search-table__cell search-table__cell--break search-table__user-name-cell wide-column"><%= user.name %></td>
                     <td class="search-table__cell search-table__cell--break search-table__user-email-cell wide-column"><%= user.email %></td>
                     <td class="search-table__cell search-table__cell--break wide-column"><%= pretty_print_orgs(user.organizations) %></td>
+                    <td class="search-table__cell search-table__cell--break modal-action-text thin-column" phx-click="select-user-search" phx-value-id=<%= user.id %>><%=selected_value(user.id, @selected_users)%></td>
                     <td></td>
                   </tr>
                 <% end %>
               <% end %>
             </table>
+          </div>
+        </div>
+
+        <div class="user-search-selected-users">
+          <p class="search-modal-section-header-text">Selected Users</p>
+          <div class="selected-results-from-search">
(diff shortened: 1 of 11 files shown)


apps/andi/test/unit/andi_web/live/access_group_live_view/user_table_test.exs
```diff
@@ -0,0 +1,102 @@
...
+    auth_conn_case.disable_revocation_list.()
+    :ok
+  end
+
+  describe "Basic associated users table load" do
+    test "shows \"No Associated Users\" when there are no rows to show", %{conn: conn} do
+      access_group = setup_access_group()
+      allow(Andi.Repo.preload(any(), any()), return: %{datasets: [], users: []})
+
+      assert {:ok, view, html} = live(conn, "#{@url_path}/#{access_group.id}")
```
bmitchinson, line 38: non blocking:

is there a technical reason why `live` made more sense here than `live_isolated`?  Hoping to gain more perspective on which one makes sense in what scenarios. (not worth re-writing or anything like that, just curious!)

apps/andi/lib/andi_web/live/search/manage_users_modal.ex
```diff
@@ -75,4 +85,20 @@ defmodule AndiWeb.Search.ManageUsersModal do
     |> Enum.map(fn org -> org.orgTitle end)
     |> Enum.join(", ")
   end
+
+  def selected_users(users, selected_users) do
+    Enum.map(selected_users, fn selected_user ->
+      case Enum.find(users, fn user -> user.id == selected_user end) do
+        nil -> Andi.Schemas.User.get_by_id(selected_user)
+        result -> result
```
bmitchinson, line 93: "if it's already in the search results no need to re-query", nice, very efficient.

bmitchinson (approved):

bmitchinson:
Approved, looked at this locally for a minute and it behaves great. Works just as the dataset search does.

apps/andi/test/unit/andi_web/live/access_group_live_view/user_table_test.exs
christinepoydence: We did this because the event handlers are on the parent component, so live_isolated seemed to not be the best fit!

bmitchinson: Oh yep. I mentioned that that would be a tell that `live` would be necessary when we found `live_isolated`

thanks for the explanation 👍

PR #1371, opened 2022-04-07
check_origin fixes

LtChae:
## Description

- Changed dockerfile to handle openshift runtime requirements
- Added a configurable origin property to support multiple deployment environments

## Reminders:

- [X] If changing elixir code in an "app", did you update the relevant version
      in `mix.exs`?

### DIFF
--- a/apps/andi/mix.exs
+++ b/apps/andi/mix.exs
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "2.1.15",
+      version: "2.1.16",
       build_path: "../../_build",
       config_path: "../../config/config.exs",
       deps_path: "../../deps",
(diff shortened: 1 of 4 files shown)


apps/andi/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "2.1.14",
+      version: "2.1.15",
```
bmitchinson, line 7: I assume CI/CD will flag this soon but, this needs to be .16 I think due to a recent merge

apps/andi/runtime.exs
```diff
@@ -59,7 +59,8 @@ config :andi, :brook,
 config :andi, AndiWeb.Endpoint,
   live_view: [
     signing_salt: live_view_salt
-  ]
+  ],
+  check_origin: [System.get_env("ALLOWED_ORIGIN")]
```
bmitchinson, line 63: Where is this used? Is this a new config value that we'll use later?

bmitchinson (approved):
Approved but curious about that new config value / it's usage

PR #1372, opened 2022-04-07
588 save users to access groups

christinepoydence:
## Description

- Clicking the save button adds all users in the selected user list to the access group
- Clicking the save button returns the user to the edit access group page
- The edit access group page shows a list of all users that are added to the access group
- When navigating back to the search modal, the selected users are shown in the selected box on the bottom

image

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi/schemas/user.ex
+++ b/apps/andi/lib/andi/schemas/user.ex
@@ -43,14 +43,12 @@ defmodule Andi.Schemas.User do
   end
 
   def associate_with_access_group(subject_id, access_group_id) do
-    with user <- Repo.get_by(__MODULE__, subject_id: subject_id) |> Repo.preload(:access_groups) |> IO.inspect(label: "user associate"),
+    with user <- Repo.get_by(__MODULE__, subject_id: subject_id) |> Repo.preload(:access_groups),
          access_group <- AccessGroups.get(access_group_id) do
       user
       |> Repo.preload(:access_groups)
-      |> IO.inspect(label: "before change")
       |> change()
       |> put_assoc(:access_groups, [access_group | user.access_groups])
-      |> IO.inspect(label: "after change")
       |> Repo.update()
     else
       error -> error
--- a/apps/andi/lib/andi/schemas/user.ex
+++ b/apps/andi/lib/andi/schemas/user.ex
@@ -9,6 +9,7 @@ defmodule Andi.Schemas.User do
   alias Andi.InputSchemas.Organization
   alias Andi.InputSchemas.Organizations
   alias Andi.InputSchemas.AccessGroup
+  alias Andi.InputSchemas.AccessGroups
   import Ecto.Query, only: [from: 1]
 
   @primary_key {:id, Ecto.UUID, autogenerate: true}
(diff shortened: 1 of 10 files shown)


apps/andi/lib/andi/schemas/user.ex
```diff
@@ -40,6 +42,21 @@ defmodule Andi.Schemas.User do
     |> Repo.insert_or_update()
   end
 
+  def associate_with_access_group(subject_id, access_group_id) do
+    with user <- Repo.get_by(__MODULE__, subject_id: subject_id) |> Repo.preload(:access_groups) |> IO.inspect(label: "user associate"),
+         access_group <- AccessGroups.get(access_group_id) do
+      user
+      |> Repo.preload(:access_groups)
+      |> IO.inspect(label: "before change")
```
bmitchinson, line 50: remove: inspects left in

apps/andi/test/integration/andi_web/live/access_group_live_view/edit_access_group_live_view_test.exs
```diff
@@ -352,6 +352,51 @@ defmodule AndiWeb.AccessGroupLiveView.EditAccessGroupLiveViewTest do
     end
   end
 
+  describe "emits a user_access_group_associate event" do
```
bmitchinson, line 355: Would it beneficial to add a similar test for `disassociate` events?

apps/andi/test/integration/andi_web/live/access_group_live_view/edit_access_group_live_view_test.exs
```diff
@@ -352,6 +352,51 @@ defmodule AndiWeb.AccessGroupLiveView.EditAccessGroupLiveViewTest do
...
+      assert get_text(html, ".access-groups-sub-table") =~ user.name
+      assert get_text(html, ".access-groups-sub-table") =~ user.email
+
+      # save the changes to the access group
+      save_button = element(view, ".save-edit", "Save")
+      html = render_click(save_button)
+
+      eventually(fn ->
+        access_group = AccessGroups.get(access_group.id) |> Andi.Repo.preload(:users)
+        assert [%Andi.Schemas.User{id: ^user_id}] = access_group.users
```
bmitchinson, line 395: Could we assert that a log in the log events table was also added?

Do we have a testing pattern for that, not sure if it makes sense to add here but, would be good to have tests that ensure they're added somewhere.

bmitchinson (approved):
Approved pending IO inspect removals + some additional comments

apps/andi/test/integration/andi_web/live/access_group_live_view/edit_access_group_live_view_test.exs
christinepoydence: Yes! But that is part of the next card!

apps/andi/lib/andi/schemas/user.ex
christinepoydence: good catch, thanks!

apps/andi/test/integration/andi_web/live/access_group_live_view/edit_access_group_live_view_test.exs
christinepoydence: We can absolutely do this! I'll go ahead and spin up a card for this effort - I don't think it's something that's limited to this event - we probably want to add it for datasets, orgs, ingestions, etc. When I did the initial audit event, I added those tests to the API, but not to the UI.

christinepoydence: Created internal#667 to track this effort.

PR #1374, opened 2022-04-08
589 delete users from access groups

christinepoydence:
## Description

- Clicking the remove button next to a user's name on the edit page removes them from the access group.
- Clicking the remove button next to a user's name on the search modal removes them from the access group
- Clicking the "X" next to a user's name in the selection box removes them from the access group
- I can re-add the user to the access group by clicking the "Add Users" button.

## Reminders:

      in `mix.exs`?


apps/andi/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Andi.MixProject do
   def project do
     [
       app: :andi,
-      version: "2.1.15",
+      version: "2.1.18",
```
jfolk2015, line 7: Typo? Or anticipating version bumps from other PRs?

jfolk2015 (approved):

apps/andi/mix.exs
christinepoydence: Anticipating the version bumps!

PR #1384, opened 2022-04-29
577 access groups discovery api

christinepoydence:
## Description

- When a dataset is added to an access group, all of that group's users can interact with it.
-  They can query for it on discovery ui.
- They can request it through discovery api.
- Removing a dataset from an access group removes the ability for that access group's users to interact with the dataset.
- They cannot search for it on discovery ui after removal.
- They receive a 403 when attempting to request it through discovery api after removal.
- If users have access to that dataset through other groups, then they can still interact with it.
- Apps which listen to access group related events limit the number they keep in memory.

## Reminders:

      in `mix.exs`?


apps/raptor/lib/raptor_web/router.ex
```diff
@@ -8,5 +8,6 @@ defmodule RaptorWeb.Router do
   scope "/api", RaptorWeb do
     pipe_through :api
     get("/authorize", AuthorizeController, :authorize)
+    get("listAccessGroups", ListAccessGroupsController, :list)
```
bmitchinson, line 11: RaptorWeb still confuses me a bit

What utilizes listAccessGroups? It's not a normal route like `/authorize`? (`listAccessGroups` vs `/listAccessGroups`)

apps/discovery_api/lib/discovery_api_web/utilities/model_access_utils.ex
```diff
@@ -8,12 +8,13 @@ defmodule DiscoveryApiWeb.Utilities.ModelAccessUtils do
   use Properties, otp_app: :discovery_api
 
   getter(:raptor_url, generic: true)
+  getter(:raptor_list_url, generic: true)
```
bmitchinson, line 11: This looks unused

apps/discovery_api/lib/discovery_api/search/elasticsearch/query_builder.ex
```diff
@@ -109,19 +110,13 @@ defmodule DiscoveryApi.Search.Elasticsearch.QueryBuilder do
...
-                    }
-                  }
-                ]
+              "terms" => %{
+                "organizationDetails.id" => authorized_organization_ids
+              }
+            },
+            %{
+              "terms" => %{
+                "accessGroups.keyword" => authorized_access_group_ids
```
bmitchinson, line 119: `.keyword` 🪄

apps/discovery_api/lib/discovery_api/event/event_handler.ex
```diff
@@ -124,6 +126,64 @@ defmodule DiscoveryApi.Event.EventHandler do
...
+    end)
+
+    dataset_access_group_disassociate()
+    |> add_event_count(author, relation.dataset_id)
+
+    with {:ok, dataset} <- Brook.get(@instance_name, :models, relation.dataset_id),
+         model <- Mapper.remove_access_group(dataset, relation.access_group_id) do
+      Elasticsearch.Document.update(model)
+
+      Logger.debug(fn ->
```
bmitchinson, line 170: the associate event uses logger.warn, could we unify the two so that the both use `.debug` or both use `.warn`

don't want a false indication that either didn't happen if looking while log level is default

apps/discovery_api/runtime.exs
```diff
@@ -231,3 +233,7 @@ config :telemetry_event,
 if System.get_env("RAPTOR_URL") do
   config :discovery_api, raptor_url: System.get_env("RAPTOR_URL")
 end
+
+if System.get_env("RAPTOR_LIST_URL") do
+  config :discovery_api, raptor_url: System.get_env("RAPTOR_LIST_URL")
+end
```
bmitchinson, line 239: We'll need to add `raptor_list_url` to the discovery api chart as part of this work too right? So that it can be set here like RAPTOR_URL is?

At one point we mentioned potentially refactoring to only need one raptor config URL. Needing to update the chart as a consequence of not doing that makes me think the refactor might be beneficial.

Idea for the refactor:
Could we just give the RaptorService the `raptor_url()` in all cases, and it could append the `/listAccessGroups` or `/authorize` route onto the URL as needed on a method by method basis? If we went with that refactor, we'd ironically have to change the chart anyway because the default raptor endpoint includes `/authorize` and we'd need to remove that to make my refactor suggestion work.

Not attached to any ideas but figured I'd throw that out there.

bmitchinson (approved):
Left a few comments, looks great! Excited to see it in action.

apps/discovery_api/lib/discovery_api_web/utilities/model_access_utils.ex
christinepoydence: good catch! thanks!

apps/discovery_api/lib/discovery_api/event/event_handler.ex
christinepoydence: good point! Set to all the logs to debug to follow the established pattern!

bmitchinson:
@ChristineWalsh-Accenture the raptor postman collection might need to be updated if that's not a hard ask. Hasn't been updated since December https://github.com/UrbanOS-Public/smartcitiesdata/blob/master/apps/raptor/Raptor.postman_collection.json

PR #1385, opened 2022-05-04
Ensuring unhandled exception are dead lettered

jfolk2015:
## Description

Found a bug in Valkyrie where if it didn't recognize certain kinds of data during standardization, it would throw an ArgumentError and not dead letter the message. This forces it to dead letter under any unhandled cases.

## Reminders:

      in `mix.exs`?


bmitchinson (approved):

bmitchinson (approved):
nice, good find

Issue #598, opened 2020-03-19
[documentation] [help wanted] [question] How to run locally?


tommy-elliott-awh:
I've created a fork of this repository (https://github.com/AWHServiceAccount/smartcitiesdata) and I've been trying for a few days to understand this project's parts and to get it to a point where I can run the system locally, but I'm a bit stuck in terms of understanding how to use minikube/helm/kubectl or whatever would be needed in order to get to a usable version of the system and its many components running on my machine using docker images and containers.

The main difficulty is that the main readme documentation for the repository states that to run the system locally, you just need to run `docker pull && docker-compose up -d`, but neither of those commands will execute as described.

`docker pull` needs an argument (the name of the image to pull, with optional tag); and in searching for docker images with smartcitiesdata in their name using `docker search smartcitiesdata` I get results, but trying to pull any of those using `docker pull smartcitiesdata/andi` for example doesn't work because I get
`Using default tag: latest
Error response from daemon: manifest for smartcitiesdata/andi:latest not found: manifest unknown: manifest unknown`
for all of those search results except for smartcitiesdata/data-generator, which pulls successfully.

`docker-compose up -d` needs there to be a docker-compose.yaml file in the root repository directory, which doesn't exist, since a commit from 10/8/19 (https://github.com/smartcitiesdata/smartcitiesdata/commit/65f1a15b) that removed it (commit message: "Remove docker-compose support").

What I'd like to be able to do is to get the system running locally from docker images built from my local repository code so that I can eventually modify the code and have the modifications able to be reflected in what I run locally.  It appears that the local development instructions need to be updated from what exists in the main readme to be correct for how local development actually should be initiated.

I'm developing on a Windows machine, so I set up my environment using Windows Subsystem for Linux (WSL) and I've installed and got to a workable setup all listed needed tools described in the wiki documentation, using that documentation along with finding additional guides online as they relate to using WSL with some of the tools.  I also had to make a few code modifications regarding how git handles CRLF newlines vs LF newlines in *.sh files so they could be executed.

By looking through the scripts folder's contents and interpreting those along with the .travis.yml file in the repository root folder, I was able to execute what seem to be most of the steps involved in building and running tests on the components of the system.

[...]

PR #1398, opened 2022-05-26
598 ingestion step 0 experiment

christinepoydence:
## [Ticket Link #598](https://app.zenhub.com/workspaces/mdot-615b97c1a5fde400126174f8/issues/urbanos-public/internal/598)

image

## Description

-  Ingestion name field is visible and can be updated
- Source format field is visible and can be updated
- Select dataset modal is visible and can be used to select a target dataset
- Fields turn red if they're empty when submit is pressed, since they're required fields.

## Reminders:

      in `mix.exs`?


bmitchinson:
Select datasets modal has `selected datasets` text, needs to be not plural

jfolk2015 (approved):

PR #1399, opened 2022-06-02
541 required extraction fields

jfolk2015:
## Description

Updating smartcitiesdata with new smart city version which makes extraction id fields required. Also makes sure the extraction start time is in iso 8601.

## Reminders:

      in `mix.exs`?


bmitchinson (approved):
Alchemist, Andi, disc_api / streams, Estuary, Forklift, Performance, Raptor, Reaper, Template And Valkyrie all need mix version updates.

probably just went out of sync when dependencies were updated in master

PR #1409, opened 2022-06-27
744 transformations header

christinepoydence:
## Description

- Updating the name field on a transformation updates the text in the rectangle that represents it in step 3

image

## Reminders:

      in `mix.exs`?


bmitchinson (approved):
Good tests! Glad we're unit testing here. LGTM 👍

PR #1413, opened 2022-07-11
753 ignore events which reference non-existent dataset models

jfolk2015:
## Description

Fixing an issue where discovery api would fail to process an associate or disassociate event if the dataset model was missing, and then would not process further events.

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/discovery_api/test/unit/test_helper.exs
+++ b/apps/discovery_api/test/unit/test_helper.exs
@@ -1,2 +1,2 @@
-ExUnit.start(exclude: [:skip])
+ExUnit.start(exclude: [:skip], capture_log: true)
 Faker.start()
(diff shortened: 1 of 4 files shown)


apps/discovery_api/test/unit/test_helper.exs
```diff
@@ -1,2 +1,2 @@
-ExUnit.start(exclude: [:skip])
+ExUnit.start(exclude: [:skip], capture_log: true)
```
bmitchinson, line 1: Should we use this on a test by test basis instead of the entire unit suite?

https://elixirschool.com/blog/til-capture-log-in-exunit-tests

bmitchinson (approved):
Looks great, one comment but non blocking.

PR #1417, opened 2022-07-18
WIP-644 button size

20Isabel:
## Description
- changed the order and orientation of the buttons on each page

## Reminders:

      in `mix.exs`?
-  N/A If altering an API endpoint, was the relevant postman collection updated?
  - N/A If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?


jfolk2015 (approved):
Andi has never looked so good! Thanks for doing this.

PR #1421, opened 2022-07-20
Upgrading to use build image which uses latest alpine keys

jfolk2015:
## Description

All docker builds are broken because alpine changed its signing keys, and we're using a base image which has the old keys. Swapping to use the official hex images for elixir 1.10.4, erlang 23, and alpine 3.16.


bmitchinson (approved):

bmitchinson:
Andi failed due to something with it's image build? Not integration test related.

LtChae (approved):

PR #1427, opened 2022-07-21
Add overwrite_mode to forklift config map

bmitchinson:
## Description

Leaving as a draft until base image issues are resolved.

- Add overwrite_mode to forklift config map
  - Set from system env variable. [PR for this corresponding chart value](https://github.com/UrbanOS-Public/charts/pull/358)

## Reminders:

      in `mix.exs`?
- NA If altering an API endpoint, was the relevant postman collection updated?
  - NA If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?


apps/forklift/runtime.exs
```diff
@@ -147,6 +147,9 @@ if System.get_env("COMPACTION_SCHEDULE") do
     ]
 end
 
+config :forklift,
+  overwrite_mode: !!System.get_env("OVERWRITE_MODE")
```
LtChae, line 151: I have concerns about the readability of this. Notnot in front of a direct environment variable call is a bit opaque. Consider a case block that clearly lays out the possible values of the variable and assigns a boolean value based on it

bmitchinson: 204b775e9165b9715fe6f4e03f40816061e1ded6

Corrected this in the above commit

LtChae (approved):

PR #1430, opened 2022-07-25
New HOME for apps

jfolk2015:
## Description

Updating the home variable, workdir, and permission settings in our services' Dockerfiles so that they work with the new base image.


bmitchinson:
Will approve when this leaves draft state!

bmitchinson (approved):

PR #1441, opened 2022-08-04
745 delete transformation

jfolk2015:
## Description

Curators can delete transformations now!

## Reminders:

      in `mix.exs`?


bmitchinson (approved):
Very well tested 👍

PR #1442, opened 2022-08-04
660 dynamic fields part 1

jfolk2015:
## Description

First step of showing transformation fields on the edit ingestion page. Fields will be shown dynamically based on a transformation-specific config. This covers only showing the fields on the page, it does not cover validation or persistence. The example transformation used is the remove transformation because of its simplicity.

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/live/ingestion_live_view/transformations/transformation_field_builder.ex
```diff
@@ -0,0 +1,25 @@
+defmodule AndiWeb.IngestionLiveView.Transformations.TransformationFieldBuilder do
+  import Phoenix.HTML.Form
+  import Phoenix.LiveView.Helpers
+
+  def build_input(%{field_type: "string"} = field, assigns, form) do
```
LtChae, line 5: Should there be a default case for this for error handling?

apps/transformers/lib/transformation_fields.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Transformers.TransformationFields do
+  alias Transformers
+
+  def fields_for("remove") do
+    Transformers.Remove.fields()
+  end
```
LtChae, line 6: It is possible, though I leave it to you if its a good idea, to reference a module by string:

```
module = String.to_existing_atom("Transformers.#{String.Capitalize(type)}")
module.fields()
```

LtChae: This would remove the need for this step in setting up a new transformation.

LtChae (approved):
General comments

apps/transformers/lib/transformation_fields.ex
jfolk2015: That's good to know. I think I prefer the more explicit way of doing things, since adding an extra head on the hydra is very simple.

apps/andi/lib/andi_web/live/ingestion_live_view/transformations/transformation_field_builder.ex
jfolk2015: Probably just in case. I'll put one in on the next card.

PR #1447, opened 2022-08-16
750 dynamic fields part 2

jfolk2015:
## Description

Andi now validates dynamically created fields for a transformation.

This required extensive changes to the transformers library in order to get all errors for a particular transformation, and to show errors individually by field. The validation function for each transformation now creates an accumulator object which updates with either the field values or errors, and returns either those errors as a map or an ordered list of values for use in the transform function.

Also added a test to alchemist to make sure incorrect transformations would dead letter appropriately.

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transformations/concatenation.ex
+++ b/apps/transformers/lib/transformations/concatenation.ex
@@ -5,6 +5,10 @@ defmodule Transformers.Concatenation do
   alias Transformers.Validations.NotBlank
   alias Transformers.Validations.ValidationStatus
 
+  @source_fields "sourceFields"
+  @target_field "targetField"
+  @separator "separator"
+
   @impl Transformation
   def transform(payload, parameters) do
     with {:ok, [source_fields, separator, target_field]} <- validate(parameters),
@@ -20,10 +24,10 @@ defmodule Transformers.Concatenation do
 
   def validate(parameters) do
     %ValidationStatus{}
-      |> NotBlank.check(parameters, "sourceFields")
-      |> NotBlank.check(parameters, "targetField")
-      |> IsPresent.check(parameters, "separator")
-      |> ValidationStatus.ordered_values_or_errors(["sourceFields", "separator", "targetField"])
+      |> NotBlank.check(parameters, @source_fields)
+      |> NotBlank.check(parameters, @target_field)
+      |> IsPresent.check(parameters, @separator)
+      |> ValidationStatus.ordered_values_or_errors([@source_fields, @separator, @target_field])
   end
 
   def fetch_values(payload, field_names) when is_list(field_names) do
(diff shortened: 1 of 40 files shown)


apps/transformers/lib/transformations/concatenation.ex
```diff
@@ -17,21 +23,20 @@ defmodule Transformers.Concatenation do
...
   end
 
   def fetch_values(payload, field_names) when is_list(field_names) do
     find_values_or_errors(payload, field_names)
     |> all_values_if_present_else_error()
   end
 
-  def fetch_values(_, _), do: {:error, "Expected list but received single value: sourceFields"}
+  def fetch_values(_, _),
+    do: {:error, "Expected list but received single value: #{@source_fields}"}
```
LtChae, line 39: Is this helpful for debugging? Could we instead show the value that didn't match?

apps/transformers/lib/validations/not_blank.ex
```diff
@@ -0,0 +1,56 @@
...
+
+  defp check_if_blank(status, parameters, field) do
+    value = Map.get(parameters, field)
+
+    case value do
+      nil -> ValidationStatus.add_error(status, field, "Missing or empty field")
+      value when is_binary(value) -> check_if_blank_string(status, field, value)
+      value when is_list(value) -> check_if_blank_list(status, field, value)
+      _ -> add_unsupported_type_error(status, field)
+    end
```
LtChae, line 20: Project style usually calls for a function hydra here. Is there a specific reason for a case instead?

LtChae (approved):

apps/transformers/lib/validations/not_blank.ex
jfolk2015: Logically it's the same but I find the case statement more readable. It's easier for me to see that there's branching logic and go to the next step. With a hydra I have to pick out the hydra functions from the rest of the functions in the file, and it's harder to see what the cases are because the guards are in the middle of the function declaration.

Readability is not a one size fits all approach though, so other people may feel differently.

apps/transformers/lib/transformations/concatenation.ex
jfolk2015: The thing I changed here was using the new module attribute/string constant consistently throughout the file, and then mix format changed the indentation. I'd prefer to leave it for now, because I think giving the payload validations a refactoring when we have time would be beneficial.

apps/transformers/lib/validations/not_blank.ex
LtChae:
```elixir
defp check_if_blank(status, parameters, field) do
  parameters
  |> Map.get(field)
  |> apply_validation()
end  

defp apply_validation(status, field, value) when is_nil(value), do: ValidationStatus.add_error(status, field, "Missing or empty field")

defp apply_validation(status, field, value) when is_binary(value), do: check_if_blank_string(status, field, value)

defp apply_validation(status, field, value) when is_list(value), do: check_if_blank_list(status, field, value)

defp apply_validation(_), do: add_unsupported_type_error(status, field)
```

I think I agree with you here, with the number of parameters that would need to be passed there are some benefits to being in the same scope too.

PR #1461, opened 2022-09-07
553 add

c-m-duncan:
## Description

- Added arithmetic_add.ex

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transformations/arithmetic_add.ex
+++ b/apps/transformers/lib/transformations/arithmetic_add.ex
@@ -4,16 +4,15 @@ defmodule Transformers.ArithmeticAdd do
   alias Transformers.Validations.NotBlank
   alias Transformers.Validations.ValidationStatus
 
-  @impl Transformation
-
   @addends "addends"
   @target_field "targetField"
 
+  @impl Transformation
   def transform(payload, parameters) do
     with {:ok, [addends, target_field]} <- validate(parameters),
          {:ok, value} <- sumValues(addends, payload) do
-      modified_payload = Map.put(payload, target_field, value)
-      {:ok, modified_payload}
+      transformed_payload = Map.put(payload, target_field, value)
+      {:ok, transformed_payload}
     else
       {:error, reason} -> {:error, reason}
     end
--- a/apps/transformers/lib/transformations/arithmetic_add.ex
+++ b/apps/transformers/lib/transformations/arithmetic_add.ex
@@ -45,9 +45,9 @@ defmodule Transformers.ArithmeticAdd do
         options: nil
       },
       %{
-        field_name: "targetField",
+        field_name: "addends",
         field_type: "list",
-        field_label: "List of values or fields to add to targetField",
+        field_label: "List of values or fields to add together",
         options: nil
       }
     ]
--- a/apps/transformers/lib/transformations/arithmetic_add.ex
+++ b/apps/transformers/lib/transformations/arithmetic_add.ex
@@ -11,7 +11,7 @@ defmodule Transformers.ArithmeticAdd do
   def transform(payload, parameters) do
     with {:ok, [addends, target_field]} <- validate(parameters),
          {:ok, sum} <- sumValues(addends, payload) do
-      {:ok, payload |> Map.put(target_field, sum)}
+         {:ok, payload |> Map.put(target_field, sum)}
     else
       {:error, reason} -> {:error, reason}
     end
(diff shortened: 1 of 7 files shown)


c-m-duncan:
`If changing elixir code in an "app", did you update the relevant version
in mix.exs?` <- Can someone explain this a little bit to us?

apps/transformers/test/unit/transformations/arithmetic_add_test.exs
```diff
@@ -0,0 +1,168 @@
+defmodule Transformers.ArithmeticAddTest do
+  use ExUnit.Case
+
+  alias Transformers.ArithmeticAdd
+
+  describe "transform/2" do
```
c-m-duncan, line 6: We could probably delete some of these tests now

apps/transformers/lib/transformations/arithmetic_add.ex
```diff
@@ -0,0 +1,61 @@
...
+         {:ok, value} <- sumValues(addends, payload) do
+      modified_payload = Map.put(payload, target_field, value)
+      {:ok, modified_payload}
+    else
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp sumValues(addends, payload) do
+    Enum.reduce_while(addends, {:ok, 0}, fn addend, {:ok, acc} ->
```
c-m-duncan, line 21: thought this reduce_while was pretty slick, but I don't love that we had to use multiple `cond` in here

jfolk2015: Yeah, I haven't found a concise Elixir counterpart to something like a ternary. You could try to do a one-liner such as `result = if is_number(addend), do: addend, else: Map.get(payload, addend)`. Not sure if that's more or less readable. Perhaps if you did that as part of a map before hand, then passed that in instead of the raw addends to the reduce_while?

apps/transformers/lib/transformations/arithmetic_add.ex
```diff
@@ -0,0 +1,59 @@
...
+      %{
+        field_name: "targetField",
+        field_type: "string",
+        field_label: "Field to populate with sum",
+        options: nil
+      },
+      %{
+        field_name: "targetField",
+        field_type: "list",
+        field_label: "List of values or fields to add to targetField",
```
jfolk2015, line 54: Since the labels end up on the UI, could I recommend something like "List of values or fields to add together"? Or include target field somehow in the label for that field. Just so it's clear to a user how those two relate. Hopefully it's obvious but DX isn't my strength.

apps/transformers/test/unit/transformations/arithmetic_add_test.exs
jfolk2015: If you like. Nothing particularly stood out to me as being redundant.

jfolk2015 (approved):

bmitchinson:
> `If changing elixir code in an "app", did you update the relevant version in mix.exs?` <- Can someone explain this a little bit to us?

Yep! Because code in `apps/transformers` was changed, we ask that you up the version number located in the mix.exs file for that application. [Transformers mix file L7](https://github.com/UrbanOS-Public/smartcitiesdata/blob/00c9fceae4ac615b310ea59e0f7d85d369d41e2f/apps/transformers/mix.exs#L7)

We use [symver](https://semver.org/) generally, so the `1.0.1` that's there now, I'd suggest would receive a patch bump. That line could be changed to `1.0.2`, and that would fulfill the check box listed on the PR todos.

apps/transformers/lib/transformations/arithmetic_add.ex
c-m-duncan: My logic for doing the map inside of the while was so we only had to loop through the payload once. 

I'm not sure how huge the datasets can get, or if we care about performance to such an extent that we care about things like that.

c-m-duncan:
> Transformers mix file L7

Cool, thanks. I saw that one file, but was worried I was gonna miss somewhere else that also needed an update - so we just waited to get confirmation before changing anything

apps/transformers/lib/transformations/arithmetic_add.ex
jfolk2015: Keeping performance in mind is a good idea. Each message payload can be arbitrarily large, though from what I've seen they're usually not bad. But that's just what we've seen not a guarantee.

I'm good with it either way. If you feel the readability could be better, extracting some well named private functions for each conditional might help. Up to you.

c-m-duncan: we (read: Ian) made it a lot more slick 👍

PR #1464, opened 2022-09-07
subtract transformation

ian-j-abbott-accenture:
## Description

- Implement subtract transformation

## Reminders:

- [ X] If changing elixir code in an "app", did you update the relevant version
      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transformations/arithmetic_subtract.ex
+++ b/apps/transformers/lib/transformations/arithmetic_subtract.ex
@@ -1,0 +1,70 @@
+defmodule Transformers.ArithmeticSubtract do
+  @behaviour Transformation
+
+  alias Transformers.Validations.IsPresent
+  alias Transformers.Validations.NotBlank
+  alias Transformers.Validations.ValidationStatus
+
+  @minuend "minuend"
+  @subtrahends "subtrahends"
+  @target_field "targetField"
+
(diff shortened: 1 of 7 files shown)


apps/transformers/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Transformers.MixProject do
   def project do
     [
       app: :transformers,
-      version: "1.0.2",
+      version: "1.0.4",
```
ian-j-abbott-accenture, line 7: Jumping straight to 4 assuming the multiplication PR will go out before ours

apps/transformers/lib/transformations/arithmetic_subtract.ex
```diff
@@ -0,0 +1,71 @@
...
+      {:ok, payload |> Map.put(target_field, difference)}
+    else
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp subtractValues(minuend, subtrahends, payload) do
+    minuendValue = Map.get(payload, minuend, minuend)
+
+    if is_number(minuendValue) do
```
ian-j-abbott-accenture, line 25: This can likely be cleaned up if we have a validate_parameters and validate_payload in the future.

jfolk2015 (approved):

PR #1465, opened 2022-09-08
514 division transformation

nicholas-goodwin:
## Description

- Added division to Arithmetic Transformations

## Reminders:

      in `mix.exs`?


apps/transformers/mix.exs
```diff
@@ -34,7 +34,8 @@ defmodule Transformers.MixProject do
       {:placebo, "~> 2.0.0-rc2", only: [:dev, :test, :integration]},
       {:smart_city, "~> 5.2.5"},
       {:smart_city_test, "~> 2.2.5", only: [:test, :integration]},
-      {:timex, "~> 3.6"}
+      {:timex, "~> 3.6"},
+      {:decimal, "~> 1.9.0"}
```
bmitchinson, line 38: Let's change this to `{:decimal, "~> 1.0"},` to align 1:1 with discovery streams

If we upgrade this dependency in either app, the upgrade will look the same in both places

bmitchinson (approved):
👍 💯 🔥  looks good to me !

PR #1466, opened 2022-09-08
751 save dynamic fields

jfolk2015:
## Description

Saving transformations. Right now only remove is wired up to the UI.

## Reminders:

      in `mix.exs`?


apps/andi/README.md
```diff
@@ -72,6 +72,14 @@ For integration testing, Andi encapsulates its external dependencies in Docker i
...
 `mix test.integration`
 
+Individual files can be run using the following command:
+
+`mix test.integration /path/to/file.exs`
+
+Individual tests can be run by adding the test's line number to the end of the file:
+
+`mix test.integration /path/to/file.exs:10`
+
```
bmitchinson, line 82: 🎉

apps/andi/assets/webpack.config.js
```diff
@@ -55,5 +55,6 @@ module.exports = (env, options) => ({
   plugins: [
     new MiniCssExtractPlugin({ filename: '../css/app.css' }),
     new CopyWebpackPlugin({patterns: [{ from: 'static/', to: '../' }]})
-  ]
+  ],
+  stats: 'errors-only',
```
bmitchinson, line 59: does this shorten the startup message to only print if there's an error?

apps/andi/lib/andi_web/live/ingestion_live_view/transformations/transformations_step.ex
```diff
@@ -3,17 +3,20 @@ defmodule AndiWeb.IngestionLiveView.Transformations.TransformationsStep do
   LiveComponent for organizing individual transformation configurations
   """
   use Phoenix.LiveView
+
+  # use AndiWeb.FormSection, schema_module: AndiWeb.InputSchemas.TransformationFormSchema
```
bmitchinson, line 7: comment removal

apps/andi/assets/webpack.config.js
jfolk2015: Yes, it specifically gets rid of all the webpack output which can flood the logs during tests. I don't think this was always printing out before, but with some of the dependency upgrades it became more problematic.

apps/andi/lib/andi_web/live/ingestion_live_view/transformations/transformation_form.ex
```diff
@@ -51,32 +52,25 @@ defmodule AndiWeb.IngestionLiveView.Transformations.TransformationForm do
...
 
-  def handle_event("move-transformation", %{"id" => transformation_id, "move-index" => move_index}, socket) do
-    AndiWeb.Endpoint.broadcast_from(self(), "move-transformation", "move-transformation", %{
-      "id" => transformation_id,
-      "move-index" => move_index
-    })
-
-    {:noreply, socket}
-  end
-
```
bmitchinson, line 75: I'm having trouble understanding why `move-transformation` was removed here. Is it because now a move makes a change to the changeset, and when the updated changeset is re-rendered, the order of transformations is updated as part of that? So it no longer needs an explicit event handler?

Even though the handle event for `move-transformation` is removed here, it looks like `move-transformation` is still referenced in the phx-click on like 40. Does that need to be removed as well?

bmitchinson: Wow okay kuddos for you for picking this up as one of your first cards this event stuff is a lot.

The `move-transformation` handler was removed here because it is now present in the `transformations_step` component, indicated by the form element that sends the event targeting `#transformations-form`

`#transformations-form` being the top level ID in the `transformations_step` component threw me off a bit. I think unifying those names would require an unfortunate amount of changes so, I'll make a refactor card to align those and that might help readability.

apps/andi/lib/andi_web/live/ingestion_live_view/transformations/transformations_step.ex
```diff
@@ -164,6 +177,21 @@ defmodule AndiWeb.IngestionLiveView.Transformations.TransformationsStep do
...
+
+    all_changesets_valid = Enum.map(updated_changesets, fn changeset -> changeset.valid? end) |> Enum.all?()
+
+    cond do
+      visibility == "expanded" -> "expanded"
+      all_changesets_valid -> "valid"
+      true -> "invalid"
+    end
+  end
+
```
bmitchinson, line 194: 👏👏👏

PR #1473, opened 2022-09-12
feat(830): change all key atoms to string in payload on transformation creation

c-m-duncan:
## [Ticket Link #830](https://app.zenhub.com/workspaces/mdot-615b97c1a5fde400126174f8/issues/urbanos-public/internal/830)

      in `mix.exs`

### DIFF
--- a/apps/alchemist/test/integration/alchemist_test.exs
+++ b/apps/alchemist/test/integration/alchemist_test.exs
@@ -56,7 +56,7 @@ defmodule AlchemistTest do
     ]
 
     input_topic = "#{input_topic_prefix()}-#{ingestion.id}"
-    output_topic = "#{output_topic_prefix()}-#{ingestion.id}"
+    output_topic = "#{output_topic_prefix()}-#{dataset.id}"
 
     Brook.Event.send(@instance_name, ingestion_update(), :alchemist, ingestion)
     TestHelpers.wait_for_topic(elsa_brokers(), input_topic)
@@ -73,7 +73,7 @@ defmodule AlchemistTest do
     eventually fn ->
       post_transform_messages = TestHelpers.get_data_messages_from_kafka_with_timing(output_topic, elsa_brokers())
 
-      assert pre_transform_messages = post_transform_messages
+      assert pre_transform_messages == post_transform_messages
     end
   end
 
@@ -99,7 +99,7 @@ defmodule AlchemistTest do
       Enum.map(pre_transform_messages, fn message -> Map.put(message, :dataset_id, dataset.id) end)
 
     input_topic = "#{input_topic_prefix()}-#{ingestion.id}"
-    output_topic = "#{output_topic_prefix()}-#{ingestion.id}"
+    output_topic = "#{output_topic_prefix()}-#{dataset.id}"
 
     Brook.Event.send(@instance_name, ingestion_update(), :alchemist, ingestion)
     TestHelpers.wait_for_topic(elsa_brokers(), input_topic)
@@ -114,7 +114,7 @@ defmodule AlchemistTest do
     eventually fn ->
       post_transform_messages = TestHelpers.get_data_messages_from_kafka_with_timing(output_topic, elsa_brokers())
 
-      assert pre_transform_messages = post_transform_messages
+      assert pre_transform_messages == post_transform_messages
     end
   end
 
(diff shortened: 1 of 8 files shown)


apps/alchemist/test/unit/alchemist/broadway_test.exs
```diff
@@ -32,9 +32,9 @@ defmodule Alchemist.BroadwayTest do
         TDG.create_transformation(%{
           type: "regex_extract",
           parameters: %{
-            "sourceField" => "first_name",
-            "targetField" => "first_letter",
-            "regex" => "^(\\w)"
+            sourceField: "first_name",
+            targetField: "first_letter",
+            regex: "^(\\w)"
```
c-m-duncan, line 37: We changed this test in alchemist, then realized the fix probably worked better in transformers. I still think this test is valid and covers it, so we didn't move it 🦐 

Do we need to change the alchemist version for this?

RyanRConaway: Version wouldn't need to be bumped since the tests are not packaged in the binary.

I'm cool with leaving this here as well since the idea is we can handle either case

RyanRConaway: Is there a test case that uses string for the keys as well?

apps/transformers/lib/transform_actions.ex
```diff
@@ -4,7 +4,8 @@ defmodule Transformers do
   def construct(transformations) do
     Enum.map(transformations, fn transformation ->
       with {:ok, type} <- Map.fetch(transformation, :type),
-           {:ok, parameters} <- Map.fetch(transformation, :parameters) do
+           {:ok, raw_parameters} <- Map.fetch(transformation, :parameters),
+           parameters <- SmartCity.Helpers.to_string_keys(raw_parameters) do
```
bmitchinson, line 8: 🥳

apps/alchemist/test/unit/alchemist/broadway_test.exs
bmitchinson: Because of the way test_data_generator works, this change can be reverted I think. No matter if the parameters are provided with strings or atoms as keys, the result will have entires with atoms as keys.

``` elixir
iex(1)> string_keys =
...(1)>   SmartCity.TestDataGenerator.create_transformation(%{
...(1)>     type: "regex_extract",
...(1)>     parameters: %{
...(1)>       "sourceField" => "first_name",
...(1)>       "targetField" => "first_letter",
...(1)>       "regex" => "^(\\w)"
...(1)>     }
...(1)>   })
%SmartCity.Ingestion.Transformation{
  id: "b4f3265f-6400-48c6-8de8-d1682eba0c5f",
  name: "Transformation Name",
  parameters: %{
    :regex => "^\\((\\d{3})\\)",
    :sourceField => "Sooty",
    :targetField => "Lucy",
    "regex" => "^(\\w)",
    "sourceField" => "first_name",
    "targetField" => "first_letter"
  },
  sequence: 1..1000,
  type: "regex_extract"
}
iex(2)>
nil
iex(3)> atom_keys =
...(3)>   SmartCity.TestDataGenerator.create_transformation(%{
...(3)>     type: "regex_extract",
...(3)>     parameters: %{
...(3)>       sourceField: "first_name",
...(3)>       targetField: "first_letter",
...(3)>       regex: "^(\\w)"
...(3)>     }
...(3)>   })
%SmartCity.Ingestion.Transformation{
  id: "3fe38690-9e29-40dd-bea1-4f5699e59cf7",
  name: "Transformation Name",
  parameters: %{
    regex: "^(\\w)",
    sourceField: "first_name",
    targetField: "first_letter"
  },
  sequence: 1..1000,
  type: "regex_extract"
}
```

apps/alchemist/test/integration/alchemist_test.exs
```diff
@@ -56,7 +56,7 @@ defmodule AlchemistTest do
     ]
 
     input_topic = "#{input_topic_prefix()}-#{ingestion.id}"
-    output_topic = "#{output_topic_prefix()}-#{ingestion.id}"
+    output_topic = "#{output_topic_prefix()}-#{dataset.id}"
```
bmitchinson, line 59: oh great change here, this is much more system accurate since the output is a dataset. thank you 👍

bmitchinson (approved):

bmitchinson:
Approved assuming that messages are sent fine with the alchemist readme "[example transformations](https://github.com/UrbanOS-Public/smartcitiesdata/tree/master/apps/alchemist#example-transformations)"

Transformation cards should be ready for any dev to exit after this card.

PR #1474, opened 2022-09-13
689 Overwrite mode + Compact after each data ingestion

bmitchinson:
## Description

- Enabling overwrite mode only retains the most recent ingestion per ingestion_id in the system
- Compaction occurs no longer on a specified cadence, but automatically when a data ingestion is complete.
- Adds `docs/compaction.md` detailing the compaction process for developers
- Added json cleanup step to nightly compaction to clear json data that failed compaction

## Reminders:

      in `mix.exs`?
- [ ] ~~If altering an API endpoint, was the relevant postman collection updated?~~
  - [ ] ~~If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?~~


nicholas-goodwin (approved):
looking good out there besides some tests failing

PR #1477, opened 2022-09-16
Added regex transformation test

RyanRConaway:
## Description

- ...

## Reminders:

      in `mix.exs`?


bmitchinson (approved):
Nice! I was gonna create a whole new schema / ingestion, that would have been a waste. I like reusing the current ones better. Approved.

PR #1479, opened 2022-09-20
feat(789): whitelist img-src for andi logo

c-m-duncan:
## [Ticket Link #789](https://app.zenhub.com/workspaces/mdot-615b97c1a5fde400126174f8/issues/urbanos-public/internal/789)

Screen Recording 2022-09-20 at 3 40 52 PM

### DIFF
--- a/apps/andi/lib/andi/application.ex
+++ b/apps/andi/lib/andi/application.ex
@@ -126,7 +126,7 @@ defmodule Andi.Application do
   def get_logo_url() do
     env_url = get_env_variable("ANDI_LOGO_URL", false)
 
-    case is_nil(env_url) do
+    case is_invalid_env_variable(env_url) do
       false -> env_url
       true -> "/images/UrbanOS.svg"
     end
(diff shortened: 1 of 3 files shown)


bmitchinson (approved):
Approved with the blocking comment that the mix version is updated.

We have to up the version with a patch, even for adjustments like this, to make sure images are built and pushed correctly when changes are merged.

c-m-duncan:
I seem to be having some caching issues, changing my env still has the same content policy as before. Not sure how, since I'm restarting everything locally and doing a `Clear cache and hard reload`. Anyone have any ideas as to why / how it's happening?

c-m-duncan:
> Approved with the blocking comment that the mix version is updated.
> 
> We have to up the version with a patch, even for adjustments like this, to make sure images are built and pushed correctly when changes are merged.

Ty, totally forgot about the version bump

bmitchinson:
Due to recent merging, you'll need to update the mix version again

PR #1480, opened 2022-09-21
841 configurable title

ijabbott:
## Description

- Allow all UrbanOS instances to configure their title

Screen Shot 2022-09-21 at 9 22 12 AM

## Reminders:

      in `mix.exs`?
- [x] ~~If altering an API endpoint, was the relevant postman collection updated?~~
  - [x] ~~If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?~~


apps/andi/lib/andi/application.ex
```diff
@@ -132,6 +133,15 @@ defmodule Andi.Application do
     end
   end
 
+  def get_header_text() do
+    env_url = get_env_variable("ANDI_HEADER_TEXT", false)
+
+    case is_invalid_env_variable(env_url) do
```
ijabbott, line 139: Question here: 

Do we want to have default values like this? is_invalid_env_variable already throws an error if the variable doesn't exist, and I imagine the expected default is defined in the charts repo. 

Not sure if having default values here introduces uncertainty towards what to expect when an environment variable is missing.

jfolk2015: I do think some default value is useful. On the one hand, I like the idea of minimizing the chance of errors by defining the default here as well. On the other hand, this could lead to someone down the line wondering where this header text is coming from and having to search for it.

I think it's probably pretty easy to find though, so I'm currently leaning towards minimizing the chance of errors. What do you think?

jfolk2015 (approved):

PR #1482, opened 2022-09-26
Changed publishing strategy to push image matching elixir version and…

RyanRConaway:
Changed publishing strategy to push image matching elixir version and tag as development

## Description

Changed publishing strategy to push image matching elixir version and tag as development

## Reminders:

      in `mix.exs`?

### DIFF
--- a/scripts/gh-action-publish.sh
+++ b/scripts/gh-action-publish.sh
@@ -3,6 +3,21 @@
 set -e
 
 app="${1}"
+echo "app:"
+printf "%s \n\n" "$app"
 
-./scripts/build.sh $app development false
-./scripts/publish.sh $app development
+tag_as_development="${2:-true}"
+echo "tag_as_development:"
+printf "%s \n\n" "$tag_as_development"
+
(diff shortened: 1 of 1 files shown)


scripts/gh-action-publish.sh
```diff
@@ -3,6 +3,21 @@
...
+
+mix_vsn=$(mix cmd --app $app mix app.version | tail -1)
+echo "mix_vsn:"
+printf "%s \n\n" "$mix_vsn"
+
+./scripts/build.sh $app $mix_vsn false
+./scripts/publish.sh $app $mix_vsn
+
+if [[ $tag_as_development == true ]]; then
+  docker tag smartcitiesdata/$app:$mix_vsn smartcitiesdata/$app:deployment
```
jfolk2015, line 21: Should the tag be deployment? I was expecting it to be development.

RyanRConaway: Ah, yes, good catch.

jfolk2015 (approved):

PR #1500, opened 2022-10-17
879 missing icons

jfolk2015:
## Description

Restoring icons which aren't material icons throughout Andi. Cleaning up some changes from self-hosting.

## Reminders:
  ~- [] If updating Major or Minor versions , did you update the sauron chart configuration?~
      in `mix.exs`?
~- [] If altering an API endpoint, was the relevant postman collection updated?~
  ~- [] If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?~


nicholas-goodwin (approved):

apps/andi/endpoint.ex
```diff
@@ -1,9 +0,0 @@
-defmodule Andi.Endpoint do
-  use Phoenix.Endpoint, otp_app: :andi
-
-  plug Plug.Static,
-    at: "/",
-    from: :andi,
-    gzip: false,
-    only: ~w(css fonts images js favicon.ico)
-end
```
c-m-duncan, line 9: Oh so do we not need this since we're now hosting the font ourself?

c-m-duncan (approved):

apps/andi/endpoint.ex
jfolk2015: No, we do need it. But it already exists under apps/andi/lib/andi_web/endpoint.ex.

PR #1503, opened 2022-10-18
Can tab into all header buttons and "click" them with enter or space

jfolk2015:
## Description

Improvements for screen readers

- Can tab into header buttons
- Enter and space function like clicks

I tried to make the code for this as "dry" as possible, but guard clauses have limitations I couldn't get around. For example I wanted to make the guard clause a function or the module attribute a list of all three permitted actions, but neither compiles. Happy to hear suggestions for better ways to do this.

## Reminders:
  ~- [ ] If updating Major or Minor versions , did you update the sauron chart configuration?~
      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi_web/live/header_live_view.ex
+++ b/apps/andi/lib/andi_web/live/header_live_view.ex
@@ -34,30 +34,30 @@ defmodule AndiWeb.HeaderLiveView do
       <span class="page-header__primary" phx-click="show-datasets">
         <img id="header-logo" src=<%= get_logo() %>></img>
         <span><%= get_header_text() %></span>
-        <span class="log-out-link primary-color-text" phx-click="log-out">
+        <span id="log-out-link" class="log-out-link primary-color-text" phx-click="log-out" phx-keydown="log-out" tabindex="0">
           <span class="material-icons">person</span>
           <span class="log-out-link__text">Log Out</span>
         </span>
       </span>
       <span class="page-header__secondary">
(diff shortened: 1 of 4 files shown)


apps/andi/lib/andi_web/live/header_live_view.ex
```diff
@@ -71,29 +71,57 @@ defmodule AndiWeb.HeaderLiveView do
...
+      def handle_event("show-datasets", key, socket) when key in [@click, @enter, @space] do
         AndiWeb.HeaderLiveView.__redirect__(socket, header_datasets_path())
       end
 
-      def handle_event("show-organizations", _, socket) do
+      def handle_event("show-datasets", _, socket) do
+        {:noreply, socket}
+      end
+
+      def handle_event("show-organizations", key, socket) when key in [@click, @enter, @space] do
```
c-m-duncan, line 86: Elixir is so funky. Also, space?

apps/andi/test/unit/andi_web/live/header_live_view_test.exs
```diff
@@ -2,8 +2,6 @@ defmodule AndiWeb.HeaderLiveViewTest do
   use AndiWeb.Test.AuthConnCase.UnitCase
   use Placebo
 
-  alias AndiWeb.HeaderLiveView
```
c-m-duncan, line 5: unused?

c-m-duncan (approved):

PR #1517, opened 2022-11-02
added error result to http responses

nicholas-goodwin:
## Description

- added an error handling portion for when a website will return :ok, but with an error message, which seems to break ingestion consumption.

## Reminders:
  ~~- [ ] If updating Major or Minor versions , did you update the sauron chart configuration?~~
      in `mix.exs`?

### DIFF
--- a/apps/reaper/lib/reaper/data_slurper/http.ex
+++ b/apps/reaper/lib/reaper/data_slurper/http.ex
@@ -30,6 +30,7 @@ defmodule Reaper.DataSlurper.Http do
       Logger.error(fn ->
         "Unable to retrieve data for #{ingestion_id}: #{Exception.message(error)}"
       end)
+
       reraise inspect(error), __STACKTRACE__
   end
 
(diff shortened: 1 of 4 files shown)


apps/reaper/lib/reaper/data_slurper/http.ex
```diff
@@ -31,7 +31,7 @@ defmodule Reaper.DataSlurper.Http do
         "Unable to retrieve data for #{ingestion_id}: #{Exception.message(error)}"
       end)
 
-      reraise error, __STACKTRACE__
+      reraise inspect(error), __STACKTRACE__
```
nicholas-goodwin, line 34: this came about because the HTTPError that was boiling up from the Task.Await on line 39 ended up like {{%HTTPError}} for some reason with no real way of removing it without erroring out further when another exception came through. Would gladly change this if solution could be found.

bmitchinson (approved):

PR #1518, opened 2022-11-03
904 missing icons

jfolk2015:
## Description

Restoring icons for extract step test info, transformation edit, and transformation delete. Http extract step test button uses primary color, and access group's manage users save button matches the styles for its managed datasets save button.

## Reminders:
  ~- [ ] If updating Major or Minor versions , did you update the sauron chart configuration?~
      in `mix.exs`?


apps/andi/assets/package.json
```diff
@@ -6,7 +6,6 @@
   },
   "dependencies": {
     "cacache": "^15.2.0",
-    "material-icons": "^1.11.8",
```
c-m-duncan, line 9: There's kind of a lot of mentions of material icons in the code still, sure this removal is okay? Looks like it's just what we call the classes for css (and one spot that mentions it for a font `font-family: 'Material Icons';`), but figure'd I'd ask

c-m-duncan (approved):

apps/andi/assets/package.json
jfolk2015: Should no longer be necessary because we're self-hosting the icons. The material-icons package was supposed to be a way to self host the icons without doing all the work ourselves, but I couldn't get it working. If you want to double check that it works, you can checkout my branch, go into apps/andi/assets and run and npm clean-install. Then start up andi as normal and it should still show all the material icons, such as on the transformation form header, etc.

PR #1519, opened 2022-11-04
increase wait time for brook to be ready

ijabbott:
## Description

- increase wait time for brook to be ready in an attempt to decrease integration test flakyness

## Reminders:
      in `mix.exs`?


apps/discovery_api/test/utils/test_helper.ex
```diff
@@ -92,7 +92,7 @@ defmodule DiscoveryApi.Test.Helper do
   end
 
   def wait_for_brook_to_be_ready() do
-    Process.sleep(5_000)
+    Process.sleep(20_000)
```
c-m-duncan, line 95: ideally this function could check something in brook to see if it's happy yet

c-m-duncan (approved):

apps/discovery_api/test/utils/test_helper.ex
ijabbott: 💯 If this works, that would be the ideal scenario.

PR #1527, opened 2022-11-10
Schedule job text color

ijabbott:
## Description

Old:
Screen Shot 2022-11-10 at 8 12 07 AM

New:
Screen Shot 2022-11-10 at 8 11 56 AM


nicholas-goodwin (approved):

bmitchinson:
Not worth holding up the card but while you're in here, would it be possible to fix the alignment of the radio select labels with the header?
image

PR #1529, opened 2022-11-11
Close dataset modal

ijabbott:
## Description

Screen Shot 2022-11-10 at 5 42 23 PM


c-m-duncan (approved):
Events and sockets are so foreign to me still. Glad we’re collectively getting more experience with them

PR #1536, opened 2022-11-15
Protected against org title being nil

RyanRConaway:
## Description

- Protected against org title being nil

## Reminders:
      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi/input_schemas/organization.ex
+++ b/apps/andi/lib/andi/input_schemas/organization.ex
@@ -64,13 +64,7 @@ defmodule Andi.InputSchemas.Organization do
     id = Ecto.Changeset.get_field(changeset, :id)
     org_name = Ecto.Changeset.get_field(changeset, :orgName)
 
-    safe_org_name =
-      case org_name do
-        nil -> ""
-        x -> x
-      end
-
-    case Organizations.is_unique?(id, safe_org_name) do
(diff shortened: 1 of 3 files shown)


nicholas-goodwin (approved):

apps/andi/lib/andi/input_schemas/organization.ex
```diff
@@ -64,7 +64,13 @@ defmodule Andi.InputSchemas.Organization do
...
     org_name = Ecto.Changeset.get_field(changeset, :orgName)
 
-    case Organizations.is_unique?(id, org_name) do
+    safe_org_name =
+      case org_name do
+        nil -> ""
+        x -> x
+      end
+
+    case Organizations.is_unique?(id, safe_org_name) do
```
bmitchinson, line 73: Instead of manipulating the input parameter, can we create a hyrdra head to `Organizations.is_unique?`

```elixir
# Organizations schema file or wherever this function is implemented
# ... other `is_unique?`...

def is_unique?(_id, nil) do
  false
end

# ...
```

bmitchinson (approved):

apps/andi/lib/andi/input_schemas/organization.ex
RyanRConaway: Nice, I like it. I'll make that change

PR #1539, opened 2022-11-17
XML with topLevelSelector

bmitchinson:
## Description

- topLevelSelector now shows in ingestion metadata
  - this enables xml ingestions
  - tests that used to test this on datasets have been moved to ingestions
  - topLevelSelector validation with Jaxon is restored from pre-ingestion introduction

- DataDictionaryFieldEditor used to show topLeveISelector field when editing a dataset. This was meaningless, as the field would never have any effect. It is now removed for dataset edits but enabled for ingestions.
  - It used to be marked as required always, despite being disabled when non xml sourceFormats were selected. It's required `*` symbol now aligns when it is disabled or active.

- Ingestion sourceFormat cannot be altered once an ingestion has been published.
  - This is a requirement of the urbanos system, that was not being enforced by the UI. The test to ensure formats cannot be altered is reinstated.

Bonus:

- Default auth0 tenant for local dev to the proper UrbanOS-dev tenant, utilized currently by product. Not the AIP hercules environment. Once merged, will need to update your local auth0 secret value, will notify the team when that happens
- Adjusted finalize form radio options to align with header, like other inputs in the section

## Reminders:
  - [ ] ~~If updating Major or Minor versions , did you update the sauron chart configuration?~~
      in `mix.exs`?
- [ ] ~~If altering an API endpoint, was the relevant postman collection updated?~~
  - [ ] ~~If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?~~

## Images:

| data from xml, json, json with selector all in trino |
| --- |
| image |

| ingestions for those three defined in andi |
| --- |
| image |

### DIFF
--- a/apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
+++ b/apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
@@ -10,12 +10,14 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     field(:name, :string)
     field(:sourceFormat, :string)
     field(:targetDataset, :string)
+    field(:topLevelSelector, :string)
   end
 
   @cast_fields [
     :name,
     :sourceFormat,
-    :targetDataset
+    :targetDataset,
+    :topLevelSelector
   ]
 
   @required_fields [
@@ -30,6 +32,7 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     current
     |> cast(changes, @cast_fields, empty_values: [])
     |> validate_required(@required_fields, message: "is required")
+    |> validate_top_level_selector()
     |> target_dataset_exists()
   end
 
@@ -42,8 +45,13 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
   defp target_dataset_exists(changeset) do
     validate_change(changeset, :targetDataset, fn :targetDataset, targetDataset ->
       case Andi.InputSchemas.Datasets.get(targetDataset) do
-        nil -> [targetDataset: "Dataset with id: #{targetDataset} does not exist. It may have been deleted."]
-        _ -> []
+        nil ->
+          [
+            targetDataset: "Dataset with id: #{targetDataset} does not exist. It may have been deleted."
+          ]
+
+        _ ->
+          []
       end
     end)
   end
@@ -54,4 +62,19 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     |> AtomicMap.convert(safe: false, underscore: false)
     |> changeset()
   end
+
+  defp validate_top_level_selector(%{changes: %{sourceFormat: source_format}} = changeset)
+       when source_format in ["xml", "text/xml"] do
+    validate_required(changeset, [:topLevelSelector], message: "is required")
+  end
+
+  defp validate_top_level_selector(%{changes: %{sourceFormat: source_format, topLevelSelector: top_level_selector}} = changeset)
+       when source_format in ["json", "application/json"] do
+    case Jaxon.Path.parse(top_level_selector) do
+      {:error, error_msg} -> add_error(changeset, :topLevelSelector, error_msg.message)
+      _ -> changeset
+    end
+  end
+
+  defp validate_top_level_selector(changeset), do: changeset
 end
--- a/apps/andi/lib/andi_web/live/data_dictionary/data_dictionary_field_editor.ex
+++ b/apps/andi/lib/andi_web/live/data_dictionary/data_dictionary_field_editor.ex
@@ -17,6 +17,7 @@ defmodule AndiWeb.DataDictionary.FieldEditor do
     id = Atom.to_string(assigns.id)
     form_with_errors = DataDictionaryHelpers.add_errors_to_form(assigns.form)
     field_type = input_value(assigns.form, :type)
+    editing_ingestion? = assigns.dataset_or_ingestion == :ingestion
 
     ~L"""
       <div id="<%= @id %>" class="data-dictionary-field-editor" >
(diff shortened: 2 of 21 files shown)


apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -54,4 +62,19 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     |> AtomicMap.convert(safe: false, underscore: false)
     |> changeset()
   end
+
+  defp validate_top_level_selector(%{changes: %{sourceFormat: source_format}} = changeset)
+       when source_format in ["xml", "text/xml"] do
```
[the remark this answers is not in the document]
RyanRConaway: Oh, nevermind. It's part of a function guard. The function wouldn't run

apps/andi/lib/andi_web/live/data_dictionary/data_dictionary_field_editor.ex
```diff
@@ -17,6 +17,7 @@ defmodule AndiWeb.DataDictionary.FieldEditor do
     id = Atom.to_string(assigns.id)
     form_with_errors = DataDictionaryHelpers.add_errors_to_form(assigns.form)
     field_type = input_value(assigns.form, :type)
+    editing_ingestion? = assigns.dataset_or_ingestion == :ingestion
```
RyanRConaway, line 20: Elixir question - Does the ? have special meaning? Is it an optional in case assigns.dataset_or_ingestion is nil?

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -54,4 +62,19 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
...
+    validate_required(changeset, [:topLevelSelector], message: "is required")
```
RyanRConaway, line 68: XML question(I rarely use XML) - Why is a topLevelSelector required? Isn't it similar to JSON and could automatically just parse the first element, allowing overwritting topLevelSelector to start at a nested field?

apps/andi/lib/andi_web/live/ingestion_live_view/metadata_form.ex
```diff
@@ -65,6 +81,13 @@ defmodule AndiWeb.IngestionLiveView.MetadataForm do
     {:noreply, socket}
   end
 
+  def handle_info(
+        %{topic: "ingestion-published"},
+        socket
```
RyanRConaway, line 86: Is this event/topic needed? It looks like `ingestion_published? = ingestion.submissionStatus == :published` does the logic. I'm not seeing where socket.ingestion_published is used

apps/andi/lib/andi_web/live/data_dictionary/data_dictionary_field_editor.ex
bmitchinson: Indicates a boolean: https://hexdocs.pm/elixir/main/naming-conventions.html#trailing-question-mark-foo

apps/andi/test/integration/andi_web/live/ingestion_live_view/metadata_form_test.exs
```diff
@@ -150,23 +150,39 @@ defmodule AndiWeb.IngestionLiveView.MetadataFormTest do
...
+
+      eventually(fn ->
+        ing = Ingestions.get(ingestion.id)
+        assert ing.submissionStatus == :published
+      end)
+
       assert {:ok, view, html} = live(conn, @url_path <> ingestion.id)
 
-      refute Enum.empty?(get_attributes(html, ".ingestion-metadata-form__format select", "disabled"))
+      refute Enum.empty?(get_attributes(html, "#ingestion_metadata_form_sourceFormat", "disabled"))
```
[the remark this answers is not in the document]
bmitchinson: Yeah the way ids are determined as part of pheonix live view aren't always clear. I just double checked and that selector is valid though. Here's a snippet from andi running in product. Not sure how the selector prior to this was working.

```html
<select class="select ingestion-form-fields" id="ingestion_metadata_form_sourceFormat" name="form_data[sourceFormat]" required=""><option value="">...</select>
```

It appends the schema attribute `sourceFormat` to the id attribute of the parent form `https://github.com/UrbanOS-Public/smartcitiesdata/blob/master/apps/andi/lib/andi_web/live/ingestion_live_view/metadata_form.ex#L32`

There doesn't seem to be a way to clearly provide an alternate id to make searching in the codebase for a match more productive: https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#select/4

RyanRConaway (approved):

apps/andi/lib/andi_web/live/ingestion_live_view/metadata_form.ex
bmitchinson: Yes that `ingestion_published? = ingestion.submissionStatus == :published` in the mount method determines the initial `ing_published?` state. The mount method is never called again though, despite the parent state changing.

- https://github.com/phoenixframework/phoenix_live_view/issues/396
- https://github.com/UrbanOS-Public/internal/blob/main/wiki_files/live_view_notes.md

This `handle_info` method responds to a successfully publish and updates that attribute in response.

Open to ideas on how to make that more clear. Imo it's just remembering that the mount only happens once.

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
bmitchinson: A good question that I'm not sure the answer to. All XML examples in our system use a topLevelSelector, and it's been a required field since I've been involved.

As we get feedback on ingesting xml datasets, this could definitely be a potential improvement.

PR #1542, opened 2022-11-21
added eventually to visualization test

nicholas-goodwin:
## Description

- attempts to fix flakey test - elasticsearch

## Reminders:
      in `mix.exs`?

### DIFF
--- a/apps/discovery_api/test/integration/docker-compose.yaml
+++ b/apps/discovery_api/test/integration/docker-compose.yaml
@@ -113,6 +113,7 @@ services:
     - "9300:9300"
     environment:
     - "discovery.type=single-node"
+    - "bootstrap.system_call_filter=false"
     healthcheck:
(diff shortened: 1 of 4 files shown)


apps/discovery_api/test/integration/docker-compose.yaml
```diff
@@ -113,6 +113,7 @@ services:
     - "9300:9300"
     environment:
     - "discovery.type=single-node"
+    - "bootstrap.system_call_filter=false"
```
RyanRConaway, line 116: What does this do?

RyanRConaway (approved):

apps/discovery_api/test/integration/docker-compose.yaml
nicholas-goodwin: While testing, I found that the elasticsearch would just become unhealthy after 5-6 minutes. This is to help those with M1 chips to not let that happen. Was something people have offered as a solution to the specific error presented.

PR #1548, opened 2022-11-30
873 alerting on smartcitiesdata master failures

nicholas-goodwin:
## Description

- ...

## Reminders:
      in `mix.exs`?

### DIFF
--- a/.github/workflows/alchemist.yml
+++ b/.github/workflows/alchemist.yml
@@ -55,22 +55,12 @@ jobs:
         env:
           QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
           QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-      - if: github.ref == 'refs/heads/master'
-        name: Report Status Master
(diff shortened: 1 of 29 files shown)


.github/workflows/alchemist.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         env:
           QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
           QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-      - if: always()
+      - if: always() && github.ref == 'refs/heads/master'
```
c-m-duncan, line 58: So does this not need the always part anymore? Why does an always exist anyway for conditionals like this?

nicholas-goodwin: the always forces it to always run, even if failure, so it does need to be there

c-m-duncan: Oh that's interesting

c-m-duncan (approved):

PR #1550, opened 2022-12-02
505 overwrite api key

ijabbott:
## Description

- Adding a regenerate api key endpoint to enable users to reset their api keys

## Reminders:
      in `mix.exs`?

### DIFF
--- a/apps/raptor/lib/raptor/services/auth0_management_service.ex
+++ b/apps/raptor/lib/raptor/services/auth0_management_service.ex
@@ -26,6 +26,22 @@ defmodule Raptor.Services.Auth0Management do
     end
   end
 
+  def patch_api_key(userID, apiKey) do
+    #TODO: Test
+    audience = Keyword.fetch!(auth0(), :audience)
+    url = "#{audience}users/#{userID}"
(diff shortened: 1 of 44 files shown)


apps/raptor/lib/raptor/services/auth0_management_service.ex
```diff
@@ -26,6 +26,23 @@ defmodule Raptor.Services.Auth0Management do
     end
   end
 
+  def patch_api_key(userID, apiKey) do
+    # TODO: Test
```
nicholas-goodwin, line 30: remove the todos

apps/raptor/lib/raptor/services/auth0_management_service.ex
```diff
@@ -52,10 +69,13 @@ defmodule Raptor.Services.Auth0Management do
 
     case post(url, req_body, headers: [{"content-type", "application/x-www-form-urlencoded"}]) do
       {:ok, response} ->
+        IO.inspect(response, label: "Auth response")
```
nicholas-goodwin, line 72: more IO.inspects

nicholas-goodwin (approved):

PR #1560, opened 2022-12-15
936 require api key

ijabbott:
## Description

When REQUIRE_API_KEY is true, require api_key to be passed in as a header for discovery-api requests if the request does not already contain valid user login information. This should result in api_key being required on requests not coming from discovery-ui.

## Reminders:
      in `mix.exs`?


apps/discovery_api/lib/discovery_api_web/plugs/set_current_user.ex
```diff
@@ -2,13 +2,67 @@ defmodule DiscoveryApiWeb.Plugs.SetCurrentUser do
...
+    end
+  end
+
+  defp assign_current_user(conn, current_user, api_key) when not is_nil(current_user) do
     assign(conn, :current_user, current_user)
   end
+
+  defp assign_current_user(conn, current_user, api_key) when is_nil(current_user) and is_nil(api_key) do
+    render_401_missing_api_key(conn)
+    #    assign(conn, :current_user, current_user)
```
ijabbott, line 28: testing a few things around this 401 missing case, will remove comment after that's done

apps/raptor/lib/raptor_web/controllers/api_key_controller.ex
```diff
@@ -29,9 +29,49 @@ defmodule RaptorWeb.ApiKeyController do
...
+        else
+          {:error,
+           "Only users who have validated their email address and aren't blocked may make API calls"}
+        end
+
+      _ ->
+        Logger.warn("Multiple users cannot have the same API Key.")
+        {:error, "Multiple users cannot have the same API Key."}
+    end
+  end
```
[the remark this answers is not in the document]
ijabbott: It's possible, I just wanted to changed the return values of the method, so figured I'd create a new one. I think the RaptorWeb version returns true/false, so I may be able adapt it to work with {:ok, _} / {:error, _}

apps/raptor/lib/raptor_web/views/api_key_view.ex
```diff
@@ -6,4 +6,10 @@ defmodule RaptorWeb.ApiKeyView do
       apiKey: apiKey
     }
   end
+
+  def render("getUserIdFromApiKey.json", %{user_id: user_id}) do
+    %{
+      user_id: user_id
+    }
+  end
```
bmitchinson, line 14: Non-blocking: Would you mind providing context on what utilizes this `ApiKeyView` module?

It's clearly a pattern based on the module contents outside this addition, but after looking through all of `smartcitiesdata`, I don't see any references to `getUserIdFromApiKey` or `ApiKeyView`.

What does this module enable?

ijabbott: lib/raptor_web/controllers/api_key_controller.ex contains the method getUserIdFromApiKey. When the Auth0 returns a single user connected to an api_key, we return :
```
render(conn, %{
              user_id: user_id
            })
```

That render return requires a pattern match which is specified here in the api_key_view.ex file

apps/discovery_api/lib/discovery_api_web/plugs/set_current_user.ex
```diff
@@ -2,13 +2,67 @@ defmodule DiscoveryApiWeb.Plugs.SetCurrentUser do
...
+
+  defp render_500_internal_server_error(conn) do
+    conn
+    |> DiscoveryApiWeb.RenderError.render_error(500, "Internal Server Error")
+    |> halt()
+  end
+
+  defp get_api_key_from_header(conn) do
+    get_req_header(conn, "api_key") |> List.first()
+  end
```
bmitchinson, line 67: Looks great ! 

Very readable if someone new comes to discovery api to learn about auth

apps/raptor/lib/raptor_web/controllers/api_key_controller.ex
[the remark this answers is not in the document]
bmitchinson: Makes sense. If it proves challenging to reuse the whole block, I mostly just want `if(user["email_verified"] and !user["blocked"])` encapsulated somewhere so that folks don't forget to consider either case. Should have encapsulated it in my last PR that added blocked consideration.

bmitchinson (approved):
Looks great! Thanks for utilizing the raptor service + raptor to keep api as thin as possible.

apps/discovery_api/lib/discovery_api_web/plugs/set_current_user.ex
```diff
@@ -2,13 +2,73 @@ defmodule DiscoveryApiWeb.Plugs.SetCurrentUser do
...
+        current_user = Guardian.Plug.current_resource(new_conn)
+        assign(new_conn, :current_user, current_user)
+
+      error ->
+        render_401_missing_api_key(conn)
+    end
+  end
+
+  defp assign_current_user(conn, current_user, api_key) when is_nil(current_user) and not is_nil(api_key) do
+    case RaptorService.get_user_id_from_api_key(raptor_url(), api_key) do
```
ian-j-abbott-accenture, line 41: @nicholas-goodwin This line here

PR #1561, opened 2022-12-18
WIP: Metadata form refactor

RyanRConaway:
## Description

- TODO

## Reminders:
      in `mix.exs`?


apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -63,7 +63,7 @@ defmodule Andi.InputSchemas.Ingestion do
     source_format = Map.get(changes, :sourceFormat, nil)
 
     ingestion
-    |> cast(changes_with_id, @cast_fields, empty_values: [])
+    |> cast(changes_with_id, @cast_fields, empty_values: [""])
```
RyanRConaway, line 66: Fixes a bug in product that prevents removing a dataset. (Appears to remove, but if you save and return, it restores the previous dataset)

Explanation: Because dataset is a required foreign key, the changeset rejects "" (empty string) as a valid change and falls back to the existing data. However, the changeset allows nil to be a valid change, effectively unsetting the field.

This line adds "" as an empty value, converting it to nil, therefore allowing the metadata form to unset the changeset field.

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -58,32 +58,75 @@ defmodule Andi.InputSchemas.Ingestion do
...
     source_format = Map.get(changes, :sourceFormat, nil)
 
     ingestion
-    |> cast(changes_with_id, @cast_fields, empty_values: [])
-    |> validate_required(@required_fields, message: "is required")
-    |> cast_assoc(:schema, with: &DataDictionary.changeset(&1, &2, source_format), invalid_message: "is required")
-    |> cast_assoc(:extractSteps, with: &ExtractStep.changeset/2)
-    |> cast_assoc(:transformations, with: &Transformation.changeset/2)
-    |> foreign_key_constraint(:targetDataset)
+    |> Changeset.cast(changes_with_id, @cast_fields, empty_values: [""])
```
RyanRConaway, line 66: I found it confusing to know which functions were native to Ecto.Changeset, so I prefer to explicitly use Changeset.<fun>

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -58,32 +58,75 @@ defmodule Andi.InputSchemas.Ingestion do
...
+    |> Changeset.foreign_key_constraint(:targetDataset)
     |> validate_source_format()
     |> CadenceValidator.validate()
     |> validate_top_level_selector()
     |> validate_schema()
     |> validate_extract_steps()
   end
 
-  def changeset_for_draft(ingestion, changes) do
+  def changeset(%Ecto.Changeset{ data: %__MODULE__{} } = changeset, changes) do
```
RyanRConaway, line 79: Allows creating a changeset from an existing changeset without grabbing the intermediate ingestion data (merging changesets)

apps/andi/lib/andi/input_schemas/ingestion.ex
RyanRConaway, line 66: Fixes a bug in product that prevents removing a dataset. (Appears to remove, but if you save and return, it restores the previous dataset)

Explanation: Because dataset is a required foreign key, the changeset rejects "" (empty string) as a valid change and falls back to the existing data. However, the changeset allows nil to be a valid change, effectively unsetting the field.

This line adds "" as an empty value, converting it to nil, therefore allowing the metadata form to unset the changeset field.

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -58,32 +58,75 @@ defmodule Andi.InputSchemas.Ingestion do
...
-    |> cast_assoc(:schema, with: &DataDictionary.changeset_for_draft_ingestion/2)
-    |> cast_assoc(:extractSteps, with: &ExtractStep.changeset_for_draft/2)
-    |> cast_assoc(:transformations, with: &Transformation.changeset_for_draft/2)
+    |> Changeset.cast(changes_with_id, @cast_fields, empty_values: [""])
+    |> Changeset.cast_assoc(:schema, with: &DataDictionary.changeset_for_draft_ingestion/2)
+    |> Changeset.cast_assoc(:extractSteps, with: &ExtractStep.changeset_for_draft/2)
+    |> Changeset.cast_assoc(:transformations, with: &Transformation.changeset_for_draft/2)
+  end
+
+  def changeset_for_draft(%Ecto.Changeset{ data: %__MODULE__{} } = changeset, changes) do
```
RyanRConaway, line 106: Same concept as above, except drafts require less validation. 

Note on the draft concept: Ideally, the UI would always use the full ingestion validation and then reject database saves only when publishing. I believe the current pattern allows for invalidated publishes if handled slightly wrong. I didn't look too deeply into this though.

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -58,32 +58,75 @@ defmodule Andi.InputSchemas.Ingestion do
...
+    changeset
+    |> Changeset.cast(changes, @cast_fields, empty_values: [""])
+    |> Changeset.cast_assoc(:schema, with: &DataDictionary.changeset_for_draft_ingestion/2)
+    |> Changeset.cast_assoc(:extractSteps, with: &ExtractStep.changeset_for_draft/2)
+    |> Changeset.cast_assoc(:transformations, with: &Transformation.changeset_for_draft/2)
+  end
+
+  def merge_metadata_changeset(
```
RyanRConaway, line 114: Allows easy merging from child to parent (metadata_schema -> ingestion_schema)

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -26,14 +26,39 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     :targetDataset
   ]
 
+  def extract_from_ingestion_changeset(%Ecto.Changeset{ data: %Andi.InputSchemas.Ingestion{} } = ingestion_changeset) do
```
RyanRConaway, line 29: Allows for extracting a child changeset from a parent changeset (ingestion -> metadata).

Allows for mapping fields (and errors) that don't line up identically for UI/Database separation of concerns.

This will allow all validation to happen at the API level (top level Andi parent changeset) instead of AndiWeb, then extract fields/errors down from there, mapping each as needed without the overhead of database migrations if you want to change a field name in an AndiWeb schema.

I would go as far to say, anything in AndiWeb.InputSchema should NOT have direct validation, so that all validation is made on API calls as well

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -26,14 +26,39 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
...
+    ingestion_data = ingestion_changeset
```
RyanRConaway, line 30: apply_changes gets the changed data regardless of any validations, aka, the complete latest data

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -26,14 +26,39 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
...
   def changeset(changes), do: changeset(%__MODULE__{}, changes)
 
-  def changeset(%__MODULE__{} = current, changes) do
+  def changeset(current, changes) do
     current
-    |> cast(changes, @cast_fields, empty_values: [])
-    |> validate_required(@required_fields, message: "is required")
-    |> validate_top_level_selector()
-    |> target_dataset_exists()
+    |> Changeset.cast(changes, @cast_fields, empty_values: [])
```
RyanRConaway, line 59: Removed all validations inside the AndiWeb schema. Andi schemas should be responsible for validations.

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -42,39 +67,10 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
     changeset(ingestion)
   end
 
-  defp target_dataset_exists(changeset) do
```
RyanRConaway, line 45: Duplicated code now that validations are passed down

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -42,39 +67,10 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
...
-    end)
-  end
-
   @spec changeset_from_form_data(any) :: Ecto.Changeset.t()
   def changeset_from_form_data(form_data) do
     form_data
     |> AtomicMap.convert(safe: false, underscore: false)
     |> changeset()
   end
-
```
RyanRConaway, line 65: Duplicated code now that validations are passed down

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -8,26 +8,56 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
+        page_error: false,
+        ingestion: ingestion,
+        save_success: false,
+        success_message: "",
+        user_id: user_id
+      )}
+  end
+
   def render(assigns) do
+    metadata_changeset = IngestionMetadataFormSchema.extract_from_ingestion_changeset(assigns.changeset)
```
RyanRConaway, line 44: Should build all child changesets here and pass down to live_component as assigns.

If the child has UI-only fields that the API/Database wouldn't care about, the child can merge it's current changeset to the changeset provided by the parent on render.

RyanRConaway: If performance becomes an issue (Which I doubt unless changesets are heavier than expected), an optimization may be to detect when the changeset is in the _changed property of the assigns (Meaning it just was changed this render) and only rebuild the changeset then.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -8,26 +8,56 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
-        <%= hidden_input(f, :sourceFormat) %>
         <div class="edit-ingestion-title">
           <h1 class="component-title-text">Define Data Ingestion</h1>
         </div>
 
-        <div>
-          <%= live_render(@socket, AndiWeb.IngestionLiveView.MetadataForm, id: :ingestion_metadata_form_editor, session: %{"ingestion" => @ingestion}) %>
-        </div>
+      <div>
+        <%= live_component(@socket, AndiWeb.IngestionLiveView.MetadataForm,
```
RyanRConaway, line 55: Changing child to a live_component so that:
- It shares a process with the parent, eliminating possibility of race conditions
- Reduces complexity of pub/sub tracing
- Enables re-rendering of child when passing state down

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -82,27 +111,38 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
     """
   end
 
-  def mount(
-        _params,
-        %{"is_curator" => is_curator, "ingestion" => ingestion, "user_id" => user_id} = _session,
+  def handle_info(
+        {:updated_metadata,
+          %Ecto.Changeset{ data: %AndiWeb.InputSchemas.IngestionMetadataFormSchema{} } = metadata_changeset },
```
RyanRConaway, line 116: A nice "trick" for specifying the type of changeset passed by uses the data atom to match the type.

Especially nice when paired with the unhandled/catch_all handle_info/handle_event functions below.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -82,27 +111,38 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
         socket
       ) do
-    default_changeset = InputConverter.andi_ingestion_to_full_ui_changeset(ingestion)
+    new_ingestion_changeset = Andi.InputSchemas.Ingestion.merge_metadata_changeset(socket.assigns.changeset, metadata_changeset)
```
RyanRConaway, line 119: This handler catches child messages with the :updated_metadata event. It then merges the child metadata into the parent changeset which:
- Performs any validation at the top/api level
- Keeps state in sync with other children
- Rerenders the direct child **and any other children that share information!** No need to broadcast messages between children.

Note: We unfortunately still have to broadcast about the source-format because data_dictionary_form.ex is not refactored into this parent/child pattern yet.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -132,6 +164,23 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
     {:noreply, assign(socket, page_error: true, save_success: false)}
   end
 
+  def handle_info(
+        %{topic: topic, event: event, payload: payload}, socket) do
```
RyanRConaway, line 168: I found these catch_all handlers to be helpful. It logs the info and then reassigns the current state, allowing the live_view to not crash when unexpected events come through, but still logging them.

Opinions? Are there better ways to log besides IO.inspect? This portion isn't critical to the refactor.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -158,18 +207,27 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
   def handle_event("publish", _, socket) do
     ingestion_id = socket.assigns.ingestion.id
 
+    save_ingestion_safe(socket)
```
RyanRConaway, line 210: Save the changeset on the socket before messaging all the subforms to mass-save to the DB

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -194,34 +256,80 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
     ingestion_id = socket.assigns.ingestion.id
+
     AndiWeb.Endpoint.broadcast_from(self(), "form-save", "save-all", %{ingestion_id: ingestion_id})
     # Todo: Rearchitect how concurrent events are handled and remove these sleeps from draft-save and publish of datasets and ingestions
     # This sleep is needed because other save events are executing. publish_ingestion will load the ingestion from the database.
     Process.sleep(1_000)
 
     andi_ingestion = Ingestions.get(ingestion_id)
     ingestion_changeset = InputConverter.andi_ingestion_to_full_ui_changeset(andi_ingestion)
+    updated_socket = assign(socket, changeset: ingestion_changeset)
```
RyanRConaway, line 310: The current socket is out-of-sync from the database at this point, so you must construct a new changeset from the up-to-date data from the database, assign it to the socket, then use that to determine the correct save message :(

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -194,34 +256,80 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
...
+    {:noreply, update_save_message(updated_socket, "valid")}
   end
 
   defp save_message(true = _valid?), do: "Saved successfully."
 
   defp save_message(false = _valid?),
     do: "Saved successfully. You may need to fix errors before publishing."
 
+
+  defp update_save_message(socket, status) do
```
RyanRConaway, line 322: Just extracted this to reduce duplication

apps/andi/lib/andi_web/live/ingestion_live_view/metadata/metadata_form.ex
```diff
@@ -1,209 +1,112 @@
...
-        <button class="btn btn--select-dataset-search btn--primary-outline" phx-click="select-dataset" type="button">Select Dataset</button>
-        <%= ErrorHelpers.error_tag(f, :targetDataset, bind_to_input: false) %>
-      </div>
-
-    </form>
-    <%= live_component(@socket, AndiWeb.IngestionLiveView.SelectDatasetModal, visibility: @select_dataset_modal_visibility, search_results: @search_results, search_text: @search_text, selected_dataset: @selected_dataset, id: :ingestion_metadata_search) %>
+    selected_dataset = assigns.changeset.changes
+      |> Map.get(:targetDataset, "")
 
+    ~L"""
```
RyanRConaway, line 25: Sadly, git didn't detect diffs here. The main changes for this html:

- phx_target is added to most interactions
- (within live_components) phx_target uses @myself to send messages to itself
- If phx_target is not specified, the event will be sent to the parent process (live view)

Also, since the select_dataset_modal cannot close itself, I'm passing a delegate function so the child can determine when it should close

apps/andi/lib/andi_web/live/ingestion_live_view/metadata/metadata_form.ex
```diff
@@ -1,209 +1,112 @@
...
-      new_format: form_data["sourceFormat"],
-      ingestion_id: socket.assigns.ingestion_id
-    })
-
-    form_data
-    |> IngestionMetadataFormSchema.changeset_from_form_data()
-    |> complete_validation(socket)
-  end
-
   def handle_event("validate", %{"form_data" => form_data}, socket) do
```
RyanRConaway, line 76: Since all the metadata should be validated on the API/top level, we send a message to our parent with our updated IngestionMetadataFormSchema changeset but keep our socket the same because our parent will tell us what to render

apps/andi/lib/andi_web/live/ingestion_live_view/metadata/metadata_form.ex
```diff
@@ -1,209 +1,112 @@
...
-        where: not is_nil(business.id),
-        where: ilike(business.dataTitle, type(^like_search_string, :string)),
-        or_where: ilike(business.orgTitle, type(^like_search_string, :string)),
-        or_where: ^search_value in business.keywords,
-        select: dataset
-      )
-
-    query
-    |> Andi.Repo.all()
+  defp close_modal() do
```
RyanRConaway, line 106: delegate function

apps/andi/lib/andi_web/live/ingestion_live_view/metadata/select_dataset_modal.ex
```diff
@@ -1,13 +1,28 @@
 defmodule AndiWeb.IngestionLiveView.SelectDatasetModal do
   use Phoenix.LiveComponent
 
+  import Ecto.Query, only: [from: 2]
+
+  alias Andi.InputSchemas.Datasets.Dataset
+
+  def mount(socket) do
```
RyanRConaway, line 8: Moved all local state that is not needed for API/top level validation into this component. (From the metadata form)

Note: I didn't have enough time to refactor this to it's fullest extent. Technically, this should have it's own UI changeset that plugs into the form, but since the form was already built out with local state (And therefore, no validation), I decided to leave it for time reasons.

apps/andi/lib/andi/secret_service.ex
```diff
@@ -29,10 +29,6 @@ defmodule Andi.SecretService do
     end
   end
 
-  def retrieve_aws_keys() do
-    retrieve("aws_keys/andi")
-  end
-
```
bmitchinson, line 35: Removing unused method. Only used in tests, moved to test code to make that clear.

apps/andi/lib/andi/secret_service.ex
```diff
@@ -64,5 +60,6 @@ defmodule Andi.SecretService do
     |> Vault.auth(%{role: vault_role(), jwt: token})
   end
 
-  defp set_login_ttl(time, interval), do: NaiveDateTime.utc_now() |> NaiveDateTime.add(time, interval)
```
bmitchinson, line 67: Format

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
bmitchinson: Many wins in this refactor from Ryan but "no need to broadcast messages between children" is big on the list for me

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
bmitchinson: I think this is great, I'm constantly putting these in while developing locally, they should ideally never fire, so calling them out in pod logs seems like a good tactic to reduce the occurrences of them happening.

This will make instances of this much easier to fix, and reduce the crashes that caused the "port2" bugs in the first place.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
bmitchinson: This is ":(", but a reminder to others reviewing that this is an existing issue, not new as part of this refactor 👍 

When other components switch over to livecomponent, and all utilize "safe_ingestion_save", this extra step will go away.

Prior to this PR, save messages were at times inaccurate, this should improve their accuracy by checking with the database.

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -237,16 +345,20 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
           {:ok, ingestion_changeset}
 
         error ->
+          IO.inspect("Error from brook event", label: "Publish to Brook Error")
```
bmitchinson, line 348: I'd like to ask that we keep these IO.inspects in. They're properly labeled, and make up for the current log level discrepancies that are across the umbrella.

apps/andi/lib/andi_web/live/ingestion_live_view/metadata/metadata_form.ex
bmitchinson: Hmm. Diffs alone might be enough to convince me to switch to the "html in a template file" pattern we've seen in some of those examples elsewhere. A later effort though.

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -51,71 +51,137 @@ defmodule Andi.InputSchemas.Ingestion do
     :name
   ]
 
+  def validate(%Ecto.Changeset{data: %__MODULE__{}} = changeset) do
+    data_as_changes =
```
RyanRConaway, line 55: Validate does a few important things:
1 - Get the underlying data + any pending changes into a map as changes

2 - Recasts the existing data on top of itself with force_changes: true
2.1 - force_changes: true is extremely important. We needed to upgrade ecto to get this flag.
2.2 - Apply changes over an empty changeset is invalid. It would indicate that all children are new and try to insert them, since essentially you're telling the database I'm changing from nothing to something.
2.3 - Only changes are validated, not underlying data. So validating draft data is impossible since it has already been saved to the database
2.4 - force_changes: true is a framework-supported workaround that allows us to force draft changes as "changes", which will update everything in place rather than try to recreate it.

3 - Uses children changeset function to cast_assoc, which includes validations. Once we refactor the children, we will need to use the validate function instead, or maybe a changeset_with_validation if that makes more sense

apps/andi/lib/andi/input_schemas/ingestion.ex
```diff
@@ -51,71 +51,137 @@ defmodule Andi.InputSchemas.Ingestion do
...
+    |> Changeset.foreign_key_constraint(:targetDataset)
+    |> validate_source_format()
+    |> CadenceValidator.validate()
+    |> validate_top_level_selector()
+    |> validate_schema()
+    |> validate_extract_steps()
+  end
+
+  def validate_database_safety(%Ecto.Changeset{data: %__MODULE__{}} = changeset) do
+    data_as_changes =
```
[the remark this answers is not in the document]
RyanRConaway: Also, again, uses force_changes: true to make sure any existing data is also validated

apps/andi/lib/andi_web/input_schemas/ingestion_metatadata_form_schema.ex
```diff
@@ -26,55 +26,36 @@ defmodule AndiWeb.InputSchemas.IngestionMetadataFormSchema do
...
-      end
-    end)
-  end
-
-  @spec changeset_from_form_data(any) :: Ecto.Changeset.t()
-  def changeset_from_form_data(form_data) do
-    form_data
-    |> AtomicMap.convert(safe: false, underscore: false)
-    |> changeset()
+  def extract_from_ingestion_changeset(%Ecto.Changeset{data: %Andi.InputSchemas.Ingestion{}} = ingestion_changeset) do
```
RyanRConaway, line 29: Now that the parent changeset (Ingestion) is the source of truth, we need to pass down the data and errors appropriately.

apps/andi/test/integration/andi/harvester_test.exs
```diff
@@ -22,9 +22,6 @@ defmodule Andi.Harvest.HarvesterTest do
 
   describe "data json harvesting" do
     setup do
-      Ecto.Adapters.SQL.Sandbox.checkout(Andi.Repo)
```
RyanRConaway, line 25: This is covered in the Andi.DataCase usage

apps/andi/test/integration/andi_web/live/ingestion_live_view/extract_steps/extract_http_step_form_test.exs
```diff
@@ -32,7 +32,7 @@ defmodule AndiWeb.ExtractHttpStepFormTest do
       {:ok, ingestion} =
         IngestionHelpers.create_with_http_extract_step(%{
           action: "GET",
-          url: "test.com",
+          url: "https://test.com",
```
RyanRConaway, line 35: These were getting through changesets unvalidated somehow, but now they're getting caught.

apps/andi/test/integration/andi_web/live/ingestion_live_view/metadata/metadata_form_test.exs
```diff
@@ -67,21 +67,29 @@ defmodule AndiWeb.IngestionLiveView.MetadataFormTest do
...
+      view
+      |> form("#ingestion_metadata_form", form_data: form_data)
+      |> render_change()
 
       html = render(view)
       assert get_value(html, "#ingestion_metadata_form_name") == new_name
     end
 
-    test "dataset name field defaults to it's existing association", %{
+    test "name field shows an error if blank", %{
```
RyanRConaway, line 78: Misleading diffs here. I've just added tests that cover error cases on an integration level.

apps/andi/test/unit/andi/input_schemas/ingestion_test.exs
```diff
@@ -39,12 +49,13 @@ defmodule Andi.InputSchemas.Ingestion.IngestionTest do
     assert %{name: "Updated Name"} == changeset.changes
   end
 
-  describe "changeset" do
+  describe "validate" do
```
RyanRConaway, line 52: This whole test file was making validations, which in now handled by Ingestion.validate instead

apps/discovery_api/mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule DiscoveryApi.Mixfile do
       {:checkov, "~> 1.0", only: [:test, :integration]},
       {:divo, "~> 1.3", only: [:dev, :test, :integration]},
       {:ex_json_schema, "== 0.7.3", only: [:test, :integration]},
-      {:ecto_sql, "== 3.3.3"},
+      {:ecto, "== 3.9.0", override: true},
```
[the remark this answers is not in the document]
RyanRConaway: Good catch. I'll be sure to do that

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -132,6 +175,25 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
     {:noreply, assign(socket, page_error: true, save_success: false)}
   end
 
+  def handle_info(
+        %{topic: topic, event: event, payload: payload},
+        socket
+      ) do
+    IO.inspect(topic, label: 'Unhandled Info Topic in module #{__MODULE__}')
```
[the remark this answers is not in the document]
RyanRConaway: These were deliberately left in. If an event hits these, its essentially an error, but may not always warrant crashing the application.

As for being concise, I could include topic, event, payload, and socket in a single message?

I could also just not include the socket, as that will definitely be the largest output, but the assigns, process id, etc. would be on the socket, which may help debugging.

If it helps, these shouldn't ever happen naturally and should require a code change if they are ever triggered.

Thoughts?

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
RyanRConaway: Yeah, I agree. I'll add more detail to these.

apps/andi/mix.exs
```diff
@@ -50,8 +50,8 @@ defmodule Andi.MixProject do
       {:divo_kafka, "~> 0.1.5", only: [:dev, :integration]},
       {:divo_postgres, "~> 0.2", only: [:dev, :integration]},
       {:divo_redis, "~> 0.1.4", only: [:dev, :integration]},
-      {:ecto, "== 3.5.4"},
-      {:ecto_sql, "== 3.3.3"},
+      {:ecto, "== 3.9.0", override: true},
```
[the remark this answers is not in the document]
RyanRConaway: ecto_sql requires a matching ecto, so :ecto_sql 3.6.2 requires :ecto 3.6.2

I tried to do what it wanted and bumped ecto_sql, which requires a new postgresx, which requires a new guardian_db, which requires a new postgrex, which requires a new elixir.....

Its pretty critical to have the new :ecto version. It has the force_changes: true flag for changesets, which is a game changer in design.

I feel the least egregious version discrepancy is a forced mismatch begin ecto and ecto_sql, rather than a forced mismatch of ecto_sql and postgres/guardian/elxir

apps/andi/priv/repo/migrations/20220614171000_add_codelab_tables.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Andi.Repo.Migrations.CreateCodelabTables do
+  use Ecto.Migration
+
+  # This migration is just for tests. It facilitates changeset testing in the codelabs.
```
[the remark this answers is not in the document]
RyanRConaway: Yeah, I considered that. I actually spent half a day trying to make a Repo just for the codelab and it got rather messy. I eventually concluded that having an empty Person table and Address table in production wouldn't do any real harm. If there is a better way to keep it isolated to a test environment only, I'd be open to it.

jfolk2015 (approved):

PR #1567, opened 2023-01-03
added focus-lock to modals

nicholas-goodwin:
## Description

- Forced modals to not allow tabbing (or shift+tab) out of the modal.

## Reminders:
      in `mix.exs`?


apps/andi/assets/package-lock.json
```diff
@@ -1,5478 +1,15 @@
 {
-  "name": "assets",
-  "lockfileVersion": 2,
   "requires": true,
-  "packages": {
```
c-m-duncan, line 5: Did you do this on purpose? Like a prune that had to occur?

c-m-duncan (approved):
Looks like a good solution to me!

bmitchinson:
I'm open to adding a javascript library to address this if needed. Seems like the bundle size is pretty small. https://bundlephobia.com/package/alpinejs@3.10.5

Were there no other html native ways of doing this? Using tabindex seems interesting but maybe it's not an option, and x-trap is a better way to go. https://www.w3docs.com/snippets/html/how-to-prevent-tab-indexing-on-specific-elements.html

apps/andi/assets/package-lock.json
nicholas-goodwin: no, just what happened when I built. I typically try to leave it out, but I think I had to, as I added a couple imports

nicholas-goodwin:
tabindexing was a no-go, as I would have to go into each field in the background and make it -1 every time we open a modal, which I think will turn ugly real quick.
There are bootstrap ways to build a modal that I know of via html (well, still kinda using something besides html), but don't didn't want to bring in that whole thing just for all the stylings to go to wack.
I can give x-trap a shot, hold a minute.

PR #1578, opened 2023-01-10
983 accessible naming

ian-j-abbott-accenture:
## Description

Added labels and titles where necessary

## Reminders:

      in `mix.exs`?


bmitchinson (approved):
In addition to the several improvements here, looks like all IDs are unique, so LGTM

PR #1579, opened 2023-01-11
unique IDs and unique labels everywhere

ian-j-abbott-accenture:
## Description

Add unique IDs and labels to allow screen readers to differentiate between components and labels with the same text content

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi_web/live/dataset_live_view/metadata_form.ex
+++ b/apps/andi/lib/andi_web/live/dataset_live_view/metadata_form.ex
@@ -62,8 +62,11 @@ defmodule AndiWeb.EditLiveView.MetadataForm do
       <div class="form-section">
         <%= f = form_for @changeset, "#", [phx_change: :validate, phx_submit: :save, as: :form_data] %>
           <%= hidden_input(f, :orgName) %>
+          <%= hidden_input(f, :orgTitle) %>
           <%= hidden_input(f, :orgId) %>
+          <%= hidden_input(f, :dataName) %>
           <%= hidden_input(f, :systemName) %>
+          <%= hidden_input(f, :sourceType) %>
           <%= hidden_input(f, :datasetId) %>
 
           <div class="component-edit-section--<%= @visibility %>">
@@ -75,8 +78,8 @@ defmodule AndiWeb.EditLiveView.MetadataForm do
               </div>
 
               <div class="metadata-form__data-name">
-                <%= label(f, :dataName, DisplayNames.get(:dataName), class: "label label--required") %>
-                <%= text_input(f, :dataName, [class: "input input--text", readonly: true, required: true]) %>
+                <%= label(f, :dataName, DisplayNames.get(:dataName), class: "label label--required", for: "metadata_#{@socket.id}__data-name") %>
+                <%= text_input(f, :dataName, [id: "metadata_#{@socket.id}__data-name", class: "input input--text", readonly: true, required: true]) %>
                 <%= ErrorHelpers.error_tag(f, :dataName, bind_to_input: false) %>
               </div>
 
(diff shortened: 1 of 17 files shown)


apps/andi/lib/andi_web/live/dataset_live_view/metadata_form.ex
```diff
@@ -62,11 +62,8 @@ defmodule AndiWeb.EditLiveView.MetadataForm do
       <div class="form-section">
         <%= f = form_for @changeset, "#", [phx_change: :validate, phx_submit: :save, as: :form_data] %>
           <%= hidden_input(f, :orgName) %>
-          <%= hidden_input(f, :orgTitle) %>
```
ian-j-abbott-accenture, line 65: I don't believe the deleted hidden inputs are needed because visible inputs exist for each of these items. I'm assuming these are left over from an initial iteration of the form. I could absolutely be wrong here, so please correct me if I am.

apps/andi/lib/andi_web/live/ingestion_live_view/extract_steps/extract_auth_step_form.ex
```diff
@@ -36,36 +36,36 @@ defmodule AndiWeb.ExtractSteps.ExtractAuthStepForm do
 
             <div class="extract-auth-step-form__destination">
               <%= label(f, :destination, DisplayNames.get(:destination), class: "label label--required", for: "step_#{@id}__auth_destination") %>
-              <%= text_input(f, :destination, [id: "step_#{@id}__auth_destination", class: "input", required: true]) %>
+              <%= text_input(f, :destination, [id: "step_#{@id}__auth_destination", aria_label: "step_#{@id}__auth_destination", class: "input", required: true]) %>
```
ian-j-abbott-accenture, line 39: I'm not sold on these aria_labels. They make the input have a unique label that's identifiable by a screen reader, but I'm worried the id is not going to make sense since it will probably look something like this: "step_44404bee-00ab-4f8b-a72b-3ae3b551ec95__auth_destination"

I'd be happy to try any alternatives.

apps/andi/lib/andi_web/live/ingestion_live_view/extract_steps/extract_secret_step_form.ex
```diff
@@ -37,7 +37,7 @@ defmodule AndiWeb.ExtractSteps.ExtractSecretStepForm do
 
               <div class="extract-secret-step-form__destination">
                 <%= label(f, :destination, DisplayNames.get(:destination), class: "label label--required", for: "step_#{@id}__secret_destination") %>
-                <%= text_input(f, :destination, [id: "step_#{@id}__secret_destination", class: "extract-secret-step-form__destination input", phx_target: "#step-#{@id}", required: true]) %>
+                <%= text_input(f, :destination, [id: "step_#{@id}__secret_destination", aria_label: "step_#{@id}__secret_destination", class: "extract-secret-step-form__destination input", phx_target: "#step-#{@id}", required: true]) %>
                 <%= ErrorHelpers.error_tag(f, :destination) %>
               </div>
```
ian-j-abbott-accenture, line 43: Just below here on line 47 we have an issue. I can't add an aria_label because it's an input type="password". This makes AI inspector complain about input type["password"] not being able to have an aria_label for some reason. Not sure how to solve this.

apps/andi/lib/andi_web/live/ingestion_live_view/extract_steps/extract_auth_step_form.ex
c-m-duncan: What is that ID? The process id? The component id?

Maybe the parent component could pass in a number indicating which order it was generated in / placed in, and that could be used instead?

ian-j-abbott-accenture: I thought about that, just need to figure out how to get an index from the elixir for loop render

apps/andi/lib/andi_web/live/dataset_live_view/metadata_form.ex
ian-j-abbott-accenture: undid this and explored a different options for safety

apps/andi/lib/andi_web/live/ingestion_live_view/extract_steps/extract_auth_step_form.ex
ian-j-abbott-accenture: should be good here now

jfolk2015 (approved):

PR #1586, opened 2023-01-19
Fix toast and publish

RyanRConaway:
## Description

- Added a publish specific toast
- Fixed Brook message encoding/decoding for extract steps body

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi/event/event_handler.ex
+++ b/apps/andi/lib/andi/event/event_handler.ex
@@ -50,9 +50,10 @@ defmodule Andi.Event.EventHandler do
     ingestion_update()
     |> add_event_count(author, data.id)
 
-    Ingestions.update_ingested_time(data.id, DateTime.utc_now())
+    data
+    |> Map.put(:ingestionTime, %{ingestionTime: DateTime.to_iso8601(DateTime.utc_now())})
+    |> Ingestions.update()
 
-    Ingestions.update(data)
     IngestionStore.update(data)
   end
 
--- a/apps/andi/lib/andi/input_schemas/input_converter.ex
+++ b/apps/andi/lib/andi/input_schemas/input_converter.ex
@@ -3,6 +3,8 @@ defmodule Andi.InputSchemas.InputConverter do
   Used to convert between SmartCity.Datasets, form data (defined by Andi.InputSchemas.DatasetInput), and Ecto.Changesets.
   """
 
+  require Logger
+
   alias Andi.InputSchemas.Datasets.Dataset
   alias Andi.InputSchemas.Ingestion
   alias Andi.InputSchemas.Organization
@@ -266,7 +268,17 @@ defmodule Andi.InputSchemas.InputConverter do
   defp update_context_from_smrt_step(context, _), do: context
 
   defp encode_extract_step_body_as_json(%{body: body} = smrt_extract_step) when body not in ["", nil] do
-    Map.put(smrt_extract_step, :body, Jason.encode!(body))
+    case body do
+      _ when is_binary(body) ->
+        Map.put(smrt_extract_step, :body, Jason.decode!(body))
+
+      _ when is_map(body) ->
+        Map.put(smrt_extract_step, :body, body)
+
+      _ ->
+        Logger.error("Received an extract step body that is not a string or a map. Received body: #{body}}")
+        smrt_extract_step
+    end
   end
 
   defp encode_extract_step_body_as_json(smrt_extract_step), do: smrt_extract_step
--- a/apps/andi/lib/andi/input_schemas/input_converter.ex
+++ b/apps/andi/lib/andi/input_schemas/input_converter.ex
@@ -270,15 +270,20 @@ defmodule Andi.InputSchemas.InputConverter do
   defp encode_extract_step_body_as_json(%{body: body} = smrt_extract_step) when body not in ["", nil] do
     case body do
       _ when is_binary(body) ->
-        Map.put(smrt_extract_step, :body, Jason.decode!(body))
+        raise_error_if_invalid_json(body)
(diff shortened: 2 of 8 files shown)


nicholas-goodwin (approved):

apps/andi/lib/andi/event/event_handler.ex
```diff
@@ -50,9 +50,10 @@ defmodule Andi.Event.EventHandler do
     ingestion_update()
     |> add_event_count(author, data.id)
 
-    Ingestions.update_ingested_time(data.id, DateTime.utc_now())
+    data
```
RyanRConaway, line 53: Saving twice to the database is unnecessary and error prone. The 2nd write would potentially overwrite the first if underlying implementation changed.

Just including the ingestion time as part of the single update...

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -266,7 +268,22 @@ defmodule Andi.InputSchemas.InputConverter do
   defp update_context_from_smrt_step(context, _), do: context
 
   defp encode_extract_step_body_as_json(%{body: body} = smrt_extract_step) when body not in ["", nil] do
-    Map.put(smrt_extract_step, :body, Jason.encode!(body))
+    case body do
```
RyanRConaway, line 271: Originally, I thought this was the error, but turns out it was overwriting incorrectly later in the PR. I left this here for safety though in case of a malformed message body.

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -337,6 +353,7 @@ defmodule Andi.InputSchemas.InputConverter do
 
   defp update_context_from_andi_step(context, "http") do
     context
+    |> cast_keys_to_atom_in_map_recursively()
```
RyanRConaway, line 356: This was the primary cause of the malformed body bug. The message would come in with string keys, but then all the extract_steps fields expect atom keys. This would end up creating two sets of fields, both atoms and keys. Keys being the data from the message, and atoms being re-initialized empty values that get populated later (Or maybe never at all... It's hard to tell with extract_steps)

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -345,6 +362,42 @@ defmodule Andi.InputSchemas.InputConverter do
     |> Map.update(:url, nil, &remove_query_params_from_url/1)
   end
 
+  defp cast_keys_to_atom_in_map_recursively(map) do
```
RyanRConaway, line 365: If the key is a string, cast to atom.
-then-
check if the value is a map or list and repeat if needed

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -266,7 +268,22 @@ defmodule Andi.InputSchemas.InputConverter do
...
+      _ when is_binary(body) ->
+        raise_error_if_invalid_json(body)
+        Map.put(smrt_extract_step, :body, body)
+
+      _ when is_map(body) ->
+        Map.put(smrt_extract_step, :body, Jason.encode!(body))
+
+      _ ->
+        Logger.error("Received an extract step body that is not a string or a map. Received body: #{body}}")
+        smrt_extract_step
```
bmitchinson, line 281: Looks great. Plus if this very hard to reach error happens, reaper will still attempt to make the request without the body, which should provide clear logs in the event the server rejects due to a missing body. Instead of reaper just never attempting.

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -345,6 +362,42 @@ defmodule Andi.InputSchemas.InputConverter do
...
+      {key, value} when is_list(value) ->
+        {key, cast_keys_to_atoms_in_list_recursively(value)}
+
+      {key, value} ->
+        {key, value}
+    end)
+    |> Enum.into(%{})
+  end
+
+  defp cast_keys_to_atoms_in_list_recursively(list) do
```
RyanRConaway, line 387: Find nested maps, even inside nested lists, and convert all keys to atoms

apps/andi/test/integration/andi_web/live/ingestion_live_view/edit_ingestion_live_view_test.exs
```diff
@@ -159,6 +159,73 @@ defmodule AndiWeb.EditIngestionLiveViewTest do
       assert get_text(html, "#snackbar") == "Saved successfully. You may need to fix errors before publishing."
     end
 
+    test "Success message is displayed when form data is saved", %{curator_conn: conn, ingestion: ingestion} do
```
bmitchinson, line 162: I <3 tests

apps/andi/lib/andi_web/live/ingestion_live_view/edit_ingestion_live_view.ex
```diff
@@ -233,10 +233,11 @@ defmodule AndiWeb.IngestionLiveView.EditIngestionLiveView do
 
     case publish_ingestion(ingestion_id, socket.assigns.user_id) do
       {:ok, ingestion_changeset} ->
-        {:noreply, assign(socket, changeset: ingestion_changeset)}
+        updated_socket = assign(socket, changeset: ingestion_changeset)
+        {:noreply, update_publish_message(updated_socket, "valid")}
 
       _ ->
-        {:noreply, update_save_message(socket, "invalid")}
+        {:noreply, update_publish_message(socket, "invalid")}
```
RyanRConaway, line 240: Made a message for both success and failure of publish

bmitchinson (approved):

PR #1587, opened 2023-01-19
Disallow creating new ingestions with a specific id

c-m-duncan:
## Description

Trying to fix repeat ingestion / dataset ids by locking down the PUT endpoint

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
```diff
@@ -144,11 +144,27 @@ defmodule AndiWeb.API.IngestionController do
   end
 
   defp add_uuid(message) do
-    uuid = UUID.uuid4()
-
-    Map.merge(message, %{"id" => uuid}, fn _k, v1, _v2 -> v1 end)
+    if Map.has_key?(message, "id") do
+      case IngestionStore.get(Map.get(message, "id")) do
```
c-m-duncan, line 148: Is this ingestion store the right way to do this? Super worried that a mistake will just stop updates from going through to ingestions

RyanRConaway: It should be Ingestions.get(id)

IngestionStore is essentially the cache across all services. It represents what the system knows about the ingestion as a whole. Generally the result of messages being processed.

Ingestions is a wrapper around the postgres database backing Andi. 

So Andi likely will always use Ingestions, then only after it's done all of it's work, finally send a ingestion_update message, which is processed and the IngestionStore is updated.

Simple, right?

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
```diff
@@ -144,9 +144,23 @@ defmodule AndiWeb.API.IngestionController do
...
+        {:ok, nil} ->
+          {:invalid, "Do not include id in create call"}
```
RyanRConaway, line 150: Nice

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
```diff
@@ -144,9 +144,23 @@ defmodule AndiWeb.API.IngestionController do
...
+
+        {:ok, _} ->
+          uuid = UUID.uuid4()
+          {:ok, Map.merge(%{"id" => uuid}, message)}
```
RyanRConaway, line 154: If you find an existing ingestion with the provided ID, you wouldn't want to generate a new ID. This would effectively delete the old ingestion and create a new one. I believe you would want to just return the message since this has been determined as an update message with no need to add_uuid

apps/andi/test/integration/andi_web/controllers/ingestion_controller_test.exs
```diff
@@ -132,6 +131,7 @@ defmodule Andi.IngestionControllerTest do
         |> SmartCity.Helpers.to_atom_keys()
         |> TDG.create_ingestion()
         |> struct_to_map_with_string_keys()
+        |> Map.pop("id")
```
RyanRConaway, line 134: I think you can pass in %{id: nil} to TDG.create_ingestion to overwrite the id during the original creation instead of popping/deleting it at the end

apps/andi/test/unit/andi_web/controllers/api/ingestion_controller_test.exs
```diff
@@ -162,13 +172,57 @@ defmodule AndiWeb.API.IngestionControllerTest do
 
   describe "PUT /ingestion" do
     test "PUT /api/ with data returns a 201", %{conn: conn} do
+      smrt_ingestion = TDG.create_ingestion(%{})
```
RyanRConaway, line 175: Same thing with overwriting the ID

apps/andi/test/unit/andi_web/controllers/api/ingestion_controller_test.exs
```diff
@@ -162,13 +172,57 @@ defmodule AndiWeb.API.IngestionControllerTest do
...
+      {_, ingestion_without_id} = smrt_ingestion |> struct_to_map_with_string_keys() |> Map.pop("id")
+
+      allow(Brook.Event.send(@instance_name, any(), any(), any()), return: :ok)
+      allow(Andi.InputSchemas.Datasets.get(any()), return: %{technical: %{sourceType: "ingest"}})
+
+      conn = put(conn, @route, ingestion_without_id)
+
+      {_, decoded_body} = Jason.decode(response(conn, 201))
+
+      expected_ingestion = TDG.create_ingestion(Map.put(ingestion_without_id, "id", decoded_body["id"]))
```
RyanRConaway, line 185: Do you need the TDG.create_ingestion? The Map.put should give you the full ingestion already.

apps/andi/test/unit/andi_web/controllers/api/ingestion_controller_test.exs
```diff
@@ -162,13 +172,57 @@ defmodule AndiWeb.API.IngestionControllerTest do
...
+    end
+
+    test "PUT /api/ creating a ingestion with a set id returns a 400", %{conn: conn} do
       smrt_ingestion = TDG.create_ingestion(%{})
       ingestion = smrt_ingestion |> struct_to_map_with_string_keys()
+
       allow(Brook.Event.send(@instance_name, any(), any(), any()), return: :ok)
       allow(Andi.InputSchemas.Datasets.get(any()), return: %{technical: %{sourceType: "ingest"}})
-      put(conn, @route, ingestion)
+      allow(IngestionStore.get(any()), return: {:ok, nil})
```
RyanRConaway, line 201: I like the use of expect here to confirm it was called. Preference though.

Regardless, locking down the ID as the arg is probably a good idea

expect(Ingestions.get(smart_ingestion.id), return: {:ok, nil})

apps/andi/test/unit/andi_web/controllers/api/ingestion_controller_test.exs
```diff
@@ -162,13 +172,57 @@ defmodule AndiWeb.API.IngestionControllerTest do
...
+      assert "Do not include id in create call" =~ Map.get(body, "errors")
+    end
+
+    test "PUT /api/ updating an ingestion returns a 201 when an ingestion with that ID is found in the store", %{conn: conn} do
+      smrt_ingestion = TDG.create_ingestion(%{})
+      ingestion = smrt_ingestion |> struct_to_map_with_string_keys()
+
+      allow(Brook.Event.send(@instance_name, any(), any(), any()), return: :ok)
+      allow(Andi.InputSchemas.Datasets.get(any()), return: %{technical: %{sourceType: "ingest"}})
+      allow(IngestionStore.get(any()), return: {:ok, ingestion})
```
RyanRConaway, line 214: Same as above

apps/andi/test/unit/andi_web/controllers/api/ingestion_controller_test.exs
c-m-duncan: Yeah, but that's a map. It's expecting an Ingestion ... object? Not sure what the elixir term is for it.

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
c-m-duncan: So I copied this over from the GET endpoint, figuring consistency was the way to go. Do you think we should change that endpoint to also be hitting postgres?

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
c-m-duncan: oop, yeah good catch 🤦

apps/andi/test/integration/andi_web/controllers/ingestion_controller_test.exs
c-m-duncan: Doesn't that leave %{id: nil} on the ingestion? It seems like it would be closer to the actual use case of not having the field on a call to not have that left over.

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
bmitchinson: Hm, I see why'd you reference that yeah.

Since IngestionStore is returning from the ViewState, the response will be a SmartCity struct cast to JSON. I believe if we instead used "Ingestions.get", we'd get back an AndiSchema.Ingestion response.

Because DatasetController and OrganizationController both use the ViewState Stores (Ex: IngestionsStore like you have here), my vote would be to stick with using IngestionStore, as you have it now, for consistency.

If later we want to have a conversation about switching to returning to Andi Schema responses across the API, I'd be open to that depneding on folks use case other than our internal testing, publish / draft status etc, but that convo could be down the road IMO.

apps/andi/lib/andi_web/controllers/api/ingestion_controller.ex
bmitchinson: Ohh okay. Took me a minute to parse this function. "ifIngestionAlreadyExists" or "ifNewIngestion" would have gone a long way for me to understand I think. Is this check so that the ID can be determined differently, depnding on if this is an Update operation, or a create operation? Someway to indicate that more literally in the code might be nice, but I think the logic here is sound. Does a nice job compromising for the fact that atm, creating / updating happen on the same REST endpoint.

c-m-duncan: Yup yup. And I wasn't sure the most elixir-idiomatic way of doing this. Just kind of threw some matchers at it, but I could see something like 

`if caseExists? do...`

and 
```
def caseExists? do
    case IngestionStore.get(Map.get(message, "id")) do
        {:ok, nil} -> false
        ...
```

apps/andi/test/integration/andi_web/controllers/ingestion_controller_test.exs
bmitchinson: I think either would result in your new controller logic matching correctly, but yeah pop would be a true representation of how we'd like the response to come in.

nil leaves it on the struct.

```elixir
iex(1)> SmartCity.TestDataGenerator.create_ingestion(%{id: nil})
%SmartCity.Ingestion{
  allow_duplicates: true,
  ...
  id: nil,
  ...
}%
```

RyanRConaway (approved):

PR #1592, opened 2023-01-23
Check if dataset exists during ingestion validation in api

c-m-duncan:
## Description

Ingestions with invalid datasets are slipping in through the api.

This PR checks the DatasetStore to see if a dataset with the targetDataset ID exists before allowing an ingestion to save.

image

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/test/unit/andi/input_schemas/ingestion_test.exs
+++ b/apps/andi/test/unit/andi/input_schemas/ingestion_test.exs
@@ -5,6 +5,7 @@ defmodule Andi.InputSchemas.Ingestion.IngestionTest do
 
   alias Andi.InputSchemas.Ingestion
   alias Andi.InputSchemas.Datasets
+  alias Andi.Services.DatasetStore
 
   @ingestion_id Ecto.UUID.generate()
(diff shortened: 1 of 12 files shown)


ian-j-abbott-accenture (approved):

bmitchinson (approved):
✅

Good test

apps/andi/test/unit/andi/input_schemas/ingestion_test.exs
```diff
@@ -18,7 +18,7 @@ defmodule Andi.InputSchemas.Ingestion.IngestionTest do
     schema: [
       %{id: Ecto.UUID.generate(), name: "name", type: "type", bread_crumb: "name", dataset_id: "id", selector: "/cam/cam"}
     ],
-    sourceFormat: "sourceFormat"
+    sourceFormat: "text/csv"
```
c-m-duncan, line 21: Leaving this change and the one below because they are both improvements to existing tests, even though the file under test isn't changed anymore

RyanRConaway (approved):

PR #1594, opened 2023-01-24
added ADA Compliance formatting to ANDI

nicholas-goodwin:
## Description

- Added header connection to tables when no entry
- add alt text for images

## Reminders:

~~- [ ] If changing elixir code in an "app", did you update the relevant version in `mix.exs`?~~

### DIFF
--- a/apps/transformers/lib/transformations/date_time.ex
+++ b/apps/transformers/lib/transformations/date_time.ex
@@ -69,25 +69,25 @@ defmodule Transformers.DateTime do
   def fields() do
     [
       %{
-        field_name: "sourceField",
+        field_name: @source_field,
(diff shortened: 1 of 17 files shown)


apps/transformers/lib/transformations/date_time.ex
```diff
@@ -75,21 +75,21 @@ defmodule Transformers.DateTime do
         options: nil
       },
       %{
-        field_name: "targetField",
+        field_name: "sourceFormat",
```
ian-j-abbott-accenture, line 78: Small thing here, there are constants defined on lines 9-12 in this file that you could use here.

ian-j-abbott-accenture (approved):

PR #1603, opened 2023-02-08
adjust css so accordion mouse-over area is consistent

ian-j-abbott-accenture:
## Description

Adjust css so accordion mouse over behavior is consistent

## Reminders:

      in `mix.exs`?
- ~[ ] If altering an API endpoint, was the relevant postman collection updated?~
  - ~[ ] If a new version of `smart_city` is being used (new fields on a struct), were the relevant postman collections updated?~

### DIFF
--- a/apps/andi/lib/andi_web/live/ingestion_live_view/finalize/finalize_form.ex
+++ b/apps/andi/lib/andi_web/live/ingestion_live_view/finalize/finalize_form.ex
@@ -60,7 +60,7 @@ defmodule AndiWeb.IngestionLiveView.FinalizeForm do
       end
 
     ~L"""
-    <div id="finalize_form" class="finalize-form finalize-form--<%= @visibility %>">
+    <div id="finalize_form" class="finalize-form finalize-form--<%= @visibility %> form-end">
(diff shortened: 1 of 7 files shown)


apps/andi/lib/andi_web/live/ingestion_live_view/finalize/finalize_form.ex
```diff
@@ -60,7 +60,7 @@ defmodule AndiWeb.IngestionLiveView.FinalizeForm do
       end
 
     ~L"""
-    <div id="finalize_form" class="finalize-form finalize-form--<%= @visibility %>">
```
nicholas-goodwin, line 63: does the other steps in ingestion need the same touch or do they work the same?

bmitchinson (approved):

apps/andi/lib/andi_web/live/ingestion_live_view/finalize/finalize_form.ex
ian-j-abbott-accenture: We just need it for the first and last steps. This is so the line that extends between the step numbers doesn't overflow

PR #1613, opened 2023-02-28
1068 add arithmetic transformations

ian-j-abbott-accenture:
## Description

- Add arithmetic transformations to transformations options dropdown
- Add string parsing ability to arithmetic transformations so they actually work

## Reminders:

      in `mix.exs`?


apps/transformers/lib/parse_utils.ex
```diff
@@ -0,0 +1,67 @@
...
+      {:error, reason} -> {:error, reason}
+      _ -> {:ok, Enum.reverse(result)}
+    end
+  end
+
+  def parseValue(value, _) when is_number(value) do
+    {:ok, value}
+  end
+
+  def parseValue(value, payload) do
```
ian-j-abbott-accenture, line 34: I don't love how this function operates at the moment. Ideally we wouldn't need to call all three parse functions, but would instead just return the one that is needed. This doesn't seem possible without enforcing field types on the transformation input parameters so we know which function to call.

It would be possible to just return the first function that succeeds, but I couldn't find a way of doing that without a triple nested conditional, so I went for clarity over performance here.

I'd welcome any alternative ideas people have.

nicholas-goodwin (approved):

apps/transformers/lib/parse_utils.ex
jfolk2015: What about using [Decimal](https://hexdocs.pm/decimal/Decimal.html) which seems to be a BigDecimal equivalent from other languages? Looks like we already have that wired in for the division transformation.

apps/transformers/lib/operation_builder.ex
```diff
@@ -23,12 +23,12 @@ defmodule Transformers.OperationBuilder do
     fn payload -> Transformers.Remove.transform(payload, parameters) end
   end
 
-  def build("arithmetic_add", parameters) do
```
jfolk2015, line 26: I strongly support the rename for these two transformations.

jfolk2015 (approved):

PR #1627, opened 2023-03-15
1071 download blank csv

accenture-mikeyc:
## Description

- Added reportsController and wired up to reports tab download button to download static csv file

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/controllers/reports_controller.ex
```diff
@@ -0,0 +1,14 @@
+defmodule AndiWeb.ReportsController do
+  use AndiWeb, :controller
+
+  access_levels(download_report: [:private])
+
+  def download_report(conn, _params) do
```
ian-j-abbott-accenture, line 6: super minor, but we can probably just remove _params if we're not using it unless it's going to be used in a future iteration and is just a placeholder

ian-j-abbott-accenture (approved):

apps/andi/lib/andi_web/controllers/reports_controller.ex
accenture-mikeyc: because controllers are called through phoenix there is actually a blank params value sent to the function. So removing it will cause an error saying there isn't a function match with arity 2.

PR #1629, opened 2023-03-16
added XML check to HTTP and Auth Step

nicholas-goodwin:
## Description

- Added XML check to HTTP and Auth steps so that the body can be XML without failing

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi/input_schemas/ingestions/extract_steps/extract_auth_step.ex
```diff
@@ -86,8 +86,17 @@ defmodule Andi.InputSchemas.Ingestions.ExtractAuthStep do
...
+      {:ok, _} ->
+        changeset
+
+      {:error, _} ->
+        try do
+          SweetXml.parse(body)
+          changeset
+        catch
+          :exit, _ ->
+            Changeset.add_error(changeset, :body, "could not parse json", validation: :format)
```
ian-j-abbott-accenture, line 98: We should probably add a blurb about not being able to parse xml as well

apps/andi/lib/andi/input_schemas/ingestions/extract_steps/extract_http_step.ex
```diff
@@ -95,8 +96,17 @@ defmodule Andi.InputSchemas.Ingestions.ExtractHttpStep do
...
+      {:ok, _} ->
+        changeset
+
+      {:error, _} ->
+        try do
+          SweetXml.parse(body)
+          changeset
+        catch
+          :exit, _ ->
+            Changeset.add_error(changeset, :body, "could not parse json", validation: :format)
```
ian-j-abbott-accenture, line 108: same as other comment

ian-j-abbott-accenture (approved):

apps/andi/lib/andi/input_schemas/ingestions/extract_steps/extract_auth_step.ex
nicholas-goodwin: This isn't what the UI shows, due to the error_tag function, but still good catch

PR #1636, opened 2023-03-23
Allow .tsv files to be uploaded as data dictionaries

c-m-duncan:
## Description

Previously, only .txt files could be uploaded as TSV data dictionaries. This allows .tsv file extensions as well.

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/live/dataset_live_view/data_dictionary_form.ex
```diff
@@ -156,7 +156,10 @@ defmodule AndiWeb.EditLiveView.DataDictionaryForm do
   end
 
   def handle_event("file_upload", %{"fileType" => file_type}, socket)
-      when file_type not in ["text/csv", "application/json", "application/vnd.ms-excel", "text/plain"] do
+      when file_type not in ["text/csv", "application/json", "application/vnd.ms-excel", "text/plain", "text/tab-separated-values"] do
+    IO.inspect(file_type)
```
ian-j-abbott-accenture, line 160: do you want to inspect this?

PR #1642, opened 2023-04-03
1119 filter metadata from api

RyanRConaway:
## Description

- ...

## Reminders:

      in `mix.exs`?


nicholas-goodwin (approved):

apps/raptor_service/lib/raptor_service.ex
```diff
@@ -102,7 +102,7 @@ defmodule RaptorService do
   end
 
   def check_auth0_role(raptor_url, api_key, role) do
-    case HTTPoison.get(url_for_checking_role(raptor_url, api_key, role) |> IO.inspect(label: "RYAN - URL")) do
```
nicholas-goodwin, line 105: real quick IO inspect

RyanRConaway: Uhhh, that's not mine? I've never seen that IO inspect in my life!

PR #1647, opened 2023-04-05
1116 - datetime string conversion

accenture-mikeyc:
## Description

- Additional conditional display logic for showing extra datetime format field
- new custom validation message indicating only datetime <> string conversions are legal
- update conversion transformation to include datetime conversion

## Reminders:

      in `mix.exs`?


nicholas-goodwin (approved):

accenture-mikeyc:
After testing in the product environment, it was discovered that the functionality added by this PR already implicitly exists within UrbanOS.

PR #1648, opened 2023-04-06
1018 data dictionary refactor

nicholas-goodwin:
## Description

- ...

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_add_field_editor.ex
```diff
@@ -21,38 +21,38 @@ defmodule AndiWeb.DataDictionary.AddFieldEditor do
       end
 
     ~L"""
-    <div id=<%= @id %> class="data-dictionary-add-field-editor data-dictionary-add-field-editor--<%= modifier %>">
+    <div id="<%= @id %>" class="data-dictionary-add-field-editor data-dictionary-add-field-editor--<%= modifier %>">
```
ian-j-abbott-accenture, line 24: super minor, but is there a reason we use @ id instead of the id variable on line 14?

ian-j-abbott-accenture (approved):
Approving, but someone else should probably take a look as well

apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_add_field_editor.ex
nicholas-goodwin: this is because we are passing in an atom as id, a little of the "old way" that they called an event to go to this id. I still have to compromise on this because this is a shared component with datasets, otherwise I would have done it that way

PR #1651, opened 2023-04-11
1125 nested transformations

nicholas-goodwin:
## Description

- Added a flattening and splitting path for transformations to be able to easily pluck the information needed

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/transformers/lib/transform_actions.ex
+++ b/apps/transformers/lib/transform_actions.ex
@@ -98,22 +98,23 @@ defmodule Transformers do
 
         hierarchy ->
           {parent_key, child_hierarchy} = List.pop_at(hierarchy, 0)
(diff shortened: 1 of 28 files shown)


apps/transformers/lib/transform_actions.ex
```diff
@@ -52,4 +40,76 @@ defmodule Transformers do
...
+
+  defp flatten_payload(payload, parent_key \\ "") do
+    Enum.reduce(payload, %{}, fn {key, value}, acc ->
+      case value do
+        value when is_map(value) ->
+          child_payload = flatten_payload(value, concat_key(key, parent_key))
+          Map.merge(acc, child_payload)
+
+        value when is_list(value) ->
+          IO.inspect("Nicholas - list")
```
accenture-mikeyc, line 72: inspect

PR #1666, opened 2023-04-20
1134 conditional null

nicholas-goodwin:
## Description

- This to add Is Null or Empty to the condition dropdown and added logic

## Reminders:

      in `mix.exs`?


accenture-mikeyc:
You'll have to bump versions for alchemist and transformers apps

accenture-mikeyc (approved):

PR #1673, opened 2023-05-02
1151 customize dataset ingestion fields

ian-j-abbott-accenture:
## Description

- Draft

## Reminders:

      in `mix.exs`?


apps/valkyrie/test/unit/valkyrie/broadway_test.exs
nicholas-goodwin, line 1: can we have a test that tests if the name and ingestion field selector are not the same and it still works? Just for the full test suite

PR #1678, opened 2023-05-05
1165 ingestion selector bug

nicholas-goodwin:
## Description

- fixed Ingestion Field Sync and Selector to not being disabled upon publish
- fixed Ingestion Field Sync from being dropped from the dataset upon publish/coming back

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi/input_schemas/input_converter.ex
+++ b/apps/andi/lib/andi/input_schemas/input_converter.ex
@@ -483,7 +483,6 @@ defmodule Andi.InputSchemas.InputConverter do
     |> Map.delete(:id)
     |> Map.delete(:dataset_id)
     |> Map.delete(:bread_crumb)
-    |> Map.delete(:ingestion_field_sync)
     |> Map.update(:subSchema, nil, fn sub_schema ->
       Enum.map(sub_schema, &drop_fields_from_dictionary_item/1)
     end)
--- a/apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
+++ b/apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
@@ -26,6 +26,12 @@ defmodule AndiWeb.DataDictionary.FieldEditor do
         <%= hidden_input(@form, :bread_crumb) %>
         <%= hidden_input(@form, :dataset_id) %>
         <%= hidden_input(@form, :ingestion_id) %>
+        <%= hidden_input(@form, :name) %>
+        <%= hidden_input(@form, :type) %>
+        <%= hidden_input(@form, :itemType) %>
+        <%= hidden_input(@form, :format) %>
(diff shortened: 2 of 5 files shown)


apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
```diff
@@ -26,6 +26,12 @@ defmodule AndiWeb.DataDictionary.FieldEditor do
         <%= hidden_input(@form, :bread_crumb) %>
         <%= hidden_input(@form, :dataset_id) %>
         <%= hidden_input(@form, :ingestion_id) %>
+        <%= hidden_input(@form, :name) %>
+        <%= hidden_input(@form, :type) %>
```
RyanRConaway, line 30: Why these hidden inputs?

apps/andi/lib/andi/input_schemas/input_converter.ex
```diff
@@ -483,7 +483,6 @@ defmodule Andi.InputSchemas.InputConverter do
     |> Map.delete(:id)
     |> Map.delete(:dataset_id)
     |> Map.delete(:bread_crumb)
-    |> Map.delete(:ingestion_field_sync)
```
RyanRConaway, line 486: Is this the main reason for the ingestion_field_sync state not being persisted on refresh?

apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
nicholas-goodwin: because disabled fields don't push their values onward to the submit. Because of how the changesets going into this are set up, only really the changes get populated. This is to prevent the fields from emptying out if you try to save (they would still be there, just the UI would show empty)

apps/andi/lib/andi/input_schemas/input_converter.ex
nicholas-goodwin: this was, I thought this was going to other places besides just Ecto, so was my b last pass

apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
RyanRConaway: Oh, nice catch on that additional issue. Didn't catch that when I disabled the schema

RyanRConaway (approved):

apps/andi/lib/andi_web/live/shared_components/data_dictionary/data_dictionary_field_editor.ex
nicholas-goodwin: technically, readonly is the way to go through that it allows tabbing through and submission, but not editing (but I think there was a specific requirement to not let it tabbed through, so that's y I deemed to go with this instead)

PR #1700, opened 2023-06-28
Added Validation for errors in Valkyrie

AFarrAccenture:
## Description

- Added Validation for errors in Valkyrie

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/valkyrie/lib/valkyrie/broadway.ex
+++ b/apps/valkyrie/lib/valkyrie/broadway.ex
@@ -90,11 +90,14 @@ defmodule Valkyrie.Broadway do
           error: :failed_schema_validation,
           reason: inspect(reason)
         )
-        
+
         decoded_message_data = Jason.decode!(message_data.value)
         ingestion_id = decoded_message_data["ingestion_id"]
         payload = decoded_message_data["payload"]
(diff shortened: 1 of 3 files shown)


apps/valkyrie/lib/valkyrie/broadway.ex
```diff
@@ -89,12 +90,22 @@ defmodule Valkyrie.Broadway do
           error: :failed_schema_validation,
           reason: inspect(reason)
         )
+        
+        decoded_message_data = Jason.decode!(message_data.value)
+        ingestion_id = decoded_message_data["ingestion_id"]
```
RyanRConaway, line 95: How certain are we that this actually exists? I don't think messages have ingestion_id in valkyrie. Did we hotpatch to confirm?

AFarrAccenture: We did, yes
The ingestion id exists in the message_data.value portion of the message inside a JSON string

RyanRConaway: Awesome. Lets add it to the DeadLetter.process then - Didn't realize we had access to that variable here.

RyanRConaway (approved):

PR #1701, opened 2023-07-14
potential event handling

ian-j-abbott-accenture:
## Description

- ...

## Reminders:

      in `mix.exs`?


apps/andi/lib/andi_web/live/dataset_live_view/event_log_form.ex
```diff
@@ -0,0 +1,163 @@
...
+            </div>
+          </button>
+        </div>
+      </div>
+
+      <div class="component-edit-section--<%= @visibility %>">
+        <div class="event_log_table">
+          <table class="datasets-table" title="Event Log">
+            <thead>
+              <th class="datasets-table__th datasets-table__cell datasets-table__th--sortable datasets-table__th--<%= Map.get(@event_log_order, "timestamp", "unsorted") %>" phx-click="order-by" phx-value-field="timestamp">Timestamp</th>
```
RyanRConaway, line 68: Are we defaulting values to "unsorted"? Would that display the word "unsorted" in the timestamp field if there isn't a value?

RyanRConaway (approved):

PR #1710, opened 2023-08-15
delete event logs after 7 days

ian-j-abbott-accenture:
## Description

Delete event logs after 7 days and limit event log view to 50

## Reminders:

      in `mix.exs`?

### DIFF
--- a/apps/andi/lib/andi/input_schemas/event_logs.ex
+++ b/apps/andi/lib/andi/input_schemas/event_logs.ex
@@ -28,6 +28,17 @@ defmodule Andi.InputSchemas.EventLogs do
     Repo.all(query)
   end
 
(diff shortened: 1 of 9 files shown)


apps/andi/lib/andi/input_schemas/event_logs.ex
```diff
@@ -39,4 +51,15 @@ defmodule Andi.InputSchemas.EventLogs do
...
   end
+
+  def delete_all_before_date(value, unit) do
+    query =
+      from(
+        eventlog in EventLog,
+        where: eventlog.timestamp < ago(^value, ^unit)
+      )
+
+    Repo.all(query) |> IO.inspect(label: "query")
```
RyanRConaway, line 62: IO.inspect

RyanRConaway (approved):

PR #1714, opened 2023-08-28
1228 event log limit

ian-j-abbott-accenture:
## Description

- ...

## Reminders:

      in `mix.exs`?


apps/forklift/lib/forklift/data_writer.ex
```diff
@@ -189,9 +164,11 @@ defmodule Forklift.DataWriter do
     datum |> Forklift.Util.add_to_metadata(:forklift_start_time, Data.Timing.current_time())
   end
 
-  defp add_ingestion_info(data) do
-    payload = Map.get(data, :payload, %{})
+  defp add_ingestion_info(%{payload: payload} = data) when not is_map(payload) do
```
RyanRConaway, line 167: Hmm, should we be more specific here? Besides end_of_data, if the payload isn't a map, I would want an error to be raised. What is payload when it's an end_of_data message? Is it just a string that we could match on exactly in the hydra?

apps/forklift/test/unit/forklift/data_writer_test.exs
```diff
@@ -249,12 +232,15 @@ defmodule Forklift.DataWriterTest do
         technical: %{systemName: "some_system_name"}
```
RyanRConaway, line 232: Rename this test to "Compaction *is* kicked off if end_of_data message is received".

Remove ingestion_status field.

apps/forklift/test/unit/forklift/event/event_handling_test.exs
```diff
@@ -166,120 +166,21 @@ defmodule Forklift.Event.EventHandlingTest do
...
-          Brook.Event.new(
-            type: data_extract_end(),
-            data: fake_extract_end_msg,
-            author: :author
-          )
-        )
-      end)
-    end
-
-    test "*does not* kick off compaction if ingestion_progress is not complete", %{
```
RyanRConaway, line 251: Probably should keep this test, but repurpose it to "*does not* kick off compaction if end_of_data message is not received"

RyanRConaway: Actually, this should just be in the "data_writer_test" file next to the other similar test

apps/reaper/test/integration/reaper/reaper_full_test.exs
```diff
@@ -195,13 +204,14 @@ defmodule Reaper.FullTest do
...
     @tag capture_log: true
+    @tag timeout: 120_000
     test "configures and ingests a csv datasource that was partially loaded before reaper restarted", %{bypass: _bypass} do
       topic = "#{output_topic_prefix()}-#{@partial_load_ingestion_id}"
 
       eventually(
         fn ->
           result = :brod.resolve_offset(brod_endpoints(), topic, 0)
-          assert {:ok, 10_000} == result
+          assert {:ok, 10_001} == result
```
RyanRConaway, line 214: why?

ian-j-abbott-accenture: In the setup function for this test, we bypass the file read and replace it with 10000 lines of random strings. With the EOD message being added now as well, that increments the offset by 1

RyanRConaway (approved):

PR #1718, opened 2023-09-09
Event log polish

ian-j-abbott-accenture:
## Description

- Move timestamps to millisecond precision in the event log table
- Add fixed percentage width to source column

## Reminders:

      in `mix.exs`?


apps/andi/priv/repo/migrations/20230907190131_modify_event_log_timestamp_type.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Andi.Repo.Migrations.ModifyEventLogTimestampType do
+  use Ecto.Migration
+
+  def change do
+    IO.inspect("running migration")
```
RyanRConaway, line 5: IO Inspect

RyanRConaway (approved):

PR #1719, opened 2023-09-21
1237 update deadletter status

ian-j-abbott-accenture:
## Description

- Update logic between getting the dataset status

## Reminders:

      in `mix.exs`?


apps/forklift/lib/forklift/data_writer.ex
```diff
@@ -218,4 +223,18 @@ defmodule Forklift.DataWriter do
       dataset_id: dataset_id
     }
   end
+
+  defp create_ingestion_complete_data(dataset_id, ingestion_id, actual_message_count, extract_time) do
+    {expected_message_count, _} =
+      Redix.command!(:redix, ["GET", ingestion_id])
```
RyanRConaway, line 229: I think this Redis key should be ingestion_id + extraction start time.

Just using the ingestion ID potentially leads to race conditions between ingestion triggers. If another ingestion starts before this one is complete for some reason, the expected message count would have a race condition.

This is more of a consideration if push-based ingestions were ever added (Which is likely a feature if UrbanOS ever handled IoT data). An ingestion could have 2 concurrent ingestions running.

apps/reaper/lib/reaper/data_extract/processor.ex
```diff
@@ -182,4 +184,14 @@ defmodule Reaper.DataExtract.Processor do
       dataset_id: dataset_id
     }
   end
+
+  defp create_data_retrieved_count_message(dataset_id, ingestion_id, expected_message_count, extraction_start_time) do
```
RyanRConaway, line 188: Is this function used?

apps/reaper/lib/reaper/data_extract/processor.ex
```diff
@@ -64,14 +64,16 @@ defmodule Reaper.DataExtract.Processor do
 
     wait_for_completion([producer_stage, validation_stage, schema_stage, load_stage])
 
+    messages_processed_count = messages_processed_count(cache_name)
```
RyanRConaway, line 67: Why rearrange the messages_processed_count?

apps/reaper/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Reaper.MixProject do
   def project do
     [
       app: :reaper,
-      version: "2.0.42",
+      version: "2.0.43",
```
RyanRConaway, line 7: Based on the changes I'm seeing. Looks like reaper may not need to be changed at all?

apps/reaper/lib/reaper/data_extract/processor.ex
ian-j-abbott-accenture: Another symptom of re-doing things a few times and forgetting what all I did...

PR #1730, opened 2024-03-28
Valk Alpine Update

AFarrAccenture:
## Description

- Updating Alpine to solve security issues

## Reminders:

      in `mix.exs`?


AFarrAccenture:
A lot of files got modified unnecessarily.

PR #1733, opened 2024-03-28
update discovery streams alpine

acdouglas00:
## Description
- Updating Discovery Streams Alpine to version 3.16.9

## Reminders:

  - [ x ] If updating patch version, are you sure there are no chart changes required to maintain functionality? If so, you should bump minor version instead.
- [ x] If changing elixir code in an "app", did you update the relevant version
      in `mix.exs`?

### DIFF
--- a/apps/discovery_streams/mix.exs
+++ b/apps/discovery_streams/mix.exs
@@ -4,7 +4,7 @@ defmodule DiscoveryStreams.Mixfile do
   def project do
     [
       app: :discovery_streams,
-      version: "3.16.9",
+      version: "3.0.24",
       build_path: "../../_build",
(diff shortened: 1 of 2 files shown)


apps/discovery_streams/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule DiscoveryStreams.Mixfile do
   def project do
     [
       app: :discovery_streams,
-      version: "3.0.23",
+      version: "3.16.9",
```
jfolk2015, line 7: Should this be 3.0.24?

jfolk2015 (changes requested):

PR #1751, opened 2025-04-28
Update to ubuntu 2404 build image

som-sewardr1:
ubuntu-20.04 has expired updated the build test image to ubuntu-24.04


rseward (approved):
Let's go.

PR #1753, opened 2025-04-28
Update to ubuntu 2404 build image

som-sewardr1:
update Erlang/OTP on build testing image.


rseward (approved):
Let's a give it a try!

PR #1754, opened 2025-04-28
Update erlang/OTP version

som-sewardr1:
Let's give 'er a go.


rseward (approved):
Maybe now?

som-sewardr1:
Let's give 'er a go.

